PR #11, opened 2018-07-11
Uglifying css to speed up times

ryelo:


ryelo:
I don't think sourcemaps would be necessary in this situation. I'd guess most CSS debugging would occur in the inspector which shows all the styles normally.

PR #12, opened 2018-07-12
Fixing pf vertical navigation rendering

ryelo:
Instead of using imports, I'm going to copy the `.css` file for the VerticalNav components into the head while still doing the css reset inside the `chrome.scss`


ryelo:
FYI @bmclaughlin

PR #21, opened 2018-08-14
changing Jenkinsfile to work with OpenShift Jenkins instance

Kinlaw:


Jenkinsfile
```diff
@@ -8,11 +8,23 @@ def wrapStep(String stepName, Closure step) {
...
+
+    stage('deploy') {
+        withCredentials(bindings: [sshUserPrivateKey(credentialsId: 'insightsbot',
+                                                     keyFileVariable: 'insightsbot',
+                                                     passphraseVariable: '',
+                                                     usernameVariable: '')]) {
+            sh '''
+                eval `ssh-agent`
+                ssh-add "$insightsbot"
+                rsync -arv -e "ssh -2" * sshacs@unprotected.upload.akamai.com:/114034/insights/platform/dashboard/
```
ryelo: do we want to rsync on the dashboard?

Kinlaw: nope. Good catch.

Kinlaw: Also, this deploys everything on chrome-build to both insights and insightsbeta. I'm working on a solution to get master to insightsbeta and static to insights

PR #29, opened 2018-09-20
changes all paths from /insights to /insightsbeta

Kinlaw:


karelhala:
I guess this requires change in all apps as wel, am I right?

Kinlaw:
@karelhala yes, it will. I have changed the Jenkinsfile to deploy all master changes to /insightsbeta/ which until now has not worked. This shouldn't affect the apps on /insights/ which is what we've been using up until this point.

Full disclosure: I have broken /insights/platform/dashboard while working on beta releases for Apps. I'm working to fix that now but part of fixing that is setting up beta releases with insights-chrome.

karelhala:
I am a bit of concerned about apps using local chrome, I guess they'd have to change their settings as well right? Can we perhaps make it that if you set env variable $INSIGHTS_ROOT to somethig it will be used? So we don't force consumers to uodate their config everytime we do these kind of changes in a future?

ryelo:
@karelhala @Kinlaw any followup on this?

karelhala:
I think that the best approach here would be to use variable `isDemo` or something like that and when set to true the URL would change to `insightsbeta/platform` if unset or set to false the URL would be same as is now.

Kinlaw:
@ryelo I haven't been able to work on it. I've been working on Fakamai deployments for all the apps.

PR #30, opened 2018-09-20
adding pendo

ryelo:
I guess I won't be able to see this actually working until it hits a live site?


ryelo:
Closing due to having a better and cleaner solution

PR #41, opened 2018-10-03
switching from pf-m-active to  pf-m-current

ryelo:


src/js/nav/globalNav.js
```diff
@@ -35,7 +35,7 @@ function toNavElement(item) {
     li.setAttribute('id', item.id);
     if (item.active) {
```
mattnolting: should this be `item.current` as well?

ryelo: Yep. Fixed!

PR #43, opened 2018-10-03
fixes stable release

Kinlaw:
Previously on stable release, we were checking out the beta Jenkinsfiles and overwriting the stable Jenkinsfile. This change makes it so that we checkout the stable Jenkinsfile.


Kinlaw:
udabes

PR #48, opened 2018-10-11
Allow applications to get user information from JWT

karelhala:


iphands:
@karelhala I dont think this is gonna work. I think we have a bad chance of a race happening with this.
We need should wrap this function in something promise based that waits for Jwt.onInit Jwt.isAuthenticated or onInitialUpdateToken I think.

Do you agree?

PR #49, opened 2018-10-12
Enable secondary nav to be staticly loaded

karelhala:
* [x] Allow secondary nav to be staticly loaded
* [x] Remove old way of using secondary nav

New look:
screenshot from 2018-10-22 16-44-32


iphands:
@karelhala I really like removing the old way, maybe spitting something out into the console so other teams know... point to a doc link about the change in that message??

src/js/nav/globalNav.js
```diff
@@ -17,7 +42,21 @@ export const options = Object.freeze([{
     title: 'Remediations'
 }, {
     id: 'cost-management',
-    title: 'Cost Management'
+    title: 'Cost Management',
+    subItems: [
+        {
+            id: '',
```
ryelo: Should we add IDs here?

karelhala: I guess it doesn't really matter since `'' == undefined == false` but I like to indicate that `overview` leads to nowhere

PR #51, opened 2018-10-16
Add pendo and user information

ryelo:


src/pug/header.pug
```diff
@@ -1,3 +1,7 @@
+script(src='/insights/static/chrome/js/chrome.js')
+script.
+  window.getName();
```
karelhala: Can you call this function from chrome instead of HTML? I am kinda worried that the chrome file will get loaded twice.

src/js/chrome.js
```diff
@@ -2,10 +2,15 @@ import { spinUpStore } from './redux-config';
 import * as actionTypes from './redux/action-types';
 import loadInventory from './inventory';
 import auth from './auth';
+import analytics from './analytics';
 import jwt from 'jwt-redhat';
 
 const onAuth = auth();
 
+onAuth.then(() => {
+    analytics(jwt.getUserInfo());
```
karelhala: Can you call the `getName` function from here? So the data will be loaded emediately when user logs in.

ryelo:
@karelhala I abstracted some of the stuff, so take another look and tell me what you think. Not sure if abstracting the `onAuth` inside the `getName()` function is okay since it gets called within `onAuth` in the beginning of the file. If it's needed, I can add it back into the function.

PR #59, opened 2018-10-30
adding insights jwt

ryelo:
This is a basic jwt module that goes off redhat-jwt, but with tweaks and written in js.

Still a work in progress, but feel free to leave feedback.

Still to do:
- [x] bounce unauthed users to the auth page
- [ ] let apps conditionally turn off the bounce (ex on page /advisor/xyz i dont get bounced)... maybe look at something in the dom?
- [x] refresh an about to expire JWT


src/js/jwt/simulatedPromise.js
```diff
@@ -0,0 +1,39 @@
+export function createPromise() {
```
karelhala: So we really need this file?

PR #64, opened 2018-11-09
changes absolute paths to relative to make chrome agnostic to beta vs…

Kinlaw:
… stable releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,16 @@ node_js:
 - stable
 install:
 - npm ci
-script:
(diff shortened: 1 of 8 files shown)


Kinlaw:
Relative paths weren't enough because if a user went further than one level deep, the paths broke. I ended up using the basepath in head.pug and basing the urls off of that. This keeps the urls consistent regardless of how many levels deep the user navigates. I also updated the travis.yml to build beta or stable depending on the branch that is updated.

PR #70, opened 2018-11-26
Convert header to react with pf's components

karelhala:


karelhala:
>  the icons (settings, search, etc) on mobile are really hard to see.
That's because they are disabled, I am gonna change them so when on large screen they will be icons and smaller screens they will be part of the navigation as text, preferably secondary dropdown (when we will know what to show there)

ryelo:
@karelhala you've got a few unused vars that is failing the build

src/js/redux/action-types.js
```diff
@@ -1,3 +1,5 @@
+export const TOGGLE_ACTION = '@@chrome/toggle';
```
jharting: &lt;nitpick&gt;This is quite ambiguous as it is not clear which toggle this is. Might be helpful to be more specific

src/js/App/Header/Brand.js
```diff
@@ -0,0 +1,32 @@
...
+            <BarsIcon size="md"/>
+        </Button>
+    </div>
+    <a className="pf-l-page__header-brand-link" href="./platform/dashboard/">
+        <Logo />
+    </a>
+</div>);
+
+function mapDispatchToProps(dispatch) {
+    return { toggleNav: () => {
```
jharting: &lt;nitpick&gt;weird indentation

PR #86, opened 2018-12-13
Update FE components to newer version

karelhala:
**Warning** This will change how inventory is loaded so it will break inventory in all applications!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1143,9 +1143,9 @@
       "optional": true
     },
     "@red-hat-insights/insights-frontend-components": {
-      "version": "3.25.0",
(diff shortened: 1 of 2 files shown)


ryelo:
@karelhala you probably want to ping #insights-frontend

dLobatog:
FWIW I adapted to use the new Inventory & InventoryDetail in https://github.com/RedHatInsights/compliance-frontend/pull/23, thanks for the new version, it works great! Looking forward to see it in insights-chrome for CI/QA to see my changes live :+1:

karelhala:
Oh okay, since #88 was closed, this is no longer needed 😀

PR #90, opened 2018-12-19
[WIP] Convert settings icon to a drop-down; add sources link to settings.

martinpovolny:
Sources is supposed to be accessible from the menu. So according to discussion with @karelhala, I am adding `items` to the `actions` with a list of URL fragments to navigate to that render a drop down list below the icon.

The navigation is done using setting of the `window.location` not using the router.

~~requires: https://github.com/patternfly/patternfly-react/pull/1096~~ (redone w/o a change in PF4)

### DIFF
--- a/src/js/App/Header/ToolbarToggle.js
+++ b/src/js/App/Header/ToolbarToggle.js
@@ -1,0 +1,59 @@
+import React, { Component } from 'react';
+import {
+    Dropdown,
+    DropdownToggle,
+    DropdownItem,
+    DropdownPosition
+} from '@patternfly/react-core/dist/esm/components/Dropdown';
+import PropTypes from 'prop-types';
+
+class ToolbarToggle extends Component {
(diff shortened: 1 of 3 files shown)


src/js/App/Header/ToolbarToggle.js
```diff
@@ -0,0 +1,59 @@
...
+    onToggle(isOpen) {
+        this.setState({ isOpen });
+    }
+
+    onClick(_ev, url) {
+        window.location = document.baseURI + url;
+    }
+
+    render() {
+        const dropdownItems = this.props.dropdownItems.map(({ url }) =>
```
karelhala: Title is missing
```JSX
const dropdownItems = this.props.dropdownItems.map(({ url, title }) =>
  <DropdownItem component='button' onClick={ev => this.onClick(ev, url)}>{ title }</DropdownItem>
);
```

PR #102, opened 2019-01-09
Convert settings icon to a drop-down; add sources link to settings.

martinpovolny:
Sources is supposed to be accessible from the menu. So according to discussion with @karelhala, I am adding `items` to the `actions` with a list of URL fragments to navigate to that render a drop down list below the icon.

The navigation is done using setting of the `window.location` not using the router.

Replacement for https://github.com/RedHatInsights/insights-chrome/pull/90

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6236,7 +6236,7 @@
       "dependencies": {
         "core-js": {
           "version": "1.2.7",
-          "resolved": "https://registry.npmjs.org/core-js/-/core-js-1.2.7.tgz",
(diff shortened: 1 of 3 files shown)


martinpovolny:
ping @karelhala

martinpovolny:
I had some disagreement with github (I deleted the old branch and recreated under the same name, but could not get the previous PR to update) so I created a new PR.

martinpovolny:
Fixed, thx!

martinpovolny:
Rebased.

PR #114, opened 2019-01-18
Add loading of other apps to async object of insights

karelhala:


ryelo:
Also, make sure Travis isn't broken

karelhala:
@ryelo sadly we don't know what any of these imports will show. We could assume some things, but that wouldn't be wise. I'd say let consumers who import these decide what to show in meantime of loading.

PR #115, opened 2019-01-18
Fixing collapsable navigation

ryelo:
RHIPLAT2-266


src/js/App/Header/Brand.js
```diff
@@ -11,7 +11,7 @@ const Brand = ({ toggleNav }) => (<div className="pf-l-page__header-brand">
             variant="plain"
             aria-label="Toggle primary navigation"
             widget-type="InsightsNavToggle"
-            onClick={() => toggleNav && toggleNav()}
+            onClick={() => toggleNav && window.navToggle()}
```
karelhala: Please revert this change, becasue you are not calling toggleNav from connected dispatchToProps function.

PR #119, opened 2019-01-23
Fix settings URL and color

karelhala:
RHIPLAT2-350


ryelo:
@karelhala I think this is a bug in Patternfly that will probably be fixed. See: https://patternfly-react.surge.sh/patternfly-4/demos/pagelayout/examples/page-layout-default-nav and click on "Kyle Baker", then hover over an item.

PR #121, opened 2019-01-24
Update about page not to throw errors

karelhala:


src/js/App/Header/InsightsAbout.js
```diff
@@ -28,6 +28,7 @@ class InsightsAbout extends Component {
             ] },
             currentApp: app && app.title
         };
+        this.updateAppVersion = this.updateAppVersion.bind(this);
```
karelhala: We need to bind this so we can call `setStatw`

kruai:
It's true that the About page no longer seems to throw errors with this code. But running on my machine, it doesn't fetch the version numbers either. Do the versions show on your machines @ryelo and @karelhala

PR #129, opened 2019-01-28
Do not call listener if no data were actually called after app nav

karelhala:


ryelo:
@karelhala just missing a semicolon

PR #146, opened 2019-02-01
Features/temp display before loading

kruai:
This pull request addresses RHIPLAT2-369.
The purpose of these changes is to show something less jarring until the app is ready.

This is what displays while the application loads:
screenshot from 2019-02-01 13-38-39


ryelo:
@kruai This looks really good, but I wonder if there's a way we can do this with our Async Component and do it in React instead of Pug - which will probably need some overhaul of the babel file since I don't think it would work as it currently is. If it was not possible, then this should work.

Thoughts?

cc: @karelhala

kruai:
Would doing it that way be faster? All this is really trying to do is get something visually similar to the app displayed as soon as possible. If it would be faster with the Async Component, I'd be interested in learning more, but it sounds like it would have to load some React components for that (instead of just rendering HTML snippets)

ryelo:
@kruai The async component essentially renders HTML snippits while the actual react component is loading. It would be nice because, if possible, we wouldn't have to wait until the entire page is loaded, we can have a separate async component for each "section" of the page, which should allow us to render the sidenav and masthead separately.

iphands:
@ryelo Part of the litmus here is we need this *before* JS comes along.
They real reason we have such nasty experience is there is a few hundred millis or so before React gets strapped and does its thing.

I think we want purs HTML CSS in the pre state... then tear that all down once we have real JS.
But if we can achieve something that looks great I dont really care how we get there.

iphands:
@kruai two things I see...
1) Inline SVG... I think the main reason we do that is to color things in CSS, but dont it blows up the dom size and prevents caching. Just make a white one and  or CSS it in.



2) There is some fade in on the nav... please to kll

iphands:
@kruai Oh wait... the no JS page looks great now. But there is still some flash that happens.
https://youtu.be/Fla9lyYxcZo < check in slomo mebbe

I think we need to move the transition around... delay it more perhaps. Remind me and we can look together Monday.

kruai:
@ryelo OK, fixing that now.
@iphands It looks like the current flash effect is caused by 2 loads. First is the background image we're using, and the second is the actual page content. When you have time, let's chat about how to speed this up, or at least make it smoother.

ryelo:
@kruai it looks like 
```
@import 'node_modules/@patternfly/patternfly-next/patternfly-base.scss';
@import 'node_modules/@patternfly/patternfly-next/layouts/Page/page.scss';
```
from https://github.com/RedHatInsights/insights-chrome/blob/master/src/sass/chrome.scss#L5 needs to be changed to use component instead of layout

kruai:
Debugging the styling took a bit of time bc I forgot I'd have to modify the frontend-starter-app styling as well to see my expected results.
However, we'll need to wait for the patternfly-next changes to be in `master` before we can merge this.

ryelo:
@kruai yeah we should wait until #147 merges before merging this one

karelhala:
The `package-lock.json` is really strange, id you by any chance removed the file and ran `npm install`? If so please check upstream version and re-run `npm install`

kruai:
Ah yes, I did when I was running into installation issues while working with Ian. I wasn't aware that would cause issues, sorry! I'll go fix that and read up on the lock file

kruai:
I did a few things that hopefully fixed it and brought my project more in line with the coming patternfly changes:

- Replaced my `package-lock.json` with the one from `master` and ran `npm install`
- Merged Karel's patternfly branch into mine
- Ran `npm install` again
- Verified with `npm run-script build`, ran the app to make sure everything looks good

The changes to `package-lock.json` don't look quite as drastic now. Is this more in line with what you'd expect?

karelhala:
@kruai yest, that looks much better! Btw are we OK with that once user loads the UI and tries clicking on anything untill the page loads there is no interaction possible?

kruai:
I'm removing the temp display for the main content. Figuring out when to remove the temp node in the DOM is too hard to figure out right now, since that panel loads in from an async call in the child app.

PR #150, opened 2019-02-06
Enhancements/single esi include

kruai:
The addition of this one file allows the frontend-starter project to chrome using one `<esi:include>` tag.
Because `master` won't pass Travis until the patternfly fixes are complete, I doubt Travis will be happy with this change, either.


src/pug/body.pug
```diff
@@ -0,0 +1,6 @@
+div.pf-c-background-image
+div.pf-c-page#page
+    include header.pug
+    include sidebar.pug
+    main.pf-l-page__main#root(role='main')
+        include footer.pug
```
iphands: Hrrmmm this is tricky.
Right now Chrome.js is incorrectly included in sidebar.html.

We probably want to correct this w/o causing breakage for current consumers.

Can we
- duplicate sidebar (sidebar2)
- remove the chrome.js include in sidebar2
- use sidebard2 in body.html
- "after.html" or "final.html"
- put the chrome.js include in after.html
- put after.html on the bottom
- put a script console.log('Deprecation warning') in the old sidebar.html

I think that leaves unupdated applications working and lets updated apps transition to the new unbroken stuff.

kruai: I assumed we'd want more than just "Deprecation warning", so I made it a bit more descriptive:
`Deprecation warning: Please update your usage of Chrome.`

src/pug/body.pug
```diff
@@ -0,0 +1,6 @@
+div.pf-c-background-image
+div.pf-c-page#page
+    include header.pug
+    include sidebar2.pug
+    main.pf-c-page__main#root(role='main')
```
karelhala: We are missing `include footer.pug` here.

src/pug/sidebar.pug
```diff
@@ -1,2 +1,4 @@
+script
+    | console.log('Deprecation warning: Please update your usage of Chrome.')
```
karelhala: `console.warn` is much better. Logs are usually overlooked.

src/pug/body.pug
kruai: It sounds like we're trying to kill `footer.pug` though. @iphands ?

src/pug/sidebar.pug
kruai: Will do.

kruai:
I renamed the old sidebar, and changed console.log to console.warn.
Do we still need `footer.pug`? I deleted its contents a while ago since we moved its info to the About modal. If its removal doesn't break anything, I say we should scrap it.

kruai:
In regards to the renaming of `sidebar2.pug`, that's actually the newer one. 
`sidebar.pug` is going to be deprecated, but we're keeping that naming so that we don't break the app for people that haven't updated their index page yet.
So, I've renamed that file again to `sidebar-new.pug`.

iphands:
> In regards to the renaming of sidebar2.pug, that's actually the newer one.
> sidebar.pug is going to be deprecated, but we're keeping that naming so that we don't break the app for people that haven't updated their index page yet.
> So, I've renamed that file again to sidebar-new.pug.

Exactly we need to leave sidebar alone because people are currently still linking directly to it.
We can should make a new one... In general I would never use -new because then a newer one needs to be -new-new but it does not matter a ton because we can rename it back to sidebar once everyone is migrated over.

src/pug/body.pug
iphands: Yeah. Kill footer.
For the longest time it was embedded in <main> so it would disappear as soon as the page loaded anyway. Because of <- no one can rightly claim they were using it for anything. Also for a few weeks now footer has been an empty string.

kruai:
README updated and fixed. Do the new implementation instructions make sense to you?

ryelo:
@kruai Yep! Thanks!

kruai:
@iphands @karelhala @ryelo - Anything else I should change? If not, I don't think there's anything else holding this one back.

PR #156, opened 2019-02-12
Update globalNav.js

jonnyfiveiq:
The order of the left nav should be Dashboard, Catalog.....and so on....why? The Dashboard is always first....then you have Day 1 operations eg Catalog then Day 2 follows after.....thanks


iphands:
@jonnyfiveiq Fine with me as long as UXD thinks is cool. Talking to UXD now. Should be easy peasy

iphands:
Tracking in #157

PR #173, opened 2019-02-22
Change pre-load sidenav to use spooky skeletons

kruai:
Instead of using actual text for the sidenav, we're going to use some skeletons.

### DIFF
--- a/src/pug/sidebar.pug
+++ b/src/pug/sidebar.pug
@@ -1,23 +1,17 @@
+- var i = 0;
 script
     | console.warn('Deprecation warning: Please update your usage of Chrome.')
 script(src='./static/chrome/js/chrome.js')
 aside.pf-c-page__sidebar.pf-l-page__sidebar#ins-c-sidebar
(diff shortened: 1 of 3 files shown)


src/pug/sidebar.pug
```diff
@@ -1,23 +1,17 @@
...
-                a(class='pf-c-nav__link')
-                    div.ins-c-skeleton.ins-c-skeleton__lg &nbsp;
-            li.pf-c-nav__item
-                a(class='pf-c-nav__link')
-                    div.ins-c-skeleton.ins-c-skeleton__lg &nbsp;
+            while i < 4
+                li.pf-c-nav__item
+                    a.pf-c-nav__link.ins-c-skeleton__link
+                        .ins-c-skeleton.ins-c-skeleton__lg &nbsp;
+                        - i++ss
```
ryelo: `- i++ss` ?

kruai: Sometimes my new keyboard likes to not pick up the ctrl input sometimes and I end up with things like this ugh

PR #176, opened 2019-03-06
Add groupped navigation

karelhala:
This is preparation to add landing page application. If no cloud services echnology is selected (e.g. not clicked from landing page) legacy sring will be shown with home button and such.

### Before
screenshot from 2019-03-06 16-31-00

### After
screenshot from 2019-03-06 16-30-23

### DIFF
--- a/src/js/App/Sidenav/index.js
+++ b/src/js/App/Sidenav/index.js
@@ -1,4 +1,21 @@
-import React from 'react';
+import React, { Fragment, Component } from 'react';
 import Navigation from './Navigation';
+import { Split, SplitItem } from '@patternfly/react-core/dist/esm/layouts/Split';
+import HomeIcon from '@patternfly/react-icons/dist/esm/icons/home-icon';
 
-export default () => <Navigation />;
+class SideNav extends Component {
+    onHomeButtonClick() {
+        window.location.href = `${document.baseURI}platform/landing`;
+    }
+
+    render() {
+        return (<Fragment>
+            <Split className="ins-c-navigation__header">
+                <SplitItem onClick={this.onHomeButtonClick}><HomeIcon size="md" /></SplitItem>
+                <SplitItem isMain>{localStorage.getItem('cs-app-title') || `Legacy`}</SplitItem>
+            </Split>
+            <Navigation />
+        </Fragment>);
+    }
+}
+export default SideNav;
--- a/src/js/App/Sidenav/index.js
+++ b/src/js/App/Sidenav/index.js
@@ -4,15 +4,18 @@ import { Split, SplitItem } from '@patternfly/react-core/dist/esm/layouts/Split'
 import HomeIcon from '@patternfly/react-icons/dist/esm/icons/home-icon';
 
 class SideNav extends Component {
-    onHomeButtonClick() {
-        window.location.href = `${document.baseURI}platform/landing`;
-    }
 
     render() {
         return (<Fragment>
             <Split className="ins-c-navigation__header">
-                <SplitItem onClick={this.onHomeButtonClick}><HomeIcon size="md" /></SplitItem>
-                <SplitItem isMain>{localStorage.getItem('cs-app-title') || `Legacy`}</SplitItem>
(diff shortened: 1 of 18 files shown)


src/sass/partials/_sidebar.scss
```diff
@@ -0,0 +1,20 @@
+aside.pf-c-page__sidebar {
+    padding-top: 0px;
+
+    .ins-c-navigation__header {
+        max-height: 4.4rem;
```
ryelo: According to the mocks, this should be the same height as the page header. Also, we need to use the rem utility here

src/js/App/Sidenav/index.js
```diff
@@ -1,4 +1,21 @@
...
-export default () => <Navigation />;
+class SideNav extends Component {
+    onHomeButtonClick() {
+        window.location.href = `${document.baseURI}platform/landing`;
+    }
+
+    render() {
+        return (<Fragment>
+            <Split className="ins-c-navigation__header">
+                <SplitItem onClick={this.onHomeButtonClick}><HomeIcon size="md" /></SplitItem>
```
ryelo: We should either use an `a` or a `button` here for accessibility reasons

src/sass/partials/_sidebar.scss
```diff
@@ -0,0 +1,20 @@
...
+        overflow: hidden;
+
+        .pf-l-split__item {
+            padding: 20px;
+            border-bottom: var(--pf-global--BorderWidth--sm) solid #cccccc;
+            &:first-child {
+                cursor: pointer;
```
ryelo: We shouldn't need this when the home button is converted to an `a` or a `button`

src/sass/partials/_sidebar.scss
```diff
@@ -0,0 +1,20 @@
+aside.pf-c-page__sidebar {
+    padding-top: 0px;
+
+    .ins-c-navigation__header {
+        max-height: 4.4rem;
+        overflow: hidden;
+
+        .pf-l-split__item {
+            padding: 20px;
```
ryelo: Please use the rem Utility

src/sass/partials/_sidebar.scss
```diff
@@ -0,0 +1,20 @@
+aside.pf-c-page__sidebar {
+    padding-top: 0px;
```
ryelo: `padding-top: 0` should suffice. No need to put `px` afterwards

src/sass/partials/_sidebar.scss
```diff
@@ -0,0 +1,20 @@
...
+    .ins-c-navigation__header {
+        max-height: 4.4rem;
+        overflow: hidden;
+
+        .pf-l-split__item {
+            padding: 20px;
+            border-bottom: var(--pf-global--BorderWidth--sm) solid #cccccc;
+            &:first-child {
+                cursor: pointer;
+                border-right: var(--pf-global--BorderWidth--sm) solid #cccccc;
```
ryelo: Could we use a PF color here?

src/sass/partials/_footer.scss
```diff
@@ -1,5 +1,5 @@
 .ins-c-footer {
-  background: #1d1d1d;
+  background: var(--pf-global--BackgroundColor--300);
   color: #6c6c6c;
```
ryelo: Is there any reason we wouldn't want to use a PF color here?

src/sass/partials/_sidebar.scss
karelhala: Good point!

src/sass/partials/_sidebar.scss
karelhala: Will do.

src/js/App/Sidenav/index.js
karelhala: Oh yeah, good point.

iphands:
Using admin privs to squash and merge. I think we would like to update the things @ryelo called out... but I'd like to circle back and do it. We need to get the cloud CI working asap.

PR #179, opened 2019-03-08
Complete rename of service-portal to catalog

gmcculloug:
Change top-level nav id from `service-portal` to `catalog`.


gmcculloug:
cc @iphands

PR #200, opened 2019-03-22
Update second level nav to work correctly

karelhala:
~~* No visual change~~
* Adds settings and notification under it
Screenshot from 2019-03-22 15-01-38

* Allows switching from catalog to cost management

### DIFF
--- a/src/js/App/Sidenav/Navigation.js
+++ b/src/js/App/Sidenav/Navigation.js
@@ -22,7 +22,7 @@ class Navigation extends Component {
     };
 
     onClick(_event, item, parent) {
-        const { onNavigate, onClearActive, activeGroup } = this.props;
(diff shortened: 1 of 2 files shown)


karelhala:
ping @bastilian and @adamruzicka this PR adds your navigation endpoint. So far only into `insights` app, if you want any other please specify under which your service should be available.

Your future URL will be `/insights/notifications` is that OK?

adamruzicka:
`/insights/webhooks` for now

karelhala:
> `/insights/webhooks`for now

Oh gotcha, you might want to update that in https://github.com/RedHatInsights/notifications-frontend/pull/19 as well (`insights.chrome.identifyApp('notifications');` -> `insights.chrome.identifyApp('webhooks');`, nginx and common config)

PR #208, opened 2019-03-26
conditionally enable iqe plugin used to determine if there are pending XHR requests

iphands:


src/js/iqeEnablement.js
```diff
@@ -0,0 +1,42 @@
...
+        return send.apply(this, arguments);
+    };
+}
+
+export default {
+    init: () => {
+        if (!initted) {
+            initted = true;
+            if (window.localStorage &&
+                window.localStorage.getItem('iqe:chome:init') === 'true') {
```
ryelo: Should this read `iqe:chrome:init`?

iphands: yeah! Thanks.

PR #219, opened 2019-04-01
Add entitlements API and use it with each user check

karelhala:
**WARNING**
This will cause probably that Red Hat Enterprise Linux management services will not be accessible, because smart management is always evalueted as not entitled.

### DIFF
--- a/src/js/jwt/insights/user.js
+++ b/src/js/jwt/insights/user.js
@@ -4,7 +4,7 @@ const pathMapper = {
     rhel: 'smart_management',
     hybrid: 'hybrid_cloud',
     insights: 'insights',
-    uhc: 'openshift'
+    openshift: 'openshift'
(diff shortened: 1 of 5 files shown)


src/js/jwt/insights/user.js
```diff
@@ -1,4 +1,11 @@
 const log = require('../logger')('insights/user.js');
+const servicesApi = require('./entitlements');
+const pathMapper = {
+    rhel: 'smart_management',
+    hybrid: 'hybrid_cloud',
+    insights: 'insights',
+    uhc: 'openshift'
```
ryelo: do we want to change this to `openshift` instead of `uhc`?

karelhala: Oh yeah, good point now that routes has changed.

PR #221, opened 2019-04-01
Updating adobe analytics to include adobe prod/pre-prod

ryelo:


src/js/analytics.js
```diff
@@ -1,11 +1,36 @@
 'use strict';
 
+function isDev() {
+    return window.location.hostname.split('.')[1] === 'foo' ? true : false;
```
kruai: This can just be:
`return window.location.hostname.split('.')[1] === 'foo';`

src/js/analytics.js
```diff
@@ -1,11 +1,36 @@
...
+}
+
+function initAdobe() {
+    const adobe = document.createElement('script');
+
+    // production: dtm.js
+    // pre-prod: dtm-staging.js
+    const isStaging = (window.location.hostname === 'cloud.redhat.com' ? '' : '-staging');
```
kruai: `isStaging` would imply that it's a boolean. If you want to keep it as a string, I'd use a more descriptive name.
Or, you can keep it as a boolean and change things later:
`const isStaging = (window.location.hostname === 'cloud.redhat.com')`;

PR #224, opened 2019-04-02
Updating skeleton loading for new navigation

ryelo:
Before:
Screen Shot 2019-04-02 at 10 44 43 AM
After:
Screen Shot 2019-04-02 at 10 42 29 AM

### DIFF
--- a/src/pug/sidebar-new.pug
+++ b/src/pug/sidebar-new.pug
@@ -1,5 +1,12 @@
 - var i = 0;
 aside.pf-c-page__sidebar.pf-l-page__sidebar#ins-c-sidebar
+    .pf-l-split.ins-c-navigation__header
+        .pf-l-split__item.ins-c-page__home-icon
+            a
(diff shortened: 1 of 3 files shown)


src/pug/sidebar-new.pug
```diff
@@ -1,5 +1,12 @@
 - var i = 0;
 aside.pf-c-page__sidebar.pf-l-page__sidebar#ins-c-sidebar
+    .pf-l-split.ins-c-navigation__header
+        .pf-l-split__item.ins-c-page__home-icon
+            a
+                svg(style='vertical-align: -0.1875em;' fill='currentColor' height='1.5em' width='1.5em' viewbox='0 0 576 512' aria-hidden='true' role='img')
```
kruai: Since the Home icon is also used once the page is loaded, can we put the svg in /static/images and reference it in both locations?

ryelo: We _could_, but the home icon is actually coming from the Patternfly/react-icons library

ryelo: @kruai Updated to just import the icon from font-awesome (which PF provides)

PR #245, opened 2019-04-10
Rename webhooks to hooks

adamruzicka:
Opening as a draft since the changes need to be coordinated across projects to minimize downtime

### DIFF
--- a/src/js/nav/globalNav.js
+++ b/src/js/nav/globalNav.js
@@ -25,9 +25,9 @@ export const grouppedNav = {
                 title: 'Settings',
                 subItems: [
                     {
-                        id: 'webhooks',
-                        title: 'Webhooks',
-                        reload: 'webhooks'
+                        id: 'hooks',
+                        title: 'Hooks',
+                        reload: 'hooks'
                     },
                     {
                         id: 'sources',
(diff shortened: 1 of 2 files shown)


src/js/nav/globalNav.js
```diff
@@ -25,9 +25,9 @@ export const grouppedNav = {
                 title: 'Settings',
                 subItems: [
                     {
-                        id: 'webhooks',
-                        title: 'Webhooks',
-                        reload: 'webhooks'
+                        id: 'hooks',
+                        title: 'Hooks',
+                        reload: 'hooks'
```
kruai: To follow the pattern for Settings apps, this should probably be `reload: 'settings/hooks'`.

iphands:
@adamruzicka we are going to change the appid too... your assets will be at /apps/hooks and /beta/apps/hooks so a small change on your app is required.

adamruzicka:
I was counting on that, we have a PR ready to handle it when it happens

kruai:
@adamruzicka in addition to the comment I made above, the `config/index_map.json` file should also be updated accordingly.

adamruzicka:
It should be good to go from our side, things work differently under the hood than I thought they would.

Issue #222, opened 2019-04-01
Navigation pane does not highlight properly


dlabrecq:
The Insights navigation pane does to highlight properly when navigating between pages. For example, the user can jump from the Cost Management's overview page via a link to the OpenShift details page. Is there something the Cost Management UI needs to do in order for Insights to understand which route the user has jumped to?

iphands:
Yeah there is some code that an app has to do to tell chrome which subnav item is currectly "active".
@dlabrecq this is what you are asking about right?

@karelhala whats the best example/doc we have of this?

karelhala:
@dlabrecq I checked current UI and can't reproduce your error. When I navigate to `/hybrid/catalog/portfolios` and change url to `/hybrid/cost-management/ocp?group_by[project]=*&order_by[cost]=desc` correct item in navigation is selected and highlighted.

![Peek 2019-04-08 17-51](https://user-images.githubusercontent.com/3439771/55738223-f825e180-5a26-11e9-911d-78ff25dc87e5.gif)

dlabrecq:
When in the overview page, click on the "All Projects" link. The URL in the browser changes to https://ci.foo.redhat.com:1337/hybrid/cost-management/ocp?order_by[cost]=desc; however, the "Overview" navigation link is still highlighted. 

![chrome-capture](https://user-images.githubusercontent.com/17481322/55741244-d0208900-59fa-11e9-8b2a-00f58a431701.gif)

karelhala:
Oh I see thanks for clearing it up. So you want to perform in app redirect that activates app in navigation. I don't think there's an action to that, but we will add it to allow such thing.

dlabrecq:
When I call `insights.chrome.appNavClick({id: 'ocp-on-aws'})`, the navigation pane collapses. I cannot see if the selection was made.

![chrome-capture](https://user-images.githubusercontent.com/17481322/56585806-8f08a700-65ac-11e9-9a13-a97c246fe1b6.gif)

karelhala:
Thanks for pointing this out, the initial fix was only for first level navs and I did not checked with secondary level navigation, PR https://github.com/RedHatInsights/insights-chrome/pull/262 fixes this issue, so calling `insights.chrome.appNavClick({id: 'ocp-on-aws', secondaryNav: true})` will highlight correct item in nav.

dlabrecq:
[...]

PR #247, opened 2019-04-11
Clear active after navClick is called

karelhala:
Fixes: https://github.com/RedHatInsights/insights-chrome/issues/222

This PR adds option to call app nav from application itself. Just call `insights.chrome.appNavClick({ id: 'page-id'})` and correct item in navigation will highlight.

Also resolves RHIPLAT2-656

### DIFF
--- a/src/js/entry.js
+++ b/src/js/entry.js
@@ -25,7 +25,7 @@ export function chromeInit(libjwt) {
     const { store, middlewareListener, actions } = spinUpStore();
 
     // public API actions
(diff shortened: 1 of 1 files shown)


dlabrecq:
When I call `insights.chrome.appNavClick({id: 'ocp-on-aws'})`, the navigation pane collapses. I cannot see if the selection was made.

Perhaps `clearActive()` should be called first?

chrome-capture

PR #248, opened 2019-04-11
Change /webhooks to settings/webhooks

kruai:
Since this is a settings link, it needs to be `settings/webhooks` to follow the pattern.


karelhala:
@kruai are you aware of this PR? https://github.com/RedHatInsights/insights-chrome/pull/245

kruai:
I was not aware, but I am now. I'll make that change now (bc we want it to be `settings/hooks` anyway)

PR #250, opened 2019-04-14
Use in memory cache for entitlements

karelhala:
Sotre entitlements request in in-memory cache and invalidte it every 5 minutes. Page refreshes will invalidate this cache as well.

### DIFF
--- a/src/js/jwt/insights/user.js
+++ b/src/js/jwt/insights/user.js
@@ -46,11 +46,14 @@ module.exports = (token) => {
         if (window.location.hostname === 'cloud.redhat.com' ||
             window.location.hostname.indexOf('qa') === 0) {
             return new Promise(resolve => {
-                resolve(user);
+                resolve({
+                    ...user,
+                    entitlements: {}
+                });
             });
         }
 
-        return servicesApi.servicesGet().then(data => {
+        return servicesApi(token.jti).servicesGet().then(data => {
             const service = pathMapper[pathName[0]];
             if (pathName.length > 0 && pathName[0] !== '') {
                 if (data[service] && data[service].is_entitled) {
--- a/src/js/jwt/insights/user.js
+++ b/src/js/jwt/insights/user.js
@@ -46,10 +46,11 @@ module.exports = (token) => {
         if (window.location.hostname === 'cloud.redhat.com' ||
             window.location.hostname.indexOf('qa') === 0) {
             return new Promise(resolve => {
-                resolve({
-                    ...user,
+                user.identity = {
+                    ...user.identity,
                     entitlements: {}
-                });
+                };
+                resolve(user);
             });
         }
 
@@ -66,10 +67,11 @@ module.exports = (token) => {
                 }
             }
 
-            return {
-                ...user,
+            user.identity = {
+                ...user.identity,
                 entitlements: data
             };
+            return user;
         });
     } else {
         log('User not ready');
--- a/src/js/jwt/insights/user.js
+++ b/src/js/jwt/insights/user.js
@@ -63,7 +63,10 @@ module.exports = (token) => {
                 }
             }
 
-            return user;
+            return {
+                ...user,
+                entitlements: data
+            };
         });
     } else {
         log('User not ready');
--- a/package-lock.json
+++ b/package-lock.json
@@ -5135,6 +5135,15 @@
         "is-buffer": "^1.1.5"
       }
     },
+    "axios-cache-adapter": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/axios-cache-adapter/-/axios-cache-adapter-2.3.0.tgz",
+      "integrity": "sha512-6wegNDj1SB8iIR0mHKThNCRV+JytUX/eg1xgwvLCrZI+wVX2AiNr9pXL3rdQeY4b9njDPuV4y4w/k6wMc9ZwNg==",
+      "requires": {
+        "axios": "^0.18.0",
+        "lodash": "^4.17.11"
+      }
+    },
     "babel-cli": {
       "version": "6.26.0",
       "resolved": "https://registry.npmjs.org/babel-cli/-/babel-cli-6.26.0.tgz",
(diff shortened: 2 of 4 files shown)


karelhala:
@iphands this kind of cache seems a bit better than localstorage that would require much more changes and invalidate the cache by ourselves.

iphands:
@karelhala ummm did you look at the source code for this?

Think about clicking between applications... the cache will disappear when going from landing -> dash -> compliance.

I'd use an abstraction layer sure so we dont have to mess with localStorage directly... but not this one. We need something that persists between applications.

iphands:
Perhaps use this?
https://github.com/RasCarlito/axios-cache-adapter#use-localforage-as-cache-store

karelhala:
@iphands I was thinking about localstorage and stting it in this cache as wel, but felt like that it could be a security problem. Users can access localstorage and change it as they please and entitlements endpoint will start returning wrong data.

Plus what if user logs out and back in again? The data could stay for some time in localstorage and he would be unable to access some part of app or on other hand could lead to security problem. So checking on each app switch nav and storing these data in memory feels kinda better. Thinking about it I guess we could increase the time before cache is invalidated to infinite and refresh it only on page refresh.

iphands:
No this is wrong. There is no security breakdown. The API does not ever trust the frontends notions entitlement cache. Also the front-end does not when pass down the entitlement data to the API. Thia even if you fake the entitlement data... You still get denied in The API.

The logout and login should we handled by using and appropriate cache key. If you key on the username then swithing users would not get you the same cache. You could also use the JWT uuid (jti) as the cache key, then a logout and login would  invalidate the cache even if you came back as the same user.

iphands:
@karelhala tested it works great :1st_place_medal:

src/js/jwt/insights/user.js
```diff
@@ -63,7 +66,10 @@ module.exports = (token) => {
                 }
             }
 
-            return user;
+            return {
```
iphands: @karelhala @ryelo 
I am pretty sure we want:

```javascript
  user.identity.entitlements = data;
  return user;
```
Ryan do you know what the 3scale object is going to look like... I think it is:

```
{
  "identity": {
    "account_number": "6089719",
    "type": "User",
    "user": {
      "username": "rhn-insightsqe-ihands",
      "email": "ihands+qa@redhat.com",
      "first_name": "Ian",
      "last_name": "Hands",
      "is_active": true,
      "is_org_admin": true,
      "is_internal": false
    },
    "internal": {
      "org_id": 52399467
    },
    "entitlements": {
      "hybrid_cloud": {
        "is_entitled": true
      },
      "insights": {
        "is_entitled": true
      },
      "openshift": {
        "is_entitled": true
      },
      "smart_management": {
        "is_entitled": false
      }
    }
  }
}
```

iphands: @ryelo @karelhala 
I think we actually want
```javascript
user.identity.entitlements = data;
return user;
```
Ryan the 3scale object is going to look like this, right?
```javascript
{
  "identity": {
    "account_number": "6089719",
    "type": "User",
    "user": {
      "username": "rhn-insightsqe-ihands",
      "email": "ihands+qa@redhat.com",
      "first_name": "Ian",
      "last_name": "Hands",
      "is_active": true,
      "is_org_admin": true,
      "is_internal": false
    },
    "internal": {
      "org_id": 52399467
    },
    "entitlements": {
      "hybrid_cloud": {
        "is_entitled": true
      },
      "insights": {
        "is_entitled": true
      },
      "openshift": {
        "is_entitled": true
      },
      "smart_management": {
        "is_entitled": false
      }
    }
  }
}
```

karelhala: I didn't wanted to mess with identity object, but if 3scale's gonna return it in this way I'll update it. Thanks for the review and test!

karelhala:
Restarting travis

src/js/jwt/insights/user.js
ryelo: The entitlements stuff isn't going to return "entitlements"
just:
```
      "hybrid_cloud": {
        "is_entitled": true
      },
      "insights": {
        "is_entitled": true
      },
      "openshift": {
        "is_entitled": true
      },
      "smart_management": {
        "is_entitled": false
      }

```

iphands: No, what we are wondering here is what the shape of the x-rh-identity header object is.
3scale adds the entitlements response to the identity object.... what does it look like after that?

@coderbydesign what does the object look like after you are done mixing in entitlements?
```javascript
{
  identity: ...
  entitlements: ...
}
```

Or
```javascript
identity: {
  entitlements: ...
  ...
}
```

?? Or something different

coderbydesign: @iphands what I'm working on right now has `entitlements` nested under `identity`, as the second example shows. 

Per your comment, @ryelo, I'm expecting the entitlements payload to come from the service as you mention, not nested under an `entitlements` key, but I'm setting entitlements on the identity header explicitly, something like:
```
identity.entitlements = entitlements_response_body
```
so that it does come back in `x-rh-identity` as:
```json
{
  "identity": {
    "account_number": "6089719",
    "type": "User",
    "user": {
      "username": "rhn-insightsqe-ihands",
      "email": "ihands+qa@redhat.com",
      "first_name": "Ian",
      "last_name": "Hands",
      "is_active": true,
      "is_org_admin": true,
      "is_internal": false
    },
    "internal": {
      "org_id": 52399467
    },
    "entitlements": {
      "hybrid_cloud": {
        "is_entitled": true
      },
      "insights": {
        "is_entitled": true
      },
      "openshift": {
        "is_entitled": true
      },
      "smart_management": {
        "is_entitled": false
      }
    }
  }
}
```

iphands: @karelhala so I think the current code matches ^ w/r/t the entitlements piece.
Merging

iphands:
@karelhala Damn! :D 
Sorry for the churn but 3scale dudes just flipped the logic on their end to make the object look like:
```javascript
{
  identity: ...,
  entitlements: ...,  
}
```

PR #256, opened 2019-04-17
Patch to let UHC make offline token requests to KeyCloak not gobble up the data

iphands:
To use this you can
```javascript
window.insights.chrome.auth.doOffline()
```

Then after the postback
```javascript
window.insights.chrome.auth.getPostbackUrl()
```

Note you can only call getPostbackUrl and obtain the data once

### DIFF
--- a/src/js/jwt/insights/offline.test.js
+++ b/src/js/jwt/insights/offline.test.js
@@ -1,5 +1,5 @@
 /*global describe, test, expect, require, beforeEach, jest*/
-import mockAxios from 'jest-mock-axios';
+const mockAxios = require('jest-mock-axios').default;
 const offline = require('./offline');
 
 const defaults = {
--- a/src/js/jwt/insights/offline.test.js
+++ b/src/js/jwt/insights/offline.test.js
@@ -1,0 +1,111 @@
+/*global describe, test, expect, require, beforeEach, jest*/
+import mockAxios from 'jest-mock-axios';
+const offline = require('./offline');
+
+const defaults = {
+    location: {
+        hash: 'fdsjkfjkfsdkfjksdjklsjf',
+        href: 'https://test.com?noauth=2402500adeacc30eb5c5a8a5e2e0ec1f#foo=bar'
+    }
+};
+
+function getMockWindow(location) {
+    const loc = location || defaults.location;
+    return {
+        location: loc,
+        history: {
+            pushState: (one, two, url) => {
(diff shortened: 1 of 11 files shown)


elad661:
I tested this locally and it works as expected

elad661:
To make the http request to get the token, I need a bit more info than the postback URL - I need the keycloak server address and the keycloak client_id. Can you add these to the API as well?

iphands:
@karelhala, @elad661 provided a patch to do the actual POST needed to get the final offline token
I went ahead and refactored a bit, then added some unit tests. Can you please review and contribute any style changes you'd like to see? I want to get this merged quickly.

@elad661 please test the change as it is now and let me know if something functional is broken. Thanks.

src/js/jwt/jwt.js
```diff
@@ -72,13 +73,33 @@ priv.decodeToken = (str) => {
...
     return str;
 };
 
+pub.doOffline = (key, val) => {
+    const url = urijs(window.location.href);
+    url.removeSearch(key);
+    url.addSearch(key, val);
+
+    const options = {
+        realm: 'redhat-external',
```
karelhala: Can you export [src/js/auth.js#L7~options](https://github.com/RedHatInsights/insights-chrome/blob/bc0725494efc0adc38059dd6bd19a7911d17de54/src/js/auth.js#L7) const and use it here as well? So we don't have to keep track of the same vars in two places.

src/js/jwt/insights/offline.test.js
```diff
@@ -0,0 +1,111 @@
...
+    describe('getOfflineToken', () => {
+        test('fails when there is no offline postbackUrl', async () => {
+            try {
+                await offline.getOfflineToken();
+            } catch (e) {
+                expect(e).toBe('not available');
+            }
+        });
+
+        test('POSTs to /token with the right parameters when input is good', () => {
```
karelhala: Missing async.

elad661:
@iphands thanks, I tested this and it seems to work well.

oourfali:
@elad661 any change needed on our side to have that fixed? as I still see network error in the UI after a few minutes.

elad661:
@oourfali this is unrelated to the issue you're seeing. This PR is to enable the "offline API tokens" page. The issue you're seeing is a different one.

oourfali:
I see.
Thanks.

PR #259, opened 2019-04-18
Changing the Ident object schema again

iphands:
Alberto and Keith are flipping it back to how you had it before...
https://github.com/RedHatInsights/insights-3scale/pull/48

Sorry for the churn @karelhala

### DIFF
--- a/src/js/jwt/insights/user.js
+++ b/src/js/jwt/insights/user.js
@@ -74,9 +74,10 @@ module.exports = (token) => {
                 }
             }
 
-            user.entitlements = data;
-
-            return user;
+            return {
+                ...user,
+                entitlements: data
+            };
         });
     } else {
         log('User not ready');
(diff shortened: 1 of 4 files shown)


src/js/jwt/insights/user.js
```diff
@@ -67,10 +67,8 @@ module.exports = (token) => {
                 }
             }
 
-            user.identity = {
-                ...user.identity,
-                entitlements: data
-            };
+            user.entitlements = data;
+
             return user;
```
karelhala: Please use spread in here, this way even if user is not defined it will return an object.
```JS
return {
  ..user,
  entitlements: data
}
```

iphands: @karelhala This pattern... eww! 
For one... the code is wrapped in an `if (user)`
Two https://jsbench.me/ihjuu95twc/1

Also, dont we want to have the thing fail if user is falsey?
I think we do. And show an error toast.

PR #261, opened 2019-04-23
Update Pendo object

ryelo:


ryelo:
@iphands I'm not sure if the org_id and account_number needs to be flipped. If it is, I can update.

src/js/analytics.js
```diff
@@ -13,14 +13,14 @@ export default (user) => {
 
     const pendoConf = {
         visitor: {
-            id: user.id,
-            internal: user.is_internal,
-            lang: user.locale
+            id: user.identity.internal.account_id,
```
iphands: Can we just pass in user.identity and user.internal as (data, internal) to simplify?

PR #265, opened 2019-04-24
Change FAQ button to dropdown

karelhala:
Peek 2019-04-24 11-25

Screenshot from 2019-04-24 13-37-13


src/js/App/Header/InsightsAbout.scss
```diff
@@ -1,10 +1,14 @@
...
 }
 
 .pf-c-about-modal-box__hero {
     --pf-c-about-modal-box__hero--sm--BackgroundImage: url('/../../build/assets/images/hero-image.jpg');
-} 
\ No newline at end of file
+}
+
+.ins-c-page__about--modal dl {
+    grid-row-gap: calc(var(--pf-c-content--dl--RowGap) / 2) !important;
```
ryelo: Can we just change the specificity of this instead of using `!important`?

karelhala: Yeah good point. This was partial solution and it got trough. Changing it to global spacer small.

PR #266, opened 2019-04-24
Disable Insights > Settings > Sources only on Stable

iphands:
Had to change subitems from lambda to full function then disable the nav subitem based on window.location.pathname containing /beta

### DIFF
--- a/src/js/nav/globalNav.js
+++ b/src/js/nav/globalNav.js
@@ -32,7 +32,8 @@ export const grouppedNav = {
                     {
                         id: 'sources',
                         title: 'Sources',
-                        reload: 'settings/sources'
+                        reload: 'settings/sources',
+                        disabled: window.location.pathname.indexOf('/beta') === -1
                     }
                 ]
             }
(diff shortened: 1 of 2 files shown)


src/js/nav/globalNav.js
```diff
@@ -32,7 +32,8 @@ export const grouppedNav = {
                     {
                         id: 'sources',
                         title: 'Sources',
-                        reload: 'settings/sources'
+                        reload: 'settings/sources',
+                        disabled: window.location.pathname.indexOf('/beta') === -1
```
ryelo: You can just use `window.insights.chrome.isBeta()`

iphands: Word!

iphands: Oh wait no :(
Try it out. This stuff is built before window.insights is available

Hence this:
https://github.com/RedHatInsights/insights-chrome/blob/master/src/js/nav/globalNav.js#L164

Going to go ahead w/o that for now

PR #268, opened 2019-04-25
Enhancements/jwt tests

kruai:
I added a bunch of unit tests and brought the coverage of the JWT code to > 80%.
I also removed some priv/pub stuff and cleaned up a few things.


kruai:
I added --forceExit to Jest's execution for now, since it isn't apparent how to solve the issue where Jest doesn't exit post-completion.

karelhala:
I's kind of a bummer to use `--forceExit` but we don't have any other option for now. The way how to overcome this is to add broadcast channel bootstrap to function and import it from other files.

PR #274, opened 2019-04-29
removing cookie domain to use default

ryelo:


kruai:
Looks fine to me. I'll update my unit test code to not use domains as well.

PR #277, opened 2019-04-30
add logout redirect

ryelo:
This will redirect someone logging out to `/logout`.

It's not going to work entirely because there isn't a logout page ready to show, but you can see the redirect hit in the url.

When we log back into the page from `/logout`, we need to make sure to redirect them to `/` or else it's going to redirect them to `/logout` since that is where they came from.


iphands:
@rlong @kruai this is safe to take off of onhold now right?

rlong:
I think you may have the wrong person.




Cheers,


Richard Long




Get Outlook for Android







On Thu, May 2, 2019 at 10:46 PM +0100, "Ian Page Hands" <notifications@github.com> wrote:












@rlong @kruai this is safe to take off of onhold now right?

ryelo:
Hahaha ^

@iphands it should be good as soon as: https://github.com/RedHatInsights/landing-page-frontend/pull/37 is merged

ryelo:
@iphands Good to go now

PR #279, opened 2019-04-30
Rename Vulnerabilities to Vulnerability

kruai:
Just renaming the instances of `Vulnerabilities` to `Vulnerability`. I'm not renaming the references to frontend-components files yet bc that will take some coordination and I don't want to break everyone's frontend stuff right now


src/js/async-loader.js
```diff
@@ -11,7 +11,7 @@ const asyncLoader = (imports, callback = ([cmp]) => cmp) => {
 export default {
     loadVulnerabilities: asyncLoader([
         import('@red-hat-insights/insights-frontend-components/components/Vulnerabilities'),
-        import('./vulnerabilitiesStyles')
+        import('./vulnerabilityStyles')
```
ryelo: Do we want to change this?

kruai: I renamed the file as well, so nothing should be breaking. Is renaming it an issue?

PR #282, opened 2019-05-01
Adds left aligned fedora hero background image

ryelo:
Needed for landing page for now. Could be used in other apps.


kruai:
What does this change? I pulled, ran `npm run build` on chrome, and started it up in incognito, and it looks the same as in Prod

ryelo:
@kruai just adds an image to the assets. There should be no visual change. This image might be referenced from multiple apps, so I'm going to host it here.

PR #287, opened 2019-05-02
Redirect to eval page if account number is undefined

karelhala:


iphands:
@karelhala I made some changes...
I think it is 100% true that we cannot force this on /openshift.
I also think we should not force this on / or /404 or /random.

Lets talk about it though.

iphands:
Closing per discussion in Scrum.
We are not doing this, instead the modal display link to the eval page.... thus this is obsolete

PR #291, opened 2019-05-03
Force a kc.login() whenever getUser is called with no cookie

iphands:
The cs_jwt cookie is now timebombed. It expires the same second as the JWT.

Any time someone calls getUser() on one of the authenticated pages we should
call kc.login() for them.


iphands:
@ryelo I didnt like importing Utils into jwt... as it is supposed to be its own module one day, but we can undo that later.

PR #293, opened 2019-05-05
Use correct scope for offline access token

elad661:
The Keycloak documentation says the correct scope for offline access
tokens is offline_access
https://www.keycloak.org/docs/2.5/server_admin/topics/sessions/offline.html

When the scope is incorrect, we still get a token, but it's not the
offline token we need, and it expires after a while.

Changing the scope to offline_access should fix this problem.

cc: @iphands @karelhala


karelhala:
This should be available in beta on all envs after we merge this PR.

elad661:
Thanks for the quick review :)

PR #297, opened 2019-05-08
updating sidenav nav to be consistent

ryelo:


psav:
> Looking good! Thanks this is much easier to read and understand.
It also works :)

PR #298, opened 2019-05-09
updating lint to use react rules, fixing errors

ryelo:
Fixes some tech debt.

Linter wasn't using react recommended rules. Updated so that it does and fixed all associated errors.

Killed the separator above the "documentation" link in favor of the Patternfly supplied one.

### DIFF
--- a/src/js/App/Header/index.js
+++ b/src/js/App/Header/index.js
@@ -11,7 +11,7 @@ export function unauthed() {
 }
 
 ;
-
+// eslint-disable-next-line
 export default () => (
     <Fragment>
         <Brand />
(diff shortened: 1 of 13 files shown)


src/js/App/Header/index.js
```diff
@@ -11,7 +11,7 @@ export function unauthed() {
 }
 
 ;
-
+// eslint-disable-next-line
 export default () => (
```
karelhala: You should do something lik
```jsx
const Header = () => ( <Fragment>
  <Brand/>
  <Tools />
</Fragment>
);

export default Header;
```

ryelo: updated!

PR #302, opened 2019-05-16
Start using frontend components monorepo

karelhala:
This PR switches old frontend components for new monorepo components.


ryelo:
@karelhala can we update the deps again to get the new inventory table in this PR?

PR #309, opened 2019-05-29
Add Sentry

ryelo:
Default configuration should capture all fatal events in all apps using chrome

If other apps want to roll their own errors, they have the ability to do so.

### DIFF
--- a/src/js/chrome.js
+++ b/src/js/chrome.js
@@ -1,5 +1,6 @@
 import auth from './auth';
 import analytics from './analytics';
+import sentry from './sentry';
 import { bootstrap, chromeInit }   from './entry';
 
 // start auth asap
@@ -8,7 +9,7 @@ const libjwt = auth();
 function noop () {}
 
 libjwt.initPromise.then(() => {
-    libjwt.jwt.getUserInfo().then(analytics).catch(noop);
+    libjwt.jwt.getUserInfo().then(analytics && sentry).catch(noop);
 });
 
 window.insights = window.insights || {};
--- a/src/js/sentry.js
+++ b/src/js/sentry.js
@@ -1,0 +1,42 @@
+const API_KEY = 'https://80e5a70255df4bd3ba6eb3b4bfebc58c@sentry.io/1466891'
+
+// - Sentry Options
+// dsn: key
+// environment: logs PROD or PROD Beta for filtering
+// maxBreadcrumbs, if there is an error, trace back up to (x) lines if needed
+// attachStacktrace: attach the actual console logs
+// debug: will attempt to print out useful debugging information if something goes wrong with sending the event
+// sampleRate: 0.0 to 1.0 - percentage of events to send (1.0 by default)
+
+function initSentry() {
+    let betaCheck = (window.location.pathname.split('/')[1] === 'beta' ? ' Beta' : '');
+
+    Sentry.init({
+        dsn: API_KEY,
+        environment: `PROD${betaCheck}`,
+        maxBreadcrumbs: 50,
+        attachStacktrace: true,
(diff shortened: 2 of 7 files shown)


src/js/sentry.js
```diff
@@ -0,0 +1,44 @@
...
+        environment: `PROD${betaCheck}`,
+        maxBreadcrumbs: 50,
+        attachStacktrace: true,
+        debug: true
+    });
+}
+
+/* eslint-disable camelcase */
+function sentryUser(user) {
+    // TODO: Add request_id to this
```
kruai: Should the request_id be a part of this PR, or is it OK to leave it off for now?

src/js/sentry.js
```diff
@@ -0,0 +1,44 @@
+import * as Sentry from '@sentry/browser';
+
+const API_KEY = 'https://80e5a70255df4bd3ba6eb3b4bfebc58c@sentry.io/1466891';
```
kruai: Is it OK for this to be in the code?

src/js/sentry.js
ryelo: It's being left off for now. This is something to circle back later.

src/js/sentry.js
ryelo: Not sure. We're exposing the api key for the pendo stuff in `analytics.js`, so I thought it should be fine? We can always wrap it up and do it somewhere else.

kruai: If it's already exposed, I guess we might as well do the same here. I'm not sure how sensitive the api key is, but if it is, we'll probably want to look at injecting it for both sentry and pendo.

src/js/chrome.js
```diff
@@ -8,7 +9,7 @@ const libjwt = auth();
 function noop () {}
 
 libjwt.initPromise.then(() => {
-    libjwt.jwt.getUserInfo().then(analytics).catch(noop);
+    libjwt.jwt.getUserInfo().then(analytics && sentry).catch(noop);
```
karelhala: I don't think that this is going to work. It's going to run analytics and then dies. To fix this you should do something like
```JS
libjwt.jwt.getUserInfo().then((...data) => {
  analytics(...data);
  sentry(...data);
}).catch(noop);
```
This way any data returned ffrom getUser info will be passed in same way.

src/js/sentry.js
```diff
@@ -0,0 +1,44 @@
...
+// - Sentry Options
+// dsn: key
+// environment: logs PROD or PROD Beta for filtering
+// maxBreadcrumbs, if there is an error, trace back up to (x) lines if needed
+// attachStacktrace: attach the actual console logs
+// debug: will attempt to print out useful debugging information if something goes wrong with sending the event
+// sampleRate: 0.0 to 1.0 - percentage of events to send (1.0 by default)
+
+function initSentry() {
+    let betaCheck = (window.location.pathname.split('/')[1] === 'beta' ? ' Beta' : '');
```
karelhala: Nitpick: No need to assign to variable if it's going to be used only in one place.

src/js/chrome.js
ryelo: Did you test the code locally? At least for me, it was hitting both analytics and sentry using this method.

karelhala: Oh `analytics` returns true? In that case we are good with shortcutting.

src/js/sentry.js
ryelo: @karelhala Sentry won't init with the window function inside the object. You have to get the location before inserting it into the object like it is now.

karelhala: Oh okay.

PR #312, opened 2019-06-03
Enhancements/global nav source of truth

kruai:
This allows Chrome to automatically generate the global nav data on the build step. This is dependent on https://github.com/RedHatInsights/cloud-services-config/pull/5 and should not be merged until that one is.


kruai:
Looks like I need to give myself a crash course in Reducers. I'll learn about those and then take a look at that failing test.

src/js/nav/buildNav.js
```diff
@@ -0,0 +1,74 @@
...
+            throw err;
+        }
+    });
+
+}
+
+// Returns a list of routes/subItems owned by an app
+function getRoutesForApp(app, masterConfig) {
+    let routes = [];
+    if (_.has(app, 'sub_apps')) {
```
karelhala: No need to use lodash's has. You can use buildin `hasOwnProperty`

src/js/nav/buildNav.js
```diff
@@ -0,0 +1,74 @@
+const yaml = require('js-yaml');
+const axios = require('axios');
+const _ = require('lodash');
```
karelhala: No need to import entire lodash, just those things that you are using
```
const { has } = require('lodash');
```

Given that only `has` is used no need to import lodash at all.

package.json
```diff
@@ -5,12 +5,13 @@
...
     "start": "NODE_ENV='development' npm run watch",
-    "build": "npm-run-all build:images build:images-chrome build:js      build:pug      build:sass build:fonts build:Fafonts",
-    "build:beta": "npm-run-all build:images build:images-chrome build:js:beta build:pug:beta build:sass build:fonts build:Fafonts",
+    "build": "npm-run-all build:images build:images-chrome build:js      build:pug      build:sass build:fonts build:Fafonts build:sidenav",
+    "build:beta": "npm-run-all build:images build:images-chrome build:js:beta build:pug:beta build:sass build:fonts build:Fafonts build:sidenav",
     "build:js": "parcel build src/js/chrome.js -d build/js --public-url /apps/chrome/js",
     "build:js:beta": "parcel build src/js/chrome.js -d build/js --public-url /beta/apps/chrome/js",
     "build:images": "cpx 'node_modules/@patternfly/patternfly/assets/images/*' build/assets/images",
     "build:images-chrome": "cpx 'static/images/*' build/assets/images",
+    "build:sidenav": "node src/js/nav/buildNav.js",
```
karelhala: I don't think this is a good idea. It would be much better to download the yaml on fly and generate object from it.

src/js/nav/buildNav.js
```diff
@@ -0,0 +1,74 @@
...
+const fs = require('fs');
+
+axios.get('https://raw.githubusercontent.com/RedHatInsights/cloud-services-config/enhancements/chrome-nav/main.yml')
+.then(response => yaml.safeLoad(response.data)).then(data => buildNavFromConfig(data));
+
+function buildNavFromConfig(masterConfig) {
```
karelhala: This would be better to dispatch new action to redux and store current nav based on the app instead of storing it in JSON file.

src/js/nav/buildNav.js
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+}
+
+// Returns a list of routes/subItems owned by an app
+function getRoutesForApp(app, masterConfig) {
+    let routes = [];
+    if (_.has(app, 'sub_apps')) {
+        app.sub_apps.forEach((subApp => {
+            let subAppData = getAppData(subApp.id, 'subItems', masterConfig);
```
karelhala: What parts of `getAppData` are you actually interested in? And what is the shape of some object? It would be much better to use something different than delete.

src/js/nav/buildNav.js
```diff
@@ -0,0 +1,74 @@
...
+            routes.push(subAppData);
+        }));
+    }
+
+    return routes;
+}
+
+// Gets the app's data from the master config, if it exists
+function getAppData(appId, propName, masterConfig) {
+    let appList = masterConfig.filter(x => x.id === appId);
```
karelhala: This can be rewritten to simpler approach:
```JS
const app = masterConfig.find(item => item.id === appId);
if (app && app.hasOwnProperty('sub_apps')) {
  return {
    ...app, // here we could list what should be returned to later use instead of deletes
    [propName]: getRoutesForApp(app, masterConfig)
  }
}
```
Implicit undefined return so we don't have to write `return ;`

karelhala:
> Looks like I need to give myself a crash course in Reducers. I'll learn about those and then take a look at that failing test.

The failing test is that JS automatically imports `.js` files but global nav is now json so you need to update that test to import `../nav/globalNav.json`

package.json
kruai: I assumed creating it on the build step would be the way to go in order to make it faster. Wouldn't generating it on the fly slow it down each time we navigate to a page?

src/js/nav/buildNav.js
kruai: OK. Yeah, now that we have even more fields added to it, I'm going to have to make that more precise instead of deleting things afterwards.

kruai:
After discussing with Karel, we're changing strategies. Now, we're going to request the yaml file on each page load, cache it in a similar way to entitlements-api, and build the JS object. No more building JSON on build.

src/js/entry.js
```diff
@@ -26,20 +29,27 @@ export function chromeInit(libjwt) {
     const { store, middlewareListener, actions } = spinUpStore();
 
     // public API actions
-    const { identifyApp, appNav, appNavClick, clearActive } = actions;
-    libjwt.initPromise.then(() => {
+    const { identifyApp, appNav, appNavClick, clearActive, chromeNavUpdate } = actions;
+
+    const jwtAndNavResolver = libjwt.initPromise.then(() => {
```
karelhala: Looking at the function you gotta rewrite it because you will get race conditions. It would be better to use async/await but I am not sure if we can use it.
```JS
const jwtAndNavResolver = libjwt.initPromise
.then(() => libjwt.jwt.getUserInfo.then()) // you have to wait for getUserInfo to resolve here
.then((user) => {
  actions.userLogIn(user);
  loadChrome(user);
})
.then(() => sourceOfTruth(ibjwt.jwt.getEncodedToken).then()) // sourceOfTruth is Promise as well
.then(({ data }) => chromeNavUpdate(loadNav(data)))
.catch(() => {
  if (allowUnauthed()) {
    loadChrome(false);
  }
});
```

src/js/nav/sourceOfTruth.js
```diff
@@ -0,0 +1,24 @@
...
+    const cache = setupCache({
+        store,
+        maxAge: 10 * 60 * 1000 // 10 minutes
+    });
+
+    const instance = axios.create({ adapter: cache.adapter });
+
+    instance.interceptors.response.use((response) => response.data || response);
+
+    return get('https://raw.githubusercontent.com/'
```
karelhala: You have to use `instance`'s get in order to use cache. `instance.get('https://raw.githubusercontent.com/RedHatInsights/cloud-services-config/master/main.yml')`

iphands:
Ouch, Asa you are planning on fetching the yaml dynamically, and from git hub? It's cached sure... But that seems messy

kruai:
@iphands - you, Ryan, Karel and I discussed this a while back. My original implementation still fetched it from github, but created the sidenav object in the build step. Karel thought it would be better to fetch it and cache it with the user data, so that if the sidenav data changed, the user would see them after a bit without us having to re-deploy. After we all discussed, the general consensus was that this was a better approach.

If you would prefer a different way of doing this, let's discuss this again in scrum today.

PR #321, opened 2019-06-12
Jwtauth bugfix

yashnalla:


yashnalla:
Bug fix paired with: https://github.com/RedHatInsights/landing-page-frontend/pull/56

src/js/jwt/jwt.js
```diff
@@ -248,7 +264,7 @@ exports.expiredToken = () => { logout(); };
 
 // Broadcast message to refresh tokens across tabs
 function refreshTokens() {
-    authChannel.postMessage({ type: 'refresh' });
+    return authChannel.postMessage({ type: 'refresh' });
```
yashnalla: have to remove this

src/js/jwt/jwt.js
```diff
@@ -235,6 +237,20 @@ exports.getUserInfo = () => {
         }
     });
 };
+// Challenge auth and login if the user could be logged in, but in an unauth state
+exports.challengeAuth = () => {
+    log('Challenging Auth');
+    //return refreshTokens();
+    priv.keycloak.login({onLoad : "login-required"})
```
karelhala: You should return this promise, otherwise this is going to return `undefined`

src/js/jwt/jwt.js
```diff
@@ -218,10 +219,11 @@ function loginAllTabs() {
 /*** User Functions ***/
 // Get user information
 exports.getUserInfo = () => {
-    log('Getting User Information');
+    log('Getting User Information CHANGES!');
```
ryelo: Do we need this `CHANGES!` in the log?

src/js/jwt/jwt.js
```diff
@@ -235,6 +237,20 @@ exports.getUserInfo = () => {
         }
     });
 };
+// Challenge auth and login if the user could be logged in, but in an unauth state
+exports.challengeAuth = () => {
+    log('Challenging Auth');
+    //return refreshTokens();
```
ryelo: I think this comment isn't needed

src/js/jwt/jwt.js
yashnalla: No i just totally forgot to remove it. I've already removed it.

src/js/jwt/jwt.js
yashnalla: Yep another thing I've removed.

src/js/jwt/jwt.js
yashnalla: fixed

yashnalla: fixed

package.json
```diff
@@ -105,11 +105,10 @@
     "@patternfly/react-icons": "^3.7.5",
     "@redhat-cloud-services/entitlements-client": "^1.0.4",
     "@redhat-cloud-services/frontend-components": "0.0.5",
-    "@redhat-cloud-services/frontend-components-inventory": "0.0.11",
+    "@redhat-cloud-services/frontend-components-inventory": "0.0.12",
     "@redhat-cloud-services/frontend-components-remediations": "0.0.3",
     "@redhat-cloud-services/frontend-components-utilities": "0.0.5",
-    "@sentry/browser": "^5.3.0",
```
karelhala: Why is this dep removed? I don't think you should update any of these dependencies in this PR.

ryelo: Yeah this dep needs to be added back so Sentry works correctly.

yashnalla: I have no idea how that got changed? I assume at some point I did something that updated dependencies but I don't know how sentry got removed. I went ahead and added it back but should I also go in and change all the other dependency versions back to what they were before they were updated?

ryelo: Yeah I would change them back to how they were, and then `npm install` again just so we can keep the scope of this PR limited to jwt auth stuff and not a dep update.

yashnalla: got it

PR #325, opened 2019-06-20
update travis for nightly build

ryelo:


ryelo:
I could also just slap this in the beta version instead of namespacing the stage this way. Doesn't really matter to me though.

PR #331, opened 2019-06-26
added keycloack config option

yashnalla:


yashnalla:
There are issues with the logout routing see: https://github.com/RedHatInsights/insights-chrome/pull/321 that i was able to locally solve by substituting /logout.html for /logout

src/js/jwt/jwt.js
```diff
@@ -219,7 +220,7 @@ function loginAllTabs() {
 /*** User Functions ***/
 // Get user information
 exports.getUserInfo = () => {
-    log('Getting User Information');
+    log('Getting User Information changes!!!');
```
iphands: Was the "changes!!!" string left in by mistake?

src/js/jwt/jwt.js
```diff
@@ -237,6 +238,12 @@ exports.getUserInfo = () => {
     });
 };
 
+// Challenge auth and login if the user could be logged in, but in an unauth state
+exports.challengeAuth = () => {
```
iphands: I hate the name here. That is not a challenge. An auth challenge is whenever a user *must* re-enter their username/pasword.

Doesnt it make way more sense to just export this as login?

src/js/jwt/jwt.js
yashnalla: Uhh im still working on this branch, so it was left in purposefully.

src/js/jwt/jwt.js
yashnalla: Same as above. I was just gonna rename the function login

yashnalla: Well not login cause thats taken but something like checkAuth

PR #332, opened 2019-07-02
Add ruledev to navigation

karelhala:


src/js/nav/globalNav.js
```diff
@@ -162,6 +162,16 @@ export const grouppedNav = {
                 default: true
             }
         ]
+    },
+    ruledev: {
+        title: 'Rule Development',
+        routes: [
+            {
+                id: 'ruledev',
```
kruai: If we have a sub-app under `ruledev`, wouldn't it make sense to name the ID something else? Like how the `insights` app has `actions` and `rules`?

karelhala: Yeah, you are correct ruleDev will be part of staging.

PR #334, opened 2019-07-09
Add page safe to iqeEnablement

psav:


src/js/iqeEnablement.js
```diff
@@ -62,6 +62,13 @@ export default {
 
         return xhrResults.length > 0 || fetchResults.length > 0;
     },
+    isPageSafe: () => {
```
karelhala: No need to return `false`/`true` we can just return the value from `querySelector().length > 0`
```JS
isPageSafe: () => document.querySelectorAll('[data-ouia-safe=false]').length > 0
```

PR #344, opened 2019-07-29
add error handling to getToken()

ryelo:


iphands:
Also the original bug says:
> Our code calls insights.chrome.auth.getToken() for every API request, see here: https://gitlab.cee.redhat.com/service/uhc-portal/blob/master/src/services/apiRequest.js

That team is not using getUserInfo afaict

ryelo:
@iphands added another check. If you'd rather me not expose the `isTokenExpired()`, I can remove it. The keycloak function returns true if expired.

ryelo:
@iphands If you have a second, can you take another look at this to make sure it hits the AC?

src/js/jwt/jwt.js
```diff
@@ -98,6 +99,8 @@ exports.init = (options) => {
     options.url = insightsUrl(((options.routes) ? options.routes : DEFAULT_ROUTES));
     options.promiseType = 'native';
 
+    options.timeSkew = 10;
```
iphands: Why are we setting this for every user?

src/js/jwt/jwt.js
```diff
@@ -294,7 +297,17 @@ function setCookieWrapper(str) {
 // Encoded WIP
 exports.getEncodedToken = () => {
     log('Getting encoded token');
-    return (priv.keycloak.token);
+    // Check if the token is expired, if expired - report to sentry
+    if (!exports.isTokenExpired()) {
+        return (priv.keycloak.token);
+    } else {
```
iphands: What do we return if the token is expired?
Is undefined what we want?

src/js/jwt/jwt.js
ryelo: This is just a default. If Keycloak can't determine the actual time skew, it will use this value. Currently, when I run this, the skew is actually 0. There are many functions that we aren't using in here (yet?) that require a default time skew.

src/js/jwt/jwt.js
ryelo: I'm not sure about the answer to that. I'm tempted to say to try and call a refresh and then send to a login page if that fails.

PR #346, opened 2019-07-31
Update the custom release to separate the environments

ryelo:


ryelo:
@karelhala You're totally right. Deleted the file!

PR #347, opened 2019-07-31
Get source of truth from current host

kruai:
Update to Chrome to get the source of truth from the current host.
It doesn't seem to work for me right now. Also, is there a better way to do this?

### DIFF
--- a/src/js/nav/sourceOfTruth.js
+++ b/src/js/nav/sourceOfTruth.js
@@ -19,6 +19,13 @@ module.exports = (cachePrefix) => {
 
     instance.interceptors.response.use((response) => response.data || response);
 
-    return instance.get('https://raw.githubusercontent.com/'
-    + 'RedHatInsights/cloud-services-config/master/main.yml');
+    // TODO: Add prefix (/beta) depending on environment
+    let prefix = '';
+    if (window.location.pathname.indexOf('/beta')) {
+        prefix = '/beta';
+    }
+
+    console.log('Getting index from ' + window.location.host + prefix + '/config/main.yml');
+
+    return instance.get(window.location.host + prefix + '/config/main.yml');
 };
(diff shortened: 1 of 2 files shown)


src/js/nav/sourceOfTruth.js
```diff
@@ -19,6 +19,11 @@ module.exports = (cachePrefix) => {
...
 
-    return instance.get('https://raw.githubusercontent.com/'
-    + 'RedHatInsights/cloud-services-config/master/main.yml');
+    // TODO: Add prefix (/beta) depending on environment
+    let prefix = '';
+    if (window.location.pathname.indexOf('/beta')) {
+        prefix = '/beta';
+    }
+
+    return instance.get(window.location.host + prefix + '/config/main.yml');
```
karelhala: Use `window.location.origin` to include protol as well.

src/js/nav/sourceOfTruth.js
```diff
@@ -19,6 +19,11 @@ module.exports = (cachePrefix) => {
 
     instance.interceptors.response.use((response) => response.data || response);
 
-    return instance.get('https://raw.githubusercontent.com/'
-    + 'RedHatInsights/cloud-services-config/master/main.yml');
+    // TODO: Add prefix (/beta) depending on environment
+    let prefix = '';
+    if (window.location.pathname.indexOf('/beta')) {
```
ryelo: This logic needs to flip

karelhala: It needs update actually
`if (window.location.pathname.indexOf('/beta') !== -1) {`

ryelo: Yeah, I'd probably roll with that pattern since we're using that in other places too.

src/js/nav/sourceOfTruth.js
kruai: Oops - just pushed a commit that added the protocol, but your solution is cleaner!

PR #354, opened 2019-08-08
Support for disabled_on_prod flag

kruai:
Add support for the disabled_on_prod flag. Makes it so the nav doesn't add an entry for an app if the app is supposed to be disabled on prod, and the hostname indicates that we're on prod.

The logic feels a little clunkier than I'd like, but I'm not immediately seeing a way to clean this up and achieve the same result. Let me know if you see a way to simplify this!


kruai:
> Looks good!
> Do you need the if statement around `appData.disabled_on_prod = app.disabled_on_prod;` though?
> Would it still work if you just set it regardless?

It would still function, but every app that didn't have that flag would instead have a null "disabled_on_prod" property, and I didn't want to clutter up the data.

PR #357, opened 2019-08-13
Log browser width in sentry

ryelo:


ryelo:
@kruai It's mainly to see if behavior changes on mobile devices versus that of a desktop. 

Like maybe someone forgot to remove a button from mobile that sends them to a 404, but the desktop version looks fine.

PR #361, opened 2019-08-16
Adding migrations bundle

ryandillinfelton:
Adding a new bundle (migration) and a tile on the landing page for it


src/js/App/Header/InsightsAbout.js
```diff
@@ -73,7 +73,9 @@ class InsightsAbout extends Component {
                 { name: 'Compliance', path: 'apps/compliance/app.info.json', version: 'N/A' },
                 { name: 'Cost Management', path: 'apps/cost-management/app.info.json', version: 'N/A' },
                 { name: 'Insights', path: 'apps/insights/app.info.json', version: 'N/A' },
-                { name: 'System Comparison', path: 'apps/drift/app.info.json', version: 'N/A' }
+                { name: 'System Comparison', path: 'apps/drift/app.info.json', version: 'N/A' },
+                //concerned about this bit
+                { name: 'Migration', path: 'migrations/migration-analytics/app.info.json', version: 'N/A' }
```
ryelo: Let's call this "Migrations" just for consistency

src/js/jwt/insights/user.js
```diff
@@ -5,7 +5,8 @@ const pathMapper = {
     rhel: 'smart_management',
     hybrid: 'hybrid_cloud',
     insights: 'insights',
-    openshift: 'openshift'
+    openshift: 'openshift',
+    migration: 'migration'
```
ryelo: Same thing here

src/js/App/Header/InsightsAbout.js
ryandillinfelton: fixed

src/js/jwt/insights/user.js
ryandillinfelton: fixed

src/js/App/Header/InsightsAbout.js
```diff
@@ -73,7 +73,9 @@ class InsightsAbout extends Component {
                 { name: 'Compliance', path: 'apps/compliance/app.info.json', version: 'N/A' },
                 { name: 'Cost Management', path: 'apps/cost-management/app.info.json', version: 'N/A' },
                 { name: 'Insights', path: 'apps/insights/app.info.json', version: 'N/A' },
-                { name: 'System Comparison', path: 'apps/drift/app.info.json', version: 'N/A' }
+                { name: 'System Comparison', path: 'apps/drift/app.info.json', version: 'N/A' },
+                //concerned about this bit
+                { name: 'Migrations', path: 'migrations/migration-analytics/app.info.json', version: 'N/A' }
```
ryelo: I think this path is actually going to be `apps/migration-analytics/app.info.json`

They're in beta right now, so you can see the path:
https://ci.cloud.paas.psi.redhat.com/beta/apps/migration-analytics/app.info.json

ryandillinfelton: Thanks! I was concerned about that bit

kruai:
In the Source of Truth, the ID for the Migration Analytics app is `migration-analytics`; in this PR, it's `migrations`. Do you know if they prefer one way or the other?

src/js/App/Header/InsightsAbout.js
```diff
@@ -73,7 +73,8 @@ class InsightsAbout extends Component {
                 { name: 'Compliance', path: 'apps/compliance/app.info.json', version: 'N/A' },
                 { name: 'Cost Management', path: 'apps/cost-management/app.info.json', version: 'N/A' },
                 { name: 'Insights', path: 'apps/insights/app.info.json', version: 'N/A' },
-                { name: 'System Comparison', path: 'apps/drift/app.info.json', version: 'N/A' }
+                { name: 'System Comparison', path: 'apps/drift/app.info.json', version: 'N/A' },
+                { name: 'Migrations', path: 'apps/migration-analytics/app.info.json', version: 'N/A' }
```
ryandillinfelton: @kruai should this line instead be
`{ name: 'Migration Analytics', path: 'apps/migration-analytics/app.info.json', version: 'N/A' }` ?

kruai:
> > In the Source of Truth, the ID for the Migration Analytics app is `migration-analytics`; in this PR, it's `migrations`. Do you know if they prefer one way or the other?
> 
> 'migration-analytics' should still be the name of the one app in the new 'migrations' bundle. If there is a spot where I should be referring to the app name instead of the bundle name then that's a mistake, not an attempt at renaming the app itself.

Ah, I see what you mean now that I'm taking a look at the cloud-services-config PR.

src/js/App/Header/InsightsAbout.js
kruai: Yes, it probably should!

PR #365, opened 2019-08-26
Update all deps to newest version

karelhala:
Tested on inventory screen and nothing broke, so we should be good to go.


src/js/jwt/insights/url.js
```diff
@@ -1,30 +1,13 @@
...
-        // If the URL isn't found and is not in the default, default the user to QA
-        if (!urlFound) {
-            log('SSO: url not found, defaulting to QA');
-            log('ENV: qa');
-            return ('https://sso.qa.redhat.com/auth');
-        }
-    }
-};
+    }) && values.sso) || acc;
+}, 'https://sso.qa.redhat.com/auth');
```
ryelo: Can we add those log messages back here?

karelhala: Good catch!

karelhala:
Closing in favor of #385

PR #370, opened 2019-09-09
Fixes/add check sso silently

yashnalla:


src/js/jwt/jwt.js
```diff
@@ -98,6 +98,10 @@ exports.init = (options) => {
 
     options.url = insightsUrl(((options.routes) ? options.routes : DEFAULT_ROUTES));
     options.promiseType = 'native';
+    options.onLoad = 'check-sso';
+    //options.silentCheckSsoRedirectUri = `https://google.com`;
+    options.silentCheckSsoRedirectUri = `https://${window.location.host}/logout.html`;
```
karelhala: If I understood the docs correctly this should be `options.silentCheckSsoRedirectUri = `https://${window.location.host}/silent-check-sso.html`;`

PR #372, opened 2019-09-11
adding new dark navigation

ryelo:
This updates some PF deps, so please test accordingly.

Screen Shot 2019-09-11 at 7 05 19 PM
Screen Shot 2019-09-11 at 1 22 10 PM


src/sass/partials/_sidebar.scss
```diff
@@ -1,3 +1,5 @@
+$dark-nav-border: #4f5255;
```
karelhala: Can you use PF var `--pf-global--BackgroundColor--dark-400` instead of direct value?

ryelo: Yeah!

PR #377, opened 2019-09-17
[WIP] beta opt in

ryandillinfelton:
Before clicking the opt in button
Screenshot_20190917_121035

After clicking opt in button (or on page load when opted in)
Screenshot_20190917_121112

Someone please explain why the primary button doesn't look how it should

### DIFF
--- a/src/js/App/Header/Beta.js
+++ b/src/js/App/Header/Beta.js
@@ -1,53 +1,59 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import { Tooltip, Button } from '@patternfly/react-core';
 import cookie from 'js-cookie';
-import "../../../sass/beta.scss"
+//import '../../../sass/beta.scss';
 
-const Beta = () => {
-    if (window.location.href.includes('/beta')) {
-        return (
-            <div className={'ins-c-page__beta '}>
-                <i className='fas fa-flask'></i>
-                <Tooltip
-                    content={
-                        <div>This is a beta environment and contains Technology Preview
-                        features that are not supported by Red Hat production service-level
-                        agreements (SLAs) and might not be functionally complete. Red Hat
-                        recommends not using them for production.</div>
+export default class Beta extends React.Component {
+    constructor() {
+        super();
+        this.state = {
+            betaDefault: cookie.get('betaDefault')
+        };
+    }
+
+    setBetaDefault = () => {
+        cookie.set('betaDefault', true, { expires: 9999 });
+        this.setState({ betaDefault: true });
+    }
+
+    goToStable() {
+        event.preventDefault();
+        cookie.remove('betaDefault');
+        window.location = window.location.href.replace('/beta', '');
+    }
+
+    render() {
+        if (window.location.href.includes('/beta')) {
+            return (
+                <div className={'ins-c-page__beta '}>
+                    <span>
+                        <i className='fas fa-flask'></i>
+                        <Tooltip
+                            content={
+                                <div>
+                                    This is a beta environment and contains Technology Preview
+                                    features that are not supported by Red Hat production service-level
+                                    agreements (SLAs) and might not be functionally complete. Red Hat
+                                    recommends not using them for production.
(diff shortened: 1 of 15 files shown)


src/js/App/Header/Beta.js
```diff
@@ -0,0 +1,44 @@
...
+                 agreements (SLAs) and might not be functionally complete. Red Hat
+                 recommends not using them for production. See fore information here.</div>
+            }
+        >
+            <p>Insights & Cloud Management Services Beta.</p>
+        </Tooltip>
+        <p><a onClick={ goToStable }> Take me to Stable</a>.</p>
+        <Button
+            variant="primary" id="beta_button"
+            className={(cookie.get('betaDefault') ? 'ins-c-button_hidden' : '')}
```
ryelo: Instead of using a classname to hide something, just render it conditionally

src/js/App/Header/Beta.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Tooltip, Button } from '@patternfly/react-core';
+import cookie from 'js-cookie';
+
+const Beta = () => (
+    <div className={'ins-c-page__beta ' + (window.location.href.includes('/beta') ? '' : 'ins-c-page_beta-hidden')}>
```
ryelo: Can you do conditional rendering here?

just check to see if it's /beta and render it instead of setting a classname that will hide something.

src/js/entry.js
```diff
@@ -183,3 +185,17 @@ function loadChrome(user) {
         }
     );
 }
+
+function checkBetaPreference() {
+    if (cookie.get('betaDefault')) {
+        const splitted = window.location.href.split('/') ;
```
ryelo: Instead of this, stay consistent with how we do it in other places: 
`window.location.pathname.split('/')[1] === 'beta'`

src/sass/chrome.scss
```diff
@@ -51,6 +51,7 @@ aside {
 // Header
 .pf-c-page__header {
     overflow: visible;
+    grid-template-areas: "beta beta beta" "header header header";
```
ryelo: We shouldn't need to use a grid template here

src/sass/chrome.scss
```diff
@@ -69,6 +70,33 @@ aside {
...
+        //color: var(--pf-c-button--m-primary--Color) !important;
+    }
+
+    i {
+        padding-right: 6px;
+    }
+}
+
+.ins-c-page_beta-hidden {
+    display: none !important;
```
ryelo: Try not to ever use `!important`. Also, this should be overwritten by the conditional rendering

src/sass/chrome.scss
```diff
@@ -69,6 +70,33 @@ aside {
     }
 }
 
+// Beta Banner
+.ins-c-page__beta {
+    background: $bg-betabar;
+    width: 100vw;
```
ryelo: Can we just use 100% here?

src/sass/chrome.scss
```diff
@@ -69,6 +70,33 @@ aside {
     }
 }
 
+// Beta Banner
```
ryelo: Can we just make this a different .scss file and import it into the beta banner component?

src/js/entry.js
ryandillinfelton: I'm having trouble recreating the issue that led me to switch from that to the current method right now. But essentially it seems to drop the final part of the url in some situations like 
a link like: cloud.redhat.com/inventory/accessDenied.ext?app=inventory would redirect to cloud.redhat.com/beta/inventory/ using the path method.

src/sass/chrome.scss
ryandillinfelton: What would you suggest using instead?

ryelo:
unfortunately, I'm going to close this because we're taking a different approach

ryelo:
it was a good run rip

PR #378, opened 2019-09-17
use NODE_ENV='production'

ryelo:


ryelo:
Ignore this. Going to rework it a different way

PR #379, opened 2019-09-20
WIP: Addin line logout

yashnalla:


src/js/nav/sourceOfTruth.js
```diff
@@ -25,5 +25,10 @@ module.exports = (cachePrefix) => {
         prefix = '/beta';
     }
 
-    return instance.get(window.location.origin + prefix + '/config/main.yml');
```
karelhala: Do not commit this file, this change is only for local development.

PR #381, opened 2019-09-23
move user management spot up one

ryelo:
In some browsers, using the arrow keys to navigate through the menu options don't work because of (what I think is) the fragment wrapping user management. Moving it up one solves this issue temporarily. I'll take a further look into this, but it may be a PF problem.


ryelo:
@karelhala that sounds great to me!

PR #382, opened 2019-09-23
Stretch drawer content and main page to entire height

karelhala:
If the content is not tall enough black background will be shown because drawer content and main root does not fill the entire content.

### Before
Screenshot from 2019-09-23 16-12-43

### After
Screenshot from 2019-09-23 16-12-00

### DIFF
--- a/src/sass/chrome.scss
+++ b/src/sass/chrome.scss
@@ -48,6 +48,11 @@ aside {
   --pf-m-emphasis--FontSize:                        var(--pf-global--FontSize--xl);
 }
 
+// Page layout styles
+.pf-c-drawer__content, #root.pf-c-page__main {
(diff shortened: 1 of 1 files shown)


karelhala:
@ryelo this can also be achieved trough flex config, do you think it would be better to use this instead of `height: 100%`?

ryelo:
@karelhala you should be able to wrap the content in the page with our `Main` component which should do it for you.

karelhala:
@ryelo it's already wrapped in main, I think that this was introduced by me in https://github.com/RedHatInsights/insights-chrome/pull/362

peekAFUQ8Z

Notice how `.pf-c-drawer__content` does not stretch over the entire page.

PR #383, opened 2019-09-23
Add native event to user navigation clicks

karelhala:


karelhala:
@AllenBW you might be interested in this issue as well.

PR #385, opened 2019-09-26
Update all deps

karelhala:
Follow up from https://github.com/RedHatInsights/insights-chrome/pull/365


ryelo:
Adding other reviewers here since I did push a commit. Just want another set of eyes.

kruai:
Tentatively looks good, but I probably shouldn't approve it until I fully test it myself. I'll do that after scrum

kruai:
After running it locally, it looks like this breaks the pre-chromed screen (the Pug stuff). Does anyone else see that?

PR #386, opened 2019-09-26
Move temp to replaced content

karelhala:
Since we now have app that is loaded before every application we can move temp to container that will be replaced by `RootApp`


kruai:
This has the same "flash" issue in the main content area that it has every time we change it to replacement:

1. The skeleton appears in the main content area
2. The skeleton disappears as the root node is replaced with empty content, and the main content area becomes grey
3. The word "Loading..." then appears in the main content area while the app loads
4. The app is finally loaded in the content area and looks normal again

ryelo:
@kruai that "loading" text actually comes from the application itself. Since we were talking about it earlier wrt the dashboard, you can change https://github.com/RedHatInsights/insights-dashboard/blob/3a8cd817f6e27e48be84f2698a61fa20d22a2846/src/Utilities/asyncComponent.js#L42 to whatever text you want and it'll be reflected in the page

ryelo:
I still think this needs to be visited, but it may be a combined app/platform effort

kruai:
> @kruai that "loading" text actually comes from the application itself. Since we were talking about it earlier wrt the dashboard, you can change https://github.com/RedHatInsights/insights-dashboard/blob/3a8cd817f6e27e48be84f2698a61fa20d22a2846/src/Utilities/asyncComponent.js#L42 to whatever text you want and it'll be reflected in the page

Yeah... I know it's happening in Dashboard, but I'm sure a few apps have it (since I believe it was in the starter-app). Regardless, the main content area (including skeleton) is cleared out while the application loads, so we'll have some amount of "flashing" regardless, which I know bothers some higher-ups

ryelo:
> so we'll have some amount of "flashing" regardless, which I know bothers some higher-ups

Yeah, agreed with this. We should see if there's a better option for that. Also, I'm going to make a PR in the starter app to at least use the skeleton instead of the loading text

karelhala:
@kruai that cleanup is happening because root app has no content, the best way would be to add spinner before root app loads, when root app load add spinner in main#root as well, once app loads it's up to application what they are going to show while the app loads.

karelhala:
@kruai how about now? I've copied over what was in temp to pug main dom element and into root app as well, both of them will be replaced once root app and actuall app loads

PR #397, opened 2019-10-10
Update FEC and PF react to new versions

karelhala:


.babelrc
```diff
@@ -11,7 +11,7 @@
   "plugins": [
     "@babel/plugin-proposal-class-properties",
     "transform-inline-environment-variables",
-    "@babel/plugin-syntax-dynamic-import",
+    ["@babel/plugin-syntax-dynamic-import", {}, "syntax-dynamic"],
```
karelhala: I had to add this line because parcel was refusing to build properly.

PR #398, opened 2019-10-11
Adding in line logout

yashnalla:
logoutNotification


src/js/jwt/jwt.js
```diff
@@ -212,6 +213,9 @@ function logout(bounce) {
 
     // Redirect to logout
     if (bounce) {
+        cookie.set('justLoggedOut', 'true', {
```
ryelo: we should follow our naming convention with this cookie, something like
`cs_loggedOut` and also scope similar to `cs_jwt`

src/sass/chrome.scss
```diff
@@ -59,6 +59,12 @@ aside {
     background: var(--pf-c-page__main-section--BackgroundColor);
 }
 
+.pf-c-alert.ins-c-alert__logout{
+    position: absolute;
+    top: var(--pf-c-page__header--MinHeight);
+    right: var(--pf-c-page__header-tools--MarginRight)
```
ryelo: semicolon here

src/js/App/Header/LogoutAlert.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    render() {
+        const { alertOneVisible } = this.state;
+        return (
+            <React.Fragment>
+                { alertOneVisible && (
+                    <Alert
+                        variant="success"
+                        title="Congratualtions you have successfully logged out!"
```
ryelo: Is this the string that the mocks have?

Otherwise, i'd just do `You have logged out.`

src/js/jwt/jwt.js
yashnalla: What exactly do you mean by scope? otherwise changed the name

src/js/App/Header/LogoutAlert.js
yashnalla: https://projects.engineering.redhat.com/browse/RHCLOUD-1810 the 'Congratulations' isn't but the rest is

src/sass/chrome.scss
yashnalla: done

src/js/jwt/jwt.js
ryelo: Ah nevermind - I was talking about scoping the domain of the cookie to cloud.redhat.com, but js-cookie does that by default.

src/js/App/Header/LogoutAlert.js
```diff
@@ -0,0 +1,52 @@
...
+
+    hideAlertOne = () => {
+        this.setState({ alertOneVisible: false });
+        cookie.set('cs_loggedOut', 'false');
+    }
+    componentDidMount() {
+        this.setTimer();
+    }
+
+    setTimer = () => {
```
karelhala: It's kinda interesting to hide the alert after 5s, but PF doesn't like auto closing alerts and since you are already using `AlertActionCloseButton` no need to close after some time as well. So we should remove this.

src/js/jwt/jwt.js
```diff
@@ -212,6 +213,9 @@ function logout(bounce) {
 
     // Redirect to logout
     if (bounce) {
+        cookie.set('cs_loggedOut', 'true', {
+            expires: 1 / 11520
```
karelhala: Can we add this to constant? Is this a 5s right?

src/js/App/Header/LogoutAlert.js
ryelo: PF has in their guidelines that autoclosing alerts are fine. They recommend 8 seconds instead of 5, so we can change it to that. But, the alert should close by itself according to the mocks.

src/js/jwt/jwt.js
ryelo: As per above, we should change this to 8s

src/js/App/Header/LogoutAlert.js
```diff
@@ -0,0 +1,52 @@
...
+        if (this._timer !== null) {
+            clearTimeout(this._timer);
+        }
+
+        // hide after `delay` milliseconds
+        this._timer = setTimeout(function() {
+            this.setState({ alertOneVisible: false });
+            this._timer = null;
```
karelhala: Instead of setting timer to `null` this should call `clearTimeout`

src/js/App/Header/LogoutAlert.js
```diff
@@ -0,0 +1,52 @@
...
+            alertOneVisible: cookie.get('cs_loggedOut') === 'true',
+            delay: 5000
+        };
+    }
+
+    hideAlertOne = () => {
+        this.setState({ alertOneVisible: false });
+        cookie.set('cs_loggedOut', 'false');
+    }
+    componentDidMount() {
```
karelhala: There should also be `componentWillUnmount` where we'd call `clearTimeout(this._timer);` in order to properly clear all timers.

src/js/App/Header/LogoutAlert.js
yashnalla: done

src/js/App/Header/LogoutAlert.js
yashnalla: done

src/js/jwt/jwt.js
yashnalla: done

src/js/App/Header/LogoutAlert.js
yashnalla: done

package-lock.json
```diff
@@ -19292,7 +19292,7 @@
     },
     "react-content-loader": {
       "version": "3.4.2",
-      "resolved": "https://registry.npmjs.org/react-content-loader/-/react-content-loader-3.4.2.tgz",
+      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/react-content-loader/-/react-content-loader-3.4.2.tgz",
```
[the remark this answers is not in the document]
yashnalla: Done! But also why exactly? I thought the idea was to change everything to use repository.engineering.redhat.com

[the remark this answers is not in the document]
ryelo: We're just supposed to use that for our registry internally. The problem is that if there is an outside contributor, they won't be able to use that since it's Red Hat internal specific.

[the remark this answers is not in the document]
yashnalla: ahh okay

PR #399, opened 2019-10-14
Added svg images for sources.

Hyperkid123:
### Added icons for commonly used images

so far these icons should be used in
- catalog
- sources
- landing page icons
- topology viewer (in the future)


Hyperkid123:
cc @karelhala

rvsia:
:+1:

PR #406, opened 2019-10-18
Update inventory package to new version 0.0.29

karelhala:


ryelo:
double approved 🚀

karelhala:
let's :shipit: after sprint demo so we won't break anything for others

PR #414, opened 2019-10-22
add nav ouia attributes

ryelo:


src/js/App/Sidenav/Navigation.js
```diff
@@ -101,6 +102,7 @@ class Navigation extends Component {
                                                     && window.location.pathname.indexOf('/beta') === -1)) {
                                                     return <NavigationItem
                                                         itemID={subItem.reload || subItem.id}
+                                                        ouia-nav-item={item.id}
```
karelhala: I think that this should be ` ouia-nav-item={subItem.reload || subItem.id`}` otherwise this will be the same as ` ouia-nav-group` attr.

ryelo: Fixed!

PR #421, opened 2019-10-25
Remove old values from local storage

PanSpagetka:
Remove duplicate values (with different hash) from local storage when they are not from cache.
Remove all values on logout.

There is also one more value that should be deleted on logout, but I wasn't able to find it yet.

https://projects.engineering.redhat.com/browse/RHCLOUD-2538


src/js/jwt/jwt.js
```diff
@@ -209,7 +210,11 @@ function logout(bounce) {
     cookie.remove(priv.cookie.cookieName);
 
     const isBeta = (window.location.pathname.split('/')[1] === 'beta' ? '/beta' : '');
-
+    const keySets = [
```
karelhala: You don't have to use two filters, just use one and we are good to go
```JS
const keys = Object.keys(localStorage).filter(key => (
  key.endsWith('/api/entitlements/v1/services') || key.endsWith('/config/main.yml')
))
```

And use them like
```JS
deleteLocalStorageItems(keys)
```

src/js/nav/sourceOfTruth.js
```diff
@@ -17,6 +18,14 @@ module.exports = (cachePrefix) => {
 
     const instance = axios.create({ adapter: cache.adapter });
 
+    instance.interceptors.response.use((response) => {
+        if (response && response.request && response.request.fromCache !== true) {
+            const keys = Object.keys(localStorage).filter(key => key.endsWith('/config/main.yml')).slice(0, -1);
+            keys.map(key => localStorage.removeItem(key));
```
karelhala: You forgot to call `deleteLocalStorageItems` here.

PR #443, opened 2019-11-05
dont send settings to a new tab

ryelo:
originally, we did

now, we don't


tahmidefaz:
@karelhala it was bugging me as well!

PR #444, opened 2019-11-06
Reuse cache from localStorage

karelhala:
Looks like our cache busting was too good and removed all items every time. This reuses active cache so when changing between tabs it will use the same data. I had to remove cache invalidation, because it was removing all things instead of active one @PanSpagetka can you look at how to enable it back? We can merge this one and reintroduce the cache invalidation after it.

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -73,10 +73,9 @@ export function deleteLocalStorageItems(keys) {
     keys.map(key => localStorage.removeItem(key));
 }
 
-export function lastActive(searhString, fallback) {
+export function lastActive(searchString, fallback) {
     return Object.keys(localStorage).reduce((acc, curr) => {
-        console.log(curr, searhString);
-        if (curr.includes(searhString)) {
+        if (curr.includes(searchString)) {
             try {
                 let accDate;
                 try {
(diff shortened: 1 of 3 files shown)


src/js/utils.js
```diff
@@ -70,3 +72,42 @@ export function createReduxListener(store, path, fn) {
...
+            }
+        }
+
+        return acc;
+    }, fallback);
+}
+
+export function bootstrapCache(endpoint, cacheKey) {
+    const name = lastActive(endpoint, cacheKey);
+    console.log(name, 'fff');
```
PanSpagetka: Forgotten console.log ;)

src/js/utils.js
```diff
@@ -70,3 +72,42 @@ export function createReduxListener(store, path, fn) {
 export function deleteLocalStorageItems(keys) {
     keys.map(key => localStorage.removeItem(key));
 }
+
+export function lastActive(searhString, fallback) {
+    return Object.keys(localStorage).reduce((acc, curr) => {
+        console.log(curr, searhString);
```
PanSpagetka: Another forgotten console.log and `searchString` typo

PR #454, opened 2019-11-08
Update Beta usage instructions

josejulio:
I tried with the `insightsbeta` and `@@insights` inside `index.html`, but it didn't work, i only got a blank page.

What worked was the suggestion from @ryelo so i'm updating these docs.


josejulio:
> These should say `/@@env/`. We don't want to hardcode these, but they should be swapped out in your app

done

PR #506, opened 2019-12-11
Add no access page for app settings

tahmidefaz:
Jira: [Here](https://projects.engineering.redhat.com/browse/RHCLOUD-3533)


tahmidefaz:
Hey guys! Do let me know if anything can be improved here!

src/js/entry.js
```diff
@@ -116,6 +117,16 @@ export function bootstrap(libjwt, initFunc) {
                         libjwt.jwt.logoutAllTabs();
                     });
                 },
+                getUserEntitlements: () => {
+                    const keys = Object.keys(localStorage).filter(key => (
```
ryelo: we should probably add a check here to make sure localStorage is enabled in the first place

ryelo: Can we change this `keys` variable name to something more specific?

src/js/entry.js
```diff
@@ -208,3 +219,21 @@ export function rootApp() {
...
     }
 }
+
+export function noAccess() {
+    const { store } = spinUpStore();
+    const app = location.pathname.split('/').pop();
+    if (app === 'insights') {
+        const userEntitlements = window.insights.chrome.auth.getUserEntitlements();
+        const grantAccess = userEntitlements.insights.is_entitled;
+        if (!grantAccess) {
```
ryelo: This isn't going to scale very well when we get more apps. Can we abstract this check to another function somewhere? That way, we don't have to do a big rework when Ansible's Automation Hub, for example, needs some kind of check.

src/sass/chrome.scss
```diff
@@ -57,6 +57,12 @@ aside {
     background: var(--pf-c-page__main-section--BackgroundColor);
 }
 
+// NoAccess styling
+#no-access {
```
ryelo: Two things here:

1. Can we do `.className#no-access` selector instead of just the ID? (where className is the class name for this ID)
2. Can we split up the height and background and then attach it to the `// Page layout styles` & the `// Page root background color`, respectively?

src/js/App/RootApp.js
```diff
@@ -40,6 +41,25 @@ const RootApp = ({
...
+                    <section
+                        className="pf-m-light pf-c-page-header pf-c-page__main-section pf-m-light"
+                        widget-type="InsightsPageHeader"
+                    >
+                        <div className="pf-c-content">
+                            <h1
+                                className="pf-c-title pf-m-2xl ins-l-page__header"
+                                widget-type='InsightsPageHeaderTitle'
+                            >
+                                <div className="appTitle">
```
ryelo: Try not to camelCase classNames. Let's stay consistent with BEM.

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,44 @@
...
+} from '@patternfly/react-core';
+
+import { LockIcon } from '@patternfly/react-icons';
+
+const NoAccess = ({ appName }) => {
+    appName = appName.charAt(0).toUpperCase() + appName.slice(1);
+    const headerClass = document.querySelector('.appTitle');
+    headerClass.innerHTML = appName;
+    return (
+        <div>
```
ryelo: is there a reason you're wrapping the page with an empty `div`?

package.json
```diff
@@ -22,6 +22,7 @@
     "lint:js": "eslint src",
     "lint:sass": "stylelint 'src/sass/**/*.scss' --config .stylelintrc.json",
     "test": "jest --forceExit --detectOpenHandles",
+    "update:snapshot": "jest --updateSnapshot",
```
ryelo: this should probably be `"test:update"`

src/js/entry.js
tahmidefaz: I went for an object here. Let me know if this doesn't work!

src/js/App/NoAccess.js
tahmidefaz: that `div` is really sneaky... thanks!

src/js/entry.js
```diff
@@ -208,3 +219,24 @@ export function rootApp() {
         );
     }
 }
+
+export function noAccess() {
+    const { store } = spinUpStore();
+    const currPath = location.pathname;
+    const app = currPath.split('/').pop();
+    const userEntitlements = window.insights.chrome.auth.getUserEntitlements();
+    const apps = { insights: userEntitlements.insights.is_entitled };
```
karelhala: Please do not bind this only to `insights` essentially we want this to be applied to all apps not only settings. You get what user is entitled in the entitlements object and you can then check if path contains such app (it should be on 0th or 1st place, based on environment)

src/js/entry.js
```diff
@@ -208,3 +219,24 @@ export function rootApp() {
...
+    //Only restrict this in settings/applications for now
+    if (currPath.includes('/settings/applications') && app in apps) {
+        const grantAccess = apps[app];
+        if (!grantAccess) {
+            document.getElementById('root').style.display = 'none';
+            render(
+                <Provider store={ store }>
+                    <NoAccess appName={ app } />
+                </Provider>,
+                document.querySelector('.pf-c-page__no-access')
```
karelhala: You can replace the entire `#no-access` container, that way you will have full access to the screen from within `NoAccess` app.

src/js/entry.js
tahmidefaz: @karelhala I am looking at the Jira for the issue, it sounds like we want this only in settings for now. Maybe I misunderstood here. Can you or @ryelo confirm that we are activating this everywhere?

src/js/entry.js
```diff
@@ -208,3 +209,25 @@ export function rootApp() {
         );
     }
 }
+
+export function noAccess() {
+    const { store } = spinUpStore();
+    const currPath = location.pathname;
+    const app = currPath.split('/').pop();
+    window.insights.chrome.auth.getUser().then(({ entitlements }) => {
+        const apps = { insights: entitlements.insights.is_entitled };
```
karelhala: This works perfectly for insights, what do you think about including other apps as well?
```JS
window.insights.chrome.auth.getUser().then(({ entitlements }) => {
  // rhel has different entitlements key and URL partial
  entitlements.rhel = entitlements.smart_management;
  const grantAccess = Object.entries(entitlements).filter(([ app, { is_entitled } ]) => {
    // check if app key from entitlements is anywhere in URL and if so check if user is entitled for such app
    return location.pathname.split('/').includes(app) && is_entitled;
  });
  if (grantAccess && grantAccess.length > 0) {
    document.getElementById('root').style.display = 'none';
    render(
      <Provider store={ store }>
        <NoAccess />
      </Provider>,
      document.querySelector('#no-access')
    )
  }
})
```
`location.pathname.split('/').includes(app) && is_entitled` has limitations, if there's going to be URL with `insights/rhel` in future this will return true. But we'll need much complicated logic to calculate such thing. But for now I feel like it is sufficient enough.

tahmidefaz: Yes, that makes sense! I will include this in the commit. Thanks!

tahmidefaz:
@karelhala Your suggestion works great! I just had to tweak the logic a little bit and grant access to pages like `settings/general`, since they don't have the app names in the path.

@karelhala @ryelo this is up for re-review, I think!

src/js/entry.js
```diff
@@ -208,3 +209,33 @@ export function rootApp() {
...
     }
 }
+
+export function noAccess() {
+    const { store } = spinUpStore();
+    window.insights.chrome.auth.getUser().then(({ entitlements }) => {
+        // rhel has different entitlements key and URL partial
+        entitlements.rhel = entitlements.smart_management;
+        const path = location.pathname.split('/');
+        const apps = [];
```
karelhala: The list of apps that require entitlements is better to be added seperately and not part of filter function.

```JS
const apps = Object.keys(entitlements);
```

src/js/entry.js
```diff
@@ -208,3 +209,33 @@ export function rootApp() {
...
+        /* eslint-disable camelcase */
+        const grantAccess = Object.entries(entitlements).filter(([app, { is_entitled }]) => {
+            // check if app key from entitlements is anywhere in URL and if so check if user is entitled for such app
+            apps.push(app);
+            return path.includes(app) && is_entitled;
+        });
+        /* eslint-enable camelcase */
+
+        // also grant access to other pages like settings/general
+        const untrackedApp = path.filter(value => apps.includes(value)).length < 1 ? true : false;
```
karelhala: It is generaly better to use `some` to check if one array contains something from another one
```JS
const isTrackedApp = path.some(value => app.includes(value));

if (!(grantAccess && grantAccess.length > 0) && isTrackedApp) {
```

The folter above can be changed to use `some` as well, but let's keep it there.

src/js/entry.js
```diff
@@ -208,3 +209,33 @@ export function rootApp() {
...
+    window.insights.chrome.auth.getUser().then(({ entitlements }) => {
+        // rhel has different entitlements key and URL partial
+        entitlements.rhel = entitlements.smart_management;
+        const path = location.pathname.split('/');
+        const apps = [];
+
+        /* eslint-disable camelcase */
+        const grantAccess = Object.entries(entitlements).filter(([app, { is_entitled }]) => {
+            // check if app key from entitlements is anywhere in URL and if so check if user is entitled for such app
+            apps.push(app);
```
karelhala: It is not good to assign/change something outside of scope filter.

tahmidefaz:
@karelhala I think this should be good now. Let me know if you see anything off here!

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,65 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+// import { Main, PageHeader, PageHeaderTitle } from '@redhat-cloud-services/frontend-components';
```
ryelo: Can we remove this comment?

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,65 @@
...
+
+import { LockIcon } from '@patternfly/react-icons';
+
+const NoAccess = ({ activeAppTitle }) => {
+    return (
+        <Fragment>
+            <section
+                className="pf-m-light pf-c-page-header pf-c-page__main-section pf-m-light"
+                widget-type="InsightsPageHeader"
+            >
```
ryelo: I know it's stylistic, but can we move the `>` up to the previous line?

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,65 @@
...
+                    <Title headingLevel="h5" size="lg">
+                        { `You do not have access to ${ activeAppTitle }` }
+                    </Title>
+                    <EmptyStateBody>
+                        Contact your organization administrator(s) for more information.
+                    </EmptyStateBody>
+                    {
+                        document.referrer ?
+                            <Button variant="primary" component="a" onClick={ () => history.back() }>Return to previous page</Button> :
+                            <Button variant="primary" component="a" href=".">Go to landing page</Button>
```
ryelo: why are you using `href="."`?

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,65 @@
...
+                    <EmptyStateIcon icon={ LockIcon } />
+                    <Title headingLevel="h5" size="lg">
+                        { `You do not have access to ${ activeAppTitle }` }
+                    </Title>
+                    <EmptyStateBody>
+                        Contact your organization administrator(s) for more information.
+                    </EmptyStateBody>
+                    {
+                        document.referrer ?
+                            <Button variant="primary" component="a" onClick={ () => history.back() }>Return to previous page</Button> :
```
ryelo: I think we should avoid using an onclick on an `a` component if necessary. Is there a reason we can't just leave this as a button?

src/js/App/RootApp.js
```diff
@@ -40,6 +40,7 @@ const RootApp = ({
                         </div>
                     </section>
                 </main>
+                <main className="pf-c-page__main pf-l-page__main" id="no-access"></main>
```
ryelo: We don't need to use the legacy classnames

src/js/entry.js
```diff
@@ -208,3 +209,32 @@ export function rootApp() {
         );
     }
 }
+
+export function noAccess() {
+    const { store } = spinUpStore();
+    window.insights.chrome.auth.getUser().then(({ entitlements }) => {
```
ryelo: We should catch this if it fails

src/sass/chrome.scss
```diff
@@ -48,12 +48,12 @@ aside {
...
 
 // Page layout styles
-.pf-c-drawer__content, #root.pf-c-page__main, #temp.pf-c-page__main {
+.pf-c-drawer__content, #root.pf-c-page__main, #temp.pf-c-page__main, #no-access.pf-c-page__main {
     height: 100%;
 }
 
 // Page root background color
-.pf-c-page__main#root {
+.pf-c-page__main#root, #no-access.pf-c-page__main {
```
ryelo: i'd rewrite this to say:

```
.pf-c-page__main {
  &#root, &#no-access {
     ....
```

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,65 @@
...
+                widget-type="InsightsPageHeader"
+            >
+                <div className="pf-c-content">
+                    <h1
+                        className="pf-c-title pf-m-2xl ins-l-page__header"
+                        widget-type='InsightsPageHeaderTitle'
+                    >
+                        <div className="apptitle">
+                            { `${activeAppTitle}` }
+                            &nbsp;
```
ryelo: Is there a reason you're using `&nbsp` here?

src/js/App/NoAccess.js
```diff
@@ -0,0 +1,65 @@
...
+    EmptyStateBody
+} from '@patternfly/react-core';
+
+import { LockIcon } from '@patternfly/react-icons';
+
+const NoAccess = ({ activeAppTitle }) => {
+    return (
+        <Fragment>
+            <section
+                className="pf-m-light pf-c-page-header pf-c-page__main-section pf-m-light"
```
ryelo: Could you just use the page header component here?

tahmidefaz: @ryelo I tried, but it kept giving some sort of router error in the console whenever I imported anything from `frontend-components`. So, had to opt-in for this workaround

src/js/App/RootApp.js
tahmidefaz: Legacy classnames as in `pf-l-page__main`?

ryelo: Yep!

src/js/entry.js
tahmidefaz: @ryelo what would be a good way to catch when it fails here? throw some logs?

ryelo: Throw a log or an error of some kind. This will generate sentry errors if we don't catch it!

src/js/App/NoAccess.js
ryelo: Works for me!

tahmidefaz:
@ryelo apart from the one above, all other requested changes were addressed. I also added something to throw logs when the entitlements fetch promise fails. Let me know if this is not sufficient

PR #507, opened 2019-12-12
add localstorage to force pendo initialization

ryelo:
We want the ability to force pendo initialization whenever an account is marked as internal


testdata/externalToken.json
```diff
@@ -0,0 +1,32 @@
+{
```
karelhala: Is this file correctly commited? What is it used for?

ryelo: Removed!

PR #548, opened 2020-01-13
remove "report bug" link from openshift

petli-openshift:


petli-openshift:
@karelhala @ryelo could you review this and push to prod beta? thanks!

cc @elad661

PR #553, opened 2020-01-16
Add permission checker constants and logic to hide navigation elements

karelhala:
### Features
Applications wants to hide navigation based on certain rules, we have to help them in this by providing helper functions and also hide these nav items if they are acutally not accessible.
### Notes
* [x] Add tests
* There is todo about api caching, but that can be ignored for now as this introduces mainly logic for checking if something can be shown.
* There is edge case when all second level navigation can be hidden, we'd probably place About in such places. But this About is not yet designed and it is not part of this PR.
### JIRA
RHCLOUD-2629


karelhala:
Sorry guys for such a big change, but I had to rewrite bunch of functions to use async/await because it was super complicated or it used array push in places where it was not necessary.

src/js/consts.js
```diff
@@ -1,7 +1,56 @@
...
+    isEntitled: async (appName) => {
+        const { entitlements } = await insights.chrome.auth.getUser();
+        return appName ? Boolean(entitlements[appName] && entitlements[appName].is_entitled) :
+            // eslint-disable-next-line camelcase
+            Object.entries(entitlements).reduce((acc, [key, { is_entitled }]) => ({ ...acc, [key]: is_entitled }), {});
+    },
+    isProd: () => {
+        return insights.chrome.isProd;
+    },
+    isBeta: () => {
```
rvsia: What about just `() => insights.chrome.isBeta()`? same above

src/js/consts.js
```diff
@@ -1,7 +1,56 @@
...
+        return insights.chrome.isBeta();
+    },
+    apiRequest: async ({ url, method, ...props }) => {
```
Hyperkid123: options

src/js/nav/globalNav.js
```diff
@@ -1,47 +1,63 @@
...
+    return [];
 }
 
 // Gets the app's data from the master config, if it exists
-function getAppData(appId, propName, masterConfig) {
+async function getAppData(appId, propName, masterConfig) {
     const app = masterConfig[appId];
     if (app && Object.prototype.hasOwnProperty.call(app, 'frontend')) {
-        const routes = getRoutesForApp(app, masterConfig);
+        if (app.permissions && !app.permissions.apps) {
```
PanSpagetka: Is `!app.permissions.apps` this check necessary?

karelhala: Yes, this hides the entire section. So if dev does not specify apps it is ment to hide the entire firt level navigation.

src/js/consts.test.js
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable camelcase */
```
Hyperkid123: Maybe we could disable this rule globally

PR #563, opened 2020-01-21
Update prod-beta to newest master

karelhala:
* New inventory with tagging
* No access page
* Removed "report but" from openshift
* Doc links for openshift updated


karelhala:
When merging use `Create a merge commit` so we have hashes aligned.

PR #597, opened 2020-02-03
Add documentation link for Cost Management.

chambridge:
Add documentation link for [cost management](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/#category-cost-management)

### DIFF
--- a/src/js/App/Sidenav/index.js
+++ b/src/js/App/Sidenav/index.js
@@ -9,7 +9,8 @@ const documentationLink = {
     rhel: 'https://access.redhat.com/documentation/en-us/cloud_management_services_for_red_hat_enterprise_linux/1.0/',
     insights: 'https://access.redhat.com/documentation/en-us/red_hat_insights/',
     openshift: 'https://docs.openshift.com/container-platform/latest/',
-    subscriptions: 'https://access.redhat.com/documentation/Subscription_Central/'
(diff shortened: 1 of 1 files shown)


src/js/App/Sidenav/index.js
```diff
@@ -9,7 +9,8 @@ const documentationLink = {
     rhel: 'https://access.redhat.com/documentation/en-us/cloud_management_services_for_red_hat_enterprise_linux/1.0/',
     insights: 'https://access.redhat.com/documentation/en-us/red_hat_insights/',
     openshift: 'https://docs.openshift.com/container-platform/latest/',
-    subscriptions: 'https://access.redhat.com/documentation/Subscription_Central/'
+    subscriptions: 'https://access.redhat.com/documentation/Subscription_Central/',
+    'cost-management': 'https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/#category-cost-management'
```
ryelo: Do you have a non specified doc link? Here, it's specifying `4.3`, so if `4.4` comes out, will this have to be updated?

chambridge: This appears to work: https://access.redhat.com/documentation/en-us/openshift_container_platform/#category-cost-management

I'll update it to drop the version. It seems to select the latest version if not provided.

chambridge: Done

PR #602, opened 2020-02-06
Add flex to toolbar and it's children

karelhala:


karelhala:
This bubled to us from PF issue https://github.com/patternfly/patternfly-react/issues/3664

PR #654, opened 2020-02-25
add ability to use pen test cookie

ryelo:


kruai:
Since Chrome uses the hostname to determine which SSO to send to, this is actually a no-op.
The hostname will still be cloud.redhat.com in the pen-test env, so even if it's the QA version of Chrome, it'll still route to Prod SSO.

PR #655, opened 2020-02-25
add subscriptions and settings to allow unentitled

ryelo:
Chrome messed up because settings and subs were allowing users without account numbers, but chrome didn't know that. The user then sees:

Screen Shot 2020-02-25 at 3 25 32 PM

To repro, use `test-user-without-insights-sku/redhat`

Adding these apps to the `pageAllowsUnentitled` lets accounts without numbers access the page and chrome won't blow up.

### DIFF
--- a/src/js/jwt/insights/user.js
+++ b/src/js/jwt/insights/user.js
@@ -7,9 +7,7 @@ const pathMapper = {
     insights: 'insights',
     openshift: 'openshift',
     migrations: 'migrations',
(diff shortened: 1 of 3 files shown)


ryelo:
I want to check with subscription watch team to verify this on their side to make sure we don't break their pendo stuff.

ryelo:
Going to close this one out. I want to separate this into two PRs to do settings in one and subs in the other.

PR #694, opened 2020-03-18
fix attributes on toolbar

ryelo:
Widgets were being identified incorrectly and we need to put ids and attributes on the buttons


karelhala:
Re running travis!

PR #736, opened 2020-04-06
add a subscription watch nav item

ryelo:
https://projects.engineering.redhat.com/browse/RHCLOUD-4951

~on hold until: https://github.com/RedHatInsights/cloud-services-config/pull/160~

### DIFF
--- a/src/js/App/Sidenav/Navigation.js
+++ b/src/js/App/Sidenav/Navigation.js
@@ -21,7 +21,7 @@ const openshiftLinks = {
 const insightsLinks = {
     subscriptionWatch: {
         title: 'Subscription Watch',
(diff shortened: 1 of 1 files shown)


src/js/App/Sidenav/Navigation.js
```diff
@@ -18,6 +18,13 @@ const openshiftLinks = {
     }
 };
 
+const insightsLinks = {
+    subscriptionWatch: {
+        title: 'Subscription Watch',
+        link: `${window.location.origin}${window.insights.chrome.isBeta() ? '/beta/' : '/'}subscriptions/`
```
karelhala, line 24: It should be enough to add `./subsciptions` because we are setting the base href. You can also use `${document.baseURI}/subscription` if you want to have full URL just check if base uri contains last slash or not.

PR #776, opened 2020-04-23
update scss to load styles

ryelo:
So here's what happened:

In some cases, chrome was not loading styles correctly. Depending on how apps would handle importing, the styles would be reset on load, and therefore making styles not load entirely for some components in chrome.

This mainly affected modals, and some hot loaded components. 

This was also the reason we had been importing select components inside of chrome.scss

Updating this to include the components and layouts seems to fix this issue. We could do these one by one, but at this point we're already including a ton of files.

This also fixes the modals not working in some places.

We should blast an email and message once we verify this is working to apps to double check their styles because this could be a breaking change.


ryelo:
Updated to only what we need

PR #791, opened 2020-04-30
Use Stage SSO in Stage env

kruai:
Prepping the Stage env. When on this hostname, it should use Stage SSO.
Are any more changes required for this?


kruai:
@ryelo yeah, we're starting to try to follow RH's standards for that stuff. I'm sure we'll eventually have cloud.qa.redhat.com too

PR #803, opened 2020-05-07
fix lint spacing

ryelo:


ryelo:
Ah this was passing for me locally, must have something messed up

PR #817, opened 2020-06-01
Upgrade patternfly

ryelo:
~*on hold until official release comes out*~

Sorry for the hacky imports, not really my favorite way to do it, but it should be fine until Inventory is updated and then the imports will be removed.

Also, there is a temp nav styling file that will be removed once we officially release. This is just a buffer so that if an app is using old styles that the nav will still look good.

# What

* Updates Patternfly and Patternfly React to their respective 4.x packages
  * Adds temporary styles for the sidenav (these can be removed when apps have finished updating)
  * Rewrote the masthead using hooks and the correct components, should be easier to read
* Removes a few console.errors from the about modal because of nested children
* Adds an "App switcher" to the sidenav (replaces the home icon + app title)

## What is left

- [x] verification on other apps
- [x] UXD signoff


src/js/App/Sidenav/AppSwitcher.js
```diff
@@ -0,0 +1,63 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import { Dropdown } from '@patternfly/react-core4/dist/js/components/Dropdown/Dropdown';
+import { DropdownItem } from '@patternfly/react-core4/dist/js/components/Dropdown/DropdownItem';
+import { DropdownToggle } from '@patternfly/react-core4/dist/js/components/Dropdown/DropdownToggle';
+import { CaretDownIcon } from '@patternfly/react-icons';
```
karelhala, line 8: Can you add direct import to this icon? That way Parcel has easier treeshaking.

josejulio:
Not sure if a bug here. I'm testing this branch with the `PF4-update` of the frontend-components and when the width of the page is between [768, 1119] px the sidebar doesn't show:

Peek 2020-06-15 13-55

note how the collapsed class is being added/removed but doesn't show the sidebar

Tested this PF page and it doesn't happen: https://patternfly-react.surge.sh/documentation/react/components/page/with-or-without-fill

ryelo:
@josejulio Ah good catch! The breakpoint was changed on the PF side, I'm updating our implementation to reflect.

ryelo:
@josejulio Updated! The nav should collapse on its own below 1200px and you should be able to toggle it again.

josejulio:
It is working well now, thanks @ryelo !

ryelo:
@karelhala @PanSpagetka I'm still facing one little bug (style issue), but I think this should be okay to merge for the time being and iterate so apps can use it to upgrade.

PR #836, opened 2020-06-10
Fix highlight of duplicate sub-nav items

fhlavac:
This should fix the issue with multiple highlights when the sub-nav item name is already used by another top-level nav item.

Needed for: https://github.com/RedHatInsights/cloud-services-config/pull/218

@karelhala

### DIFF
--- a/src/js/redux/reducers.js
+++ b/src/js/redux/reducers.js
@@ -54,7 +54,9 @@ export function appNavClick(state, { payload }) {
         globalNav: payload.custom ? state.globalNav && state.globalNav.map(item => ({
             ...item,
             active: payload && (item.id === payload.id || item.title === payload.id)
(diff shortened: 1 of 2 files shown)


src/js/redux/reducers.js
```diff
@@ -54,7 +54,7 @@ export function appNavClick(state, { payload }) {
         globalNav: payload.custom ? state.globalNav && state.globalNav.map(item => ({
             ...item,
             active: payload && (item.id === payload.id || item.title === payload.id)
-                || (item.subItems && item.subItems.some(({ id }) => id === payload.id))
+                || (item.subItems && item.subItems.some(({ id }) => id === payload.id) && item.id === state.appId)
```
karelhala, line 57: Can you add payload parent here as well?

```suggestion
                || (item.subItems && item.subItems.some(({ id }) => id === payload.id) && (item.id === state.appId || item.id === payload.parentId))
```

PR #842, opened 2020-06-18
Webpack migration

Hyperkid123:
Replace parcel by webpack and optimize the js built output

cc @karelhala 

### TO DO
- [x] js build
- [x]  js build beta
- [x] watch js
- [ ] optimize chunks

## Major development change

I was constantly dealing with cache issues in development. That was solved by adding a chunk hash to the root `chrome.js` file. However, this means that in development we also need to serve the HTML generated from pug templates.

I do believe we want to include the hash if the app is using code splitting to avoid any potential issue with state root JS asset. But this change makes the development bit harder. You have to also run some app and you need to manually add the root element to body. However, without that the development was a serious pain because I was constantly dealing with outdated JS assets and the app was sporadically failing when using minimized bundles. @ryelo @karelhala your thoughts?


Hyperkid123:
The issue with broken production build was fixed.

Now there is a another one when the navigation and the user are sometimes not set in the redux store. (they are not set fast enough)

Hyperkid123:
The issue with missing identity information are resolved. Now I will start to clean up the mess I did in the JS files and move to other build tasks.

Hyperkid123:
The migration is complete.

@karelhala @ryelo Can I ask you to do check this build? You know the platform better than me and it is very likely that I have forgot to check some app or a feature.

Looking at the build output, we are doing fairly OK. The main issue is inventory dependencies. It pulls around 1.4mb and the fact that chrome itself has some strangely large dependencies. I am optimistic that we can fix it but we will most likely need to start with FCE.

screenshot-127 0 0 1_8888-2020 06 24-08_50_10

ryelo:
@Hyperkid123 Yes! I'll dig through some of this later today

ryelo:
@Hyperkid123 I tested this with both the landing page and RBAC and I'm getting a bunch of chunk failures. The apps are still rendering, but they leave the nav and masthead unusable.

Hyperkid123:
@ryelo interesting I am not seeing that. Maybe some cache issue? I will try and reproduce it.

Hyperkid123:
@ryelo there a bug with the public path. There was missing `/` and the chunks were loaded from `apps/chrome/jsChunkName` instead of `apps/chrome/js/ChunkName`. I wonder why it was not an issue when I was setting that up :thinking:.

I was experiencing some wire cache issues while switching to the webpack. Some times it was trying to load assets that no longer existed. I hope I wont be an issue for you.

ryelo:
@Hyperkid123 So this one seems better for some pages, but a couple more bugs I'm getting:
* 404s for skeleton.css and spinner.css
* Uncaught TypeError: Cannot read property 'call' of undefined 

Although I also can't get the sourcemaps to load, so it's kind of difficult to see where number 2 is coming from.

Hyperkid123:
> Uncaught TypeError: Cannot read property 'call' of undefined

@ryelo I don't see that when I run it locally :thinking: in which app are you seeing that?

The skeleton is coming from FCE. It tries to reach totally wrong resource at 
```
/apps/chrome/node_modules/@redhat-cloud-services/frontend-components/components/skeleton.css
```

Hyperkid123:
@ryelo well, the CSS invalid routes is actually not caused by this PR. I am setting that currently on `ci-beta` env.

Hyperkid123:
So the undefined call is caused by using the Suspense/lazy combination. Turning off side effects for production optimizations (is set to true by default) might help with that.

Hyperkid123:
@ryelo as usual the simplest solution is the best. I've renamed the `webpackJSONP` function for the chrome. That fixed the problem. For reference: https://webpack.js.org/configuration/output/#outputjsonpfunction

The issue was that sometimes (completely random) the application has overwritten the chrome chunks inside the JSONP object. These objects are publically accessible on window so if you have multiple instances of webpack on one page, webpack will try to put everything into the same object.

This works fine until you start using lazy-loaded chunks.

Chome now has its own namespace called `wpJsonpChromeInstance`. Apps can still use the default `webpackJsonp`.

ryelo:
@Hyperkid123 That's a wild journey! I'll have some time tomorrow that I can go over it again. Thanks for looking over this stuff!

karelhala:
Just dropping this here from our conversation @Hyperkid123 https://github.com/jherr/wp5-intro-video-code/blob/master/packages/home/src/App.jsx so we don't forget about it. Great point using this for inventory @rvsia!

ryelo:
@Hyperkid123 I'm getting some errors on build

```
ERROR in ./node_modules/@redhat-cloud-services/frontend-components-inventory/index.js
Module not found: Error: Can't resolve 'PFReactTable' in '/Users/rlong/Insights/insights-chrome/node_modules/@redhat-cloud-services/frontend-components-inventory'
 @ ./node_modules/@redhat-cloud-services/frontend-components-inventory/index.js 2:127-150
 @ ./src/js/inventory.js
 @ ./src/js/entry.js
 @ ./src/js/chrome.js

ERROR in ./node_modules/@redhat-cloud-services/frontend-components-inventory/index.js
Module not found: Error: Can't resolve 'customReact' in '/Users/rlong/Insights/insights-chrome/node_modules/@redhat-cloud-services/frontend-components-inventory'
 @ ./node_modules/@redhat-cloud-services/frontend-components-inventory/index.js 2:104-126
 @ ./src/js/inventory.js
 @ ./src/js/entry.js
 @ ./src/js/chrome.js

ERROR in ./node_modules/@redhat-cloud-services/frontend-components-inventory/index.js
Module not found: Error: Can't resolve 'reactRedux' in '/Users/rlong/Insights/insights-chrome/node_modules/@redhat-cloud-services/frontend-components-inventory'
 @ ./node_modules/@redhat-cloud-services/frontend-components-inventory/index.js 2:375-396
 @ ./src/js/inventory.js
 @ ./src/js/entry.js
 @ ./src/js/chrome.js
```

Hyperkid123:
Ah. I guess this was caused by the latest inventory refactoring eh? We will probably need to adjust the externals configuration.

Hyperkid123:
@ryelo ok I've added new inventory dependencies package names to externals and the build is running (also tried to run it locally and worked with the catalog app).

config/webpack.config.js
```diff
@@ -0,0 +1,74 @@
...
+}) => ({
+    entry: path.resolve(__dirname, '../src/js/chrome.js'),
+    output: {
+        path: path.resolve(__dirname, '../build/js'),
+        filename: 'chrome.[chunkhash].js',
+        publicPath,
+        chunkFilename: '[name].[chunkhash].js',
+        jsonpFunction: 'wpJsonpChromeInstance'
+    },
+    externals: {
```
karelhala, line 17: Externals won't work as these are not externals but rhat pulled from different app. I think that `resolve.alias` is more suitable for this
```JS
resolve: {
        alias: {
            PFReactTable: path.resolve(__dirname, './patternfly-table-externals.js'),
            '@patternfly/react-table': path.resolve(__dirname, './patternfly-table-externals.js'),
            customReact: path.resolve(__dirname, './react-external.js'),
            reactRedux: path.resolve(__dirname, './react-redux-external.js'),
            'react-router-dom': path.resolve(__dirname, './react-router-dom-externals.js'),
            graphql: path.resolve(__dirname, './graphql-externals.js')
        }
    }
```

src/js/inventory.js
```diff
@@ -4,6 +4,6 @@ export default (dependencies) => {
     setDependencies(dependencies);
 
     return import('./inventoryStyles').then(
-        () => import('@redhat-cloud-services/frontend-components-inventory')
+        () => import(/* webpackChunkName: "inventory-styles" */ '@redhat-cloud-services/frontend-components-inventory')
```
karelhala, line 7: I'd name this `inventory` instead of `inventory-styles`. Also might be resonable to name styles as well
```JS
return import(/* webpackChunkName: "inventory" */ './inventoryStyles').then(
  () => import(/* webpackChunkName: "inventory" */ '@redhat-cloud-services/frontend-components-inventory')
)
```

src/js/inventoryStyles.js
```diff
@@ -1,2 +1,7 @@
+/**
```
karelhala, line 1: Should be fixed.

Hyperkid123:
> All of the externals from config should be reverted to previous module.exports since that's slightly different from what export default does. If we want to use imports and exports (without module.exports) we have to figure out how to spread the hot loaded deps.

@karelhala  yeah we would have to use `require('foo').default`. Do you know what consumes these packages? I would like to test it out.

karelhala:
@Hyperkid123 the easiest way would be to run localchrome and inventory application since that one is heavily relies on these external packages.

Hyperkid123:
Alright, I will check it out.

I will fix the rest and see if we can use the alias. (I was tricked by the alias object in package json and forgot that was used for parcel config)

Hyperkid123:
@karelhala right the import/export syntax won't work in the confix because of the fact that those packages are mixing default and name exports :shrug: Otherwise it would work simply by exporting the constants. We could change the way it's used inside the inventory component but honestly, I don't have time for that kind of stuff right now. I will revert it back then.

karelhala:
@Hyperkid123 I tried using `React.createContext()` and it still failed, because some component was using named exports as well. I think the safest way here would be to use `module.exports` these files are config files and not app files so I think we are fine.

Hyperkid123:
> Also, looks like running npm run build won't use newer bundles, might be reasonable to purge build/js and build/css when running build or start script.

@karelhala that is because of the bundle hash. You need to rebuild the html as well. This was happening before with the css. This just add the hashes to the JS entry point as well.

Hyperkid123:
> @Hyperkid123 I tried using `React.createContext()` and it still failed, because some component was using named exports as well. I think the safest way here would be to use `module.exports` these files are config files and not app files so I think we are fine.

Yup agree. Already did that.

karelhala:
For some reason the rebased version now fails with react hook problems, gotta dig deeper to know why is that happenning.

Hyperkid123:
> For some reason the rebased version now fails with react hook problems, gotta dig deeper to know why is that happenning.

Not again this again. Incorrect react-dom dependency?

karelhala:
```diff
diff --git a/config/webpack.config.js b/config/webpack.config.js
index 2e06f70..9efcbaf 100644
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -35,7 +35,7 @@ const commonConfig = ({
             },
             {
                 test: /\.s?[ac]ss$/,
-                use: ['css-loader', 'sass-loader']
+                use: ['style-loader', 'css-loader', 'sass-loader']
             },
             {
                 test: /\.(jpg|png|svg)$/,
diff --git a/config/webpack.plugins.js b/config/webpack.plugins.js
index 26723b3..b79b6ba 100644
--- a/config/webpack.plugins.js
+++ b/config/webpack.plugins.js
@@ -1,8 +1,10 @@
 const webpack = require('webpack');
 const LodashWebpackPlugin = require('lodash-webpack-plugin');
 const WriteFileWebpackPlugin = require('write-file-webpack-plugin');
+const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 
 const plugins = [
+    new CleanWebpackPlugin(),
     new WriteFileWebpackPlugin(),
     new webpack.SourceMapDevToolPlugin({
         test: /\.js/i,
diff --git a/package-lock.json b/package-lock.json
index 6ef8b6a..c106ea4 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -4917,6 +4917,12 @@
       "integrity": "sha512-HGyxoOTYUyCM6stUe6EJgnd4EoewAI7zMdfqO+kGjnlZmBDz/cR5pf8r/cR4Wq60sL/p0IkcjUEEPwS3GFrIyw==",
       "dev": true
     },
+    "array-uniq": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/array-uniq/-/array-uniq-1.0.3.tgz",
+      "integrity": "sha1-r2rId6Jcx/dOBYiUdThY39sk/bY=",
+      "dev": true
+    },
     "array-unique": {
       "version": "0.2.1",
       "resolved": "https://registry.npmjs.org/array-unique/-/array-unique-0.2.1.tgz",
@@ -7789,6 +7795,84 @@
       "resolved": "https://registry.npmjs.org/clean-stack/-/clean-stack-2.2.0.tgz",
       "integrity": "sha512-4diC9HaTE+KRAMWhDhrGOECgWZxoevMc5TlkObMqNSsVU62PYzXZ/SMTjzyGAFF1YusgxGcSWTEXBhp0CPwQ1A=="
     },
+    "clean-webpack-plugin": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/clean-webpack-plugin/-/clean-webpack-plugin-3.0.0.tgz",
+      "integrity": "sha512-MciirUH5r+cYLGCOL5JX/ZLzOZbVr1ot3Fw+KcvbhUb6PM+yycqd9ZhIlcigQ5gl+XhppNmw3bEFuaaMNyLj3A==",
+      "dev": true,
+      "requires": {
+        "@types/webpack": "^4.4.31",
+        "del": "^4.1.1"
+      },
+      "dependencies": {
+        "array-union": {
+          "version": "1.0.2",
+          "resolved": "https://registry.npmjs.org/array-union/-/array-union-1.0.2.tgz",
+          "integrity": "sha1-mjRBDk9OPaI96jdb5b5w8kd47Dk=",
+          "dev": true,
+          "requires": {
+            "array-uniq": "^1.0.1"
+          }
+        },
+        "del": {
+          "version": "4.1.1",
+          "resolved": "https://registry.npmjs.org/del/-/del-4.1.1.tgz",
+          "integrity": "sha512-QwGuEUouP2kVwQenAsOof5Fv8K9t3D8Ca8NxcXKrIpEHjTXK5J2nXLdP+ALI1cgv8wj7KuwBhTwBkOZSJKM5XQ==",
+          "dev": true,
+          "requires": {
+            "@types/glob": "^7.1.1",
+            "globby": "^6.1.0",
+            "is-path-cwd": "^2.0.0",
+            "is-path-in-cwd": "^2.0.0",
+            "p-map": "^2.0.0",
+            "pify": "^4.0.1",
+            "rimraf": "^2.6.3"
+          }
+        },
+        "globby": {
+          "version": "6.1.0",
+          "resolved": "https://registry.npmjs.org/globby/-/globby-6.1.0.tgz",
+          "integrity": "sha1-9abXDoOV4hyFj7BInWTfAkJNUGw=",
+          "dev": true,
+          "requires": {
+            "array-union": "^1.0.1",
+            "glob": "^7.0.3",
+            "object-assign": "^4.0.1",
+            "pify": "^2.0.0",
+            "pinkie-promise": "^2.0.0"
+          },
+          "dependencies": {
+            "pify": {
+              "version": "2.3.0",
+              "resolved": "https://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
+              "integrity": "sha1-7RQaasBDqEnqWISY59yosVMw6Qw=",
+              "dev": true
+            }
+          }
+        },
+        "p-map": {
+          "version": "2.1.0",
+          "resolved": "https://registry.npmjs.org/p-map/-/p-map-2.1.0.tgz",
+          "integrity": "sha512-y3b8Kpd8OAN444hxfBbFfj1FY/RjtTd8tzYwhUqNYXx0fXx2iX4maP4Qr6qhIKbQXI02wTLAda4fYUbDagTUFw==",
+          "dev": true
+        },
+        "pify": {
+          "version": "4.0.1",
+          "resolved": "https://registry.npmjs.org/pify/-/pify-4.0.1.tgz",
+          "integrity": "sha512-uB80kBFb/tfd68bVleG9T5GGsGPjJrLAUpR5PZIrhBnIaRTQRjqdJSsIKkOP6OAIFbj7GOrcudc5pNjZ+geV2g==",
+          "dev": true
+        },
+        "rimraf": {
+          "version": "2.7.1",
+          "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
+          "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
+          "dev": true,
+          "requires": {
+            "glob": "^7.1.3"
+          }
+        }
+      }
+    },
     "cli-cursor": {
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/cli-cursor/-/cli-cursor-3.1.0.tgz",
@@ -12455,6 +12539,26 @@
       "integrity": "sha512-w942bTcih8fdJPJmQHFzkS76NEP8Kzzvmw92cXsazb8intwLqPibPPdXf4ANdKV3rYMuuQYGIWtvz9JilB3NFQ==",
       "dev": true
     },
+    "is-path-in-cwd": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/is-path-in-cwd/-/is-path-in-cwd-2.1.0.tgz",
+      "integrity": "sha512-rNocXHgipO+rvnP6dk3zI20RpOtrAM/kzbB258Uw5BWr3TpXi861yzjo16Dn4hUox07iw5AyeMLHWsujkjzvRQ==",
+      "dev": true,
+      "requires": {
+        "is-path-inside": "^2.1.0"
+      },
+      "dependencies": {
+        "is-path-inside": {
+          "version": "2.1.0",
+          "resolved": "https://registry.npmjs.org/is-path-inside/-/is-path-inside-2.1.0.tgz",
+          "integrity": "sha512-wiyhTzfDWsvwAW53OBWF5zuvaOGlZ6PwYxAbPVDhpm+gM09xKQGjBq/8uYN12aDvMxnAnq3dxTyoSoRNmg5YFg==",
+          "dev": true,
+          "requires": {
+            "path-is-inside": "^1.0.2"
+          }
+        }
+      }
+    },
     "is-path-inside": {
       "version": "3.0.2",
       "resolved": "https://registry.npmjs.org/is-path-inside/-/is-path-inside-3.0.2.tgz",
@@ -22525,6 +22629,12 @@
       "resolved": "https://registry.npmjs.org/path-is-absolute/-/path-is-absolute-1.0.1.tgz",
       "integrity": "sha1-F0uSaHNVNP+8es5r9TpanhtcX18="
     },
+    "path-is-inside": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/path-is-inside/-/path-is-inside-1.0.2.tgz",
+      "integrity": "sha1-NlQX3t5EQw0cEa9hAn+s8HS9/FM=",
+      "dev": true
+    },
     "path-key": {
       "version": "3.1.1",
       "resolved": "https://registry.npmjs.org/path-key/-/path-key-3.1.1.tgz",
diff --git a/package.json b/package.json
index 547c56b..5e49e21 100644
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-transform-inline-environment-variables": "^0.4.3",
     "babel-runtime": "^6.26.0",
+    "clean-webpack-plugin": "^3.0.0",
     "concurrently": "^5.2.0",
     "css-loader": "^3.6.0",
     "del-cli": "^3.0.1",
diff --git a/src/js/inventory/details.js b/src/js/inventory/details.js
index 742c62b..1d53d8b 100644
--- a/src/js/inventory/details.js
+++ b/src/js/inventory/details.js
@@ -2,11 +2,26 @@
 /* eslint-disable react/display-name */
 /* eslint-disable camelcase */
 import React, { Suspense, lazy } from 'react';
-const GeneralInformation = lazy(() => import('@redhat-cloud-services/frontend-components-inventory-general-info/esm'));
-const PatchMan = lazy(() => import('@redhat-cloud-services/frontend-components-inventory-patchman'));
-const Advisor = lazy(() => import('@redhat-cloud-services/frontend-components-inventory-insights/esm'));
-const Vulnerabilities = lazy(() => import('@redhat-cloud-services/frontend-components-inventory-vulnerabilities'));
-const Compliance = lazy(() => import('@redhat-cloud-services/frontend-components-inventory-compliance/esm'));
+const GeneralInformation = lazy(() => import(
+    /* webpackChunkName: "inventory-general-info" */
+    '@redhat-cloud-services/frontend-components-inventory-general-info/esm'
+));
+const PatchMan = lazy(() => import(
+    /* webpackChunkName: "inventory-patch" */
+    '@redhat-cloud-services/frontend-components-inventory-patchman'
+));
+const Advisor = lazy(() => import(
+    /* webpackChunkName: "inventory-advisor" */
+    '@redhat-cloud-services/frontend-components-inventory-insights/esm'
+));
+const Vulnerabilities = lazy(() => import(
+    /* webpackChunkName: "inventory-vuln" */
+    '@redhat-cloud-services/frontend-components-inventory-vulnerabilities'
+));
+const Compliance = lazy(() => import(
+    /* webpackChunkName: "inventory-compliance" */
+    '@redhat-cloud-services/frontend-components-inventory-compliance/esm'
+));
 import { Spinner } from '@patternfly/react-core';
 import { Provider } from 'react-redux';
 
diff --git a/src/js/inventory/index.js b/src/js/inventory/index.js
index 8cbb80f..7b4a308 100644
--- a/src/js/inventory/index.js
+++ b/src/js/inventory/index.js
@@ -10,22 +10,34 @@ const isEnabled = async () => {
 };
 
 export default async (dependencies) => {
+    let SystemAdvisoryListStore;
+    let SystemCvesStore;
+    let systemProfileStore;
+    let RenderWrapper;
+
     setDependencies(dependencies);
 
-    await import('../inventoryStyles');
-    const invData = await import('@redhat-cloud-services/frontend-components-inventory');
-    const { SystemAdvisoryListStore } = await import(
-        '@redhat-cloud-services/frontend-components-inventory-patchman/dist/cjs/SystemAdvisoryListStore'
-    );
-    const { SystemCvesStore } = await import(
-        '@redhat-cloud-services/frontend-components-inventory-vulnerabilities/dist/cjs/SystemCvesStore'
-    );
-    const systemProfileStore = await import(
-        '@redhat-cloud-services/frontend-components-inventory-general-info/cjs/systemProfileStore'
-    );
-    const RenderWrapper = await import('./RenderWrapper');
+    console.log(await import('customReact'));
 
     const isDetailsEnabled = await isEnabled();
+
+    await import(/* webpackChunkName: "inventory-styles" */ '../inventoryStyles');
+    const invData = await import(/* webpackChunkName: "inventory" */ '@redhat-cloud-services/frontend-components-inventory');
+
+    if (isDetailsEnabled) {
+        SystemAdvisoryListStore = await import(/* webpackChunkName: "inventory-patch-store" */
+            '@redhat-cloud-services/frontend-components-inventory-patchman/dist/cjs/SystemAdvisoryListStore'
+        )?.SystemAdvisoryListStore;
+
+        SystemCvesStore = await import(/* webpackChunkName: "inventory-vuln-store" */
+            '@redhat-cloud-services/frontend-components-inventory-vulnerabilities/dist/cjs/SystemCvesStore'
+        )?.SystemCvesStore;
+        systemProfileStore = await import(/* webpackChunkName: "inventory-gen-info-store" */
+            '@redhat-cloud-services/frontend-components-inventory-general-info/cjs/systemProfileStore'
+        );
+        RenderWrapper = await import(/* webpackChunkName: "inventory-render-wrapper" */ './RenderWrapper');
+    }
+
     return {
         ...invData,
         inventoryConnector: (store) => invData.inventoryConnector(store, isDetailsEnabled ? {
````

**Comment by Hyperkid123** [comment]
@karelhala what is that?

**Comment by karelhala** [comment]
It has multiple things:
* Clears `build/js` after each run
* Builds styles using `styles-loader`
* Adds named chunks for inventory

**Comment by Hyperkid123** [comment]
Does it fix the react-dom issue?

**Comment by Hyperkid123** [comment]
@karelhala chech this out. We have multiple versions of react/react-dom installed. 
```bash
insights-chrome@0.0.0 /home/martin/insights/insights-chrome
├─┬ @redhat-cloud-services/frontend-components-inventory-patchman@0.18.0
│ └── react@16.13.1  deduped
├─┬ @redhat-cloud-services/frontend-components-inventory-vulnerabilities@1.39.13
│ ├─┬ @redhat-cloud-services/frontend-components-pdf-generator@2.2.0
│ │ └── react@16.12.0 
│ └── react@16.13.1  deduped
└── react@16.13.1 

martin:insights-chrome$ npm list react-dom
insights-chrome@0.0.0 /home/martin/insights/insights-chrome
├─┬ @redhat-cloud-services/frontend-components-inventory-patchman@0.18.0
│ └── react-dom@16.13.1  deduped
├─┬ @redhat-cloud-services/frontend-components-inventory-vulnerabilities@1.39.13
│ ├─┬ @redhat-cloud-services/frontend-components-pdf-generator@2.2.0
│ │ └── react-dom@16.12.0 
│ └── react-dom@16.13.1  deduped
└── react-dom@16.13.1 

```

Hyperkid123:
@karelhala so I have fixed the npm scripts. A couple of notes
- the watch could never work before because of the pug. It requires css/js files to be built before so I pre-build these two and only then run the parallel
- also, there is no watch option on the pug build script so I have removed it
- In order to pass additional options to the build:js and it's beta alternative I had to bump it out of the npm-run-all scope (it was propagating the webpack options to the subcommand)

So the start and watch scripts should be working now. I can't think of anything else we need to do to finish this up.

karelhala (approved):

PR #851, opened 2020-06-25
Revert "hide settings and user prefs with no account number" until API is merged

ryelo:
Reverts RedHatInsights/insights-chrome#845


john-dupuy:
Settings is still functional in qa-beta, and the hiding of the cog/alert works. Could we leave the settings code and then do User preferences later? Or would you like to do them both at once?

ryelo:
@john-dupuy i'd rather just revert the entire commit for now and then revert the revert later on

john-dupuy:
Works for me!

PR #852, opened 2020-06-25
hide settings/user-prefs when no account number is present

ryelo:


ryelo:
on hold until: https://github.com/RedHatInsights/entitlements-config/pull/16

PR #855, opened 2020-06-29
change url to do apps and bundles instead

ryelo:


mmenestr:
My check = it shows up as options in the segmentation!
image

In terms of whether it works or not, I tried "currentApp" "is equal to" "Vulnerability" and it has 0 eligible visitors. 
I then tried testing it in context by staging a guide to that segment, and it didn't pop-up. Not sure if thats because this isn't merged, or because I didn't segment it correctly or whether its just not supposed to be able to work yet.

ryelo:
@mmenestr it probably doesn't show up with eligible visitors, because it's not merged and users haven't hit this pendo config yet. Since I was the only one to use it, I should be the only one with data right now.

ryelo:
@shdunning The pendo config won't initialize if there is an error thrown, but I think it's a good idea to at least put a log in there in case it does fail so we can tell if something is broken.

ryelo:
@karelhala agreed! Would we want to add that to FEC or a chrome function?

karelhala:
@ryelo mhh, both are viable. I'd say add it to FEC and surface it in chrome as well.

PR #859, opened 2020-07-01
add styles for focus in masthead

ryelo:
Adds a focus state to items in the masthead

https://projects.engineering.redhat.com/browse/RHCLOUD-6421

PF does allow us to use a `isSelected` prop for these items, but IMO it's easier to just use the CSS with the native `:focus` selector instead of keeping track of the active item.

Screen Shot 2020-07-01 at 11 09 42 AM

Screen Shot 2020-07-01 at 11 14 44 AM


ryelo:
@mmenestr

PR #901, opened 2020-07-27
add ability to change casing in nav item

ryelo:
This should allow people in CSC to ignore the casing for certain nav items. This is requested for applications that need certain casings for trademarks.


src/js/nav/globalNav.js
```diff
@@ -33,7 +33,8 @@ async function getRoutesForApp(app, masterConfig) {
             return (await calculateVisibility(app, subItem, visibility)) && ({
                 ...subItem.title ? {
                     id: subItem.id || '',
-                    title: subItem.title
+                    title: subItem.title,
+                    ignoreCase: subItem.ignoreCase || false
```
karelhala, line 37: I don't think we need this shorthand since `undefined` is treated as `false` as well.

src/js/nav/globalNav.js
```diff
@@ -62,7 +63,8 @@ async function getAppData(appId, propName, masterConfig) {
 
         const routes = await getRoutesForApp(app, masterConfig);
         let appData = {
-            title: app.frontend.title || app.title
+            title: app.frontend.title || app.title,
+            ignoreCase: app.ignoreCase || false
```
karelhala, line 67: One more false

PR #902, opened 2020-07-28
Replace NoAccess with a frontend component

fhlavac:
https://projects.engineering.redhat.com/browse/RHCLOUD-4582

* use frontend component for noAccess
* add loading state


ryelo:
@fhlavac Great job making this a frontend component. I think we can totally take out the `NoAccess` function in chrome because apps can now hide/show navigation items based on permissions, so they should never see this page. @karelhala did I get that right?

karelhala:
@ryelo I'd still keep it in here, since someone might try to access URL they are not allowed to see and the UI would break. I know that the apps should do this instead of us, but I feel like this is nice addition to our apps.

ryelo:
Works for me 😄

PR #906, opened 2020-07-29
Proper localstorage check to null instead of unefined

karelhala:


src/js/inventory/index.js
```diff
@@ -4,7 +4,7 @@ import accountNumbers from './accountNumbers.json';
 const isEnabled = async () => {
     const isExperimentalEnabled = window.localStorage.getItem('chrome:inventory:experimental_detail');
     const { identity } = await insights.chrome.auth.getUser();
-    return isExperimentalEnabled !== undefined ||
+    return isExperimentalEnabled !== null ||
```
PanSpagetka, line 7: Maybe we could have just `isExperimentalEnabled || ...`

karelhala: I've also added a check if `isExperimentalEnabled` is not set to `false`

PR #921, opened 2020-08-04
Add debug option to enable drawer inventory detail

karelhala:


src/js/inventory/index.js
```diff
@@ -9,6 +9,11 @@ const isEnabled = async () => {
         (accountNumbers.includes(identity?.internal?.account_id) && isExperimentalEnabled !== 'false');
 };
 
+const isDrawerEnabled = () => {
+    const drawerEnabled = window.localStorage.getItem('chrome:inventory:experimental_drawer');
+    return drawerEnabled && drawerEnabled !== 'false';
```
PanSpagetka, line 14: I think this is unnecessarily complex `return !!drawerEnabled` or `return Boolean(drawerEnabled}` should be enough.

PR #943, opened 2020-08-31
Update prod-stable to newest version

karelhala:


Hyperkid123 (approved):
@karelhala I have run the chrome locally with Inventory and catalog. It is working as far as I can tell. Do get heaps of prop types error from inventory frontend though :smile:

PR #948, opened 2020-09-03
Optimize build speed

Hyperkid123:
This PR introduces some changes to the webpack build process in order to improve the build speed.

### Build time changes

|command|time no cache|time cache|
|---|---|---|
|build|68s|33s|
|build:js|60s|28s|

**Watch results**
The initial build takes around 28s. After that, the watch mode re-builds the bundle in a couple of seconds. We no longer have to wait for 20+ s in watch mode after each change.

### Changes
- removed hard `minimize:true` from webpack config
  - this was causing that even dev version of the build was minimizing the code which lead to very long build times
  - `minimize` property is now based on the `mode` argument that is passed via npm scripts
- use TerserWebpack plugin as `minimizer`
  - this is one of the most performant webpack minimizers
  - minimizing is done in parallel and these time savings
  - should be building better source maps :tm: 
- use HardSourceWebpackPlugin cache
  - this should be included by default in webpack 5
  - the cache significantly speeds up the watch build
  - we should probably set up Travis to store and read this cache to improve the CI speed we can do that later @karelhala @ryelo what do you think about that?


karelhala (approved):
Trying the watch mode and it's blazingly fast! Thank you!

PR #949, opened 2020-09-04
Add isGlobalFilterEnabled pages blacklist.

Hyperkid123:
Based on an email thread by @iphands 

> I also noticed the component showing up on all pages in Insights...
even places it does not make sense.
>- Remediations
>- Registration Assistant

Any more applications that should hide the global nav? We could also define the list to the nav config, but I think for such a small change it would be to much unnecessary work to get it out fo the YAML.

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -14,6 +14,35 @@ global.MutationObserver = class {
 global.fetch = require('jest-fetch-mock');
 global.window = Object.create(window);
 
+/**
+ * To get around Error: Not implemented: navigation (except hash changes)
+ * Also enables spy and mock implementation on replace. Add missing attributes here.
+ */
+delete window.location;
+window.location = { replace: jest.fn(), pathname: 'https://localhost' };
+
+/**
+ * Mock local storage
+ */
+const localStorageMock = (() => {
+    let store = {};
+    return {
+        getItem: function(key) {
+            return store[key];
+        },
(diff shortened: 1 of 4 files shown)


karelhala (commented):
I'd rather see a function in entry.js and let the apps call it to disable the filter rather than relying fully on URL partials.

Hyperkid123:
> I'd rather see a function in entry.js and let the apps call it to disable the filter rather than relying fully on URL partials.

I had the exact opposite idea. Make this opt-in. But I did not want to just remove it from everywhere.

karelhala:
@Hyperkid123 well the component has to be there for all apps in insights bundle hence the path check and should be forced there rather than opt-in.

Hyperkid123:
> @Hyperkid123 well the component has to be there for all apps in the insights bundle hence the path check and should be forced there rather than opt-in.

Right, that is the namespace. What about the individual apps then. I realize that currently, it's only in beta env or if the flag is set in local storage but it still seems that the component should be entirely opt-in. Just for the simple reason that once we roll this out, there might be more apps that do not use it at all and it would be just sitting there doing nothing and probably just causing confusion and unnecessary bug reports.

As you suggested, we can have a nice API in chrome to turn on/off the rendering of this component. If you want it to be opt-out I will make it so but I think having it opt-in makes more sense at this point IMO.

PR #955, opened 2020-09-09
Move things around, make the navigation calculation run in seperate thread

karelhala:
### Features
* Option to have an array of permission functions
* Reduces chrome bundle size from 3.2 MB to 1.6 MB

### Depracated
* `disabled_on_stable` - this field should no longer be used as consumers can use permissions function
* appNav - no longer possible to pass custom application navigation


src/js/App/Sidenav/Navigation.js
```diff
@@ -12,172 +12,126 @@ import './Navigation.scss';
 
 const basepath = document.baseURI;
 
-const openshiftLinks = {
-    supportcases: {
+const extraLinks = {
```
ryelo, line 15: Can we make sure the Documentation link is always the last one?

karelhala: Sure thing! I've made it to match what was here previously, but I've seen the ticket about matching position of all docs link everywhere.

src/js/App/Header/Loader.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const HeaderLoader = () => <React.Fragment />;
+
+export default HeaderLoader;
```
Hyperkid123, line 5: You can just export  `React.Fragment`. No need for additional node in the VirtualDOM.

src/js/App/Sidenav/SideNav.js
```diff
@@ -0,0 +1,51 @@
...
+    return globalNav ? <Fragment>
+        <AppSwitcher currentApp={activeTechnology}/>
+        <Navigation />
+    </Fragment> : <NavLoader />;
+};
+
+SideNav.propTypes = {
+    activeTechnology: PropTypes.string,
+    globalNav: PropTypes.arrayOf(PropTypes.shape({
+        [PropTypes.string]: PropTypes.any
```
Hyperkid123, line 34: I think you should list all used attributes instead of creating a generic object. If you want that just use `PropTypes.object`.

src/js/nav/globalNav.js
```diff
@@ -10,24 +13,24 @@ export let getNavFromConfig = async (masterConfig) => {
     }, {});
 };
 
+const isCurrVisible = (permissions) => Promise.all(
+    flatMap([permissions], ({ method } = {}) => (
+        // (null, undefined, true) !== false
```
Hyperkid123, line 18: leftover comment?

src/js/nav/globalNav.js
```diff
@@ -10,24 +13,24 @@ export let getNavFromConfig = async (masterConfig) => {
...
+        visibilityFunctions?.[method]?.(...permissions.args || []) !== false || false
```
Hyperkid123, line 19: I don't think the `|| false` is necessary here. x !== `false` will always return boolean value. Or am I crazy :laughing:  I must say this makes me think much harder than it should.

src/js/nav/globalNav.js
```diff
@@ -53,35 +54,42 @@ async function getRoutesForApp(app, masterConfig) {
...
+            ...routes?.length > 0 && { [propName]: routes }
+        };
     }
 }
+
+export async function loadNav(yamlConfig) {
+    const groupedNav = await getNavFromConfig(safeLoad(yamlConfig));
+
+    const [active, section] = [getUrl('bundle'), getUrl('app')];
+    const globalNav = (groupedNav[active] || groupedNav.insights).routes;
```
Hyperkid123, line 84: This has the potential to crash. I would add `?.`

src/js/nav/globalNav.js
```diff
@@ -53,35 +54,42 @@ async function getRoutesForApp(app, masterConfig) {
...
+    let activeSection = globalNav.find(({ id }) => id === section);
```
Hyperkid123, line 85: `activeSection` is never re-assigned.

Hyperkid123 (commented):
There is a definite improvement. I took a look at the react profiler and these are some of the observations.

### Before
We have 3 renders on the initial load. The final render is committed around 3s. The number will never be precise.

### After
We now have 4 renders but committed at around 1.6s so the cut on initial render time about 50% on my machine.

The biggest "bottleneck" i see is the transition from loaders to the actual navigation. That is mostly caused by evaluating the YAML file which seems to be quite an expensive operation. @karelhala could we just store the YAML or the nav JSON structure to sessionStorage and give it a couple of hours of timeout? This would further decrease time to first interactive when switching between apps.

src/js/nav/globalNav.js
karelhala: Not really, this is to show what is actually evalueted here because the boolean logic is slightly more complex than it could be for new commers.

src/js/nav/globalNav.js
karelhala: Yeah, the false is not necessary here, the logic is `if visibility function return something else then false return true, otherwise false`. But that's correct and doesn't need the `|| false`

karelhala:
@Hyperkid123 I was thinking about storing the calculated JSON in cache as well, but didn't wanted to bloat this PR even more. But let's do it to improve loads even more!

src/js/App/Header/Loader.js
karelhala: I wanted to use some fancy loader similiar to navigation, but had hard time doing it. Let's nuke this file and use just Fragment there.

Hyperkid123:
> @Hyperkid123 I was thinking about storing the calculated JSON in cache as well, but didn't wanted to bloat this PR even more. But let's do it to improve loads even more!

We can do it in different PR if you want. It will make it easier to review and test,

Hyperkid123 (dismissed):

src/js/App/Sidenav/__snapshots__/Navigation.test.js.snap
```diff
@@ -244,7 +244,7 @@ exports[`Navigation should render corectly 1`] = `
     ariaRightScroll="Scroll right"
   >
     <NavExpandable
-      className="ins-m-navigation-align"
```
Hyperkid123, line 247: This css is required to fix the navigation text left alignment. Kan you keep it? (introduced in https://github.com/RedHatInsights/insights-chrome/pull/958)

karelhala: Sure thing!

Hyperkid123 (approved):

PR #956, opened 2020-09-10
Use encodeURIComponent on all tag partials

karelhala:
As requested by @AllenBW let's encode all tag partials.


AllenBW:
Advisor changed its api, now we don't want anything encoded 🙃 we might want to leave this up to each of the apps if they need it 🤔

karelhala:
I'll probably change it that if you pass 2nd argument to this function it'll encode the partials for you.

AllenBW:
I like it! reallllll handy like 🙇

Hyperkid123 (approved):

PR #959, opened 2020-09-14
Introduce chrome cache

karelhala:


src/js/nav/globalNav.js
```diff
@@ -77,8 +77,12 @@ async function getAppData(appId, propName, masterConfig) {
     }
 }
 
-export async function loadNav(yamlConfig) {
-    const groupedNav = await getNavFromConfig(safeLoad(yamlConfig));
+export async function loadNav(yamlConfig, cache) {
+    let groupedNav = await cache.getItem('navigation');
```
Hyperkid123, line 81: @karelhala when is this cache supposed to be populated? I see this undefined on each page refresh.

karelhala: @Hyperkid123 mhh, that is strange. It should be populated from `entry.js` I checked it locally and it looks populated.

Hyperkid123: What key is supposed to be store in `localSotarge`? I have it stored under `chrome:global-filter/chrome`.

karelhala: Sure thing, tried clearing cache and it still works, you should be able to find it under `${session}-chrome/chrome` where expires is set and in `data` you should see navigation. I don't want to polute `localStorage` more than we did so I think we could use one localForge store for all chrome related non API data. I'd like to add global filter values to it once we agree on the shape of it and such.

Hyperkid123: This is what I have in local storage
```
chrome:global-filter/6e9e3e69-7364-4bbc-94c6-45e2f5ffe7c6: "{}"
chrome:global-filter/chrome: "{"data":{"navigation":{"ansible":{"title":"Red Hat Ansible Automation Platform","id":"ansible","routes":[{"tit
```

If I log the cache name from the nav should be extracted its `chrome:global-filter/6e9e3e69-7364-4bbc-94c6-45e2f5ffe7c6` which is empty object in my local storage.

Hyperkid123: @karelhala ok it works now. I had to logout and login to make it work. :shrug: possible bug?

Hyperkid123 (approved):

PR #961, opened 2020-09-14
Allow users to open support cases

ryelo:


Hyperkid123 (approved):

src/js/createCase.js
```diff
@@ -0,0 +1,83 @@
...
+    const currentProduct = registerProduct();
+
+    log('Creating a support case');
+
+    fetch(`https://access.${window.insights.chrome.isProd ? '' : 'qa.'}redhat.com/hydra/rest/se/sessions`, {
+        method: 'POST',
+        headers: {
+            Accept: 'application/json',
+            Authorization: `Bearer ${Cookies.get('cs_jwt')}`,
+            'Content-Type': 'application/json'
```
karelhala, line 48: You should also add `credentials: 'same-origin'` so this request works in FF - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

john-dupuy: When I tried this in FF `78.2.0esr` the request did not work due to this issue ^, and the new tab did not open. However, in FF `80.0.1` it did work.

john-dupuy:
When testing locally I noticed that I hit this error in the console, and it took me to a page of my support cases rather than the wizard/dialog for opening a new support case. 

```
Uncaught DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'https://<qa-access-redhat>/support/cases/precache-manifest.xxxxxxxxxxxxxxxxxxxx.js' failed to load.
    at https://<qa-access-redhat>/support/cases/service-worker.js:1:1
(anonymous) @ service-worker.js:1
```

Hyperkid123:
> When testing locally I noticed that I hit this error in the console, and it took me to a page of my support cases rather than the wizard/dialog for opening a new support case.
> 
> ```
> Uncaught DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'https://<qa-access-redhat>/support/cases/precache-manifest.xxxxxxxxxxxxxxxxxxxx.js' failed to load.
>     at https://<qa-access-redhat>/support/cases/service-worker.js:1:1
> (anonymous) @ service-worker.js:1
> ```

Can you check if the network request was failing because of an invalid SSL certificate?

john-dupuy:
After exploring a bit with Martin, it looks like that request was indeed failing from an invalid SSL cert.

Hyperkid123:
> After exploring a bit with Martin, it looks like that request was indeed failing from an invalid SSL cert.

If you want to know how to test it on ci env, just message me or John. You have nuke your browser a bit and I am not sure if I want to post it there. It should work with no problems on prod env since we have a proper certificate over there.

karelhala (approved):

karelhala (approved):

src/js/createCase.js
```diff
@@ -0,0 +1,84 @@
...
+    'Red Hat OpenShift Cluster Manager',
+    'Red Hat Insights',
+    'Subscription Watch'
+];
+
+// List of products that are individual apps
+const APP_PRODUCTS = [
+    { id: 'automation-hub', name: 'Ansible Automation Hub' },
+    { id: 'automation-analytics', name: 'Ansible Automation Analytics' },
+    { id: 'migrations', name: 'Red Hat Migration Analytics' },
```
john-dupuy, line 18: @ryelo I think this one should be just `Migration Analytics`, I don't see `Red Hat Migration Analytics` in the product dropdown.

src/js/createCase.js
```diff
@@ -0,0 +1,84 @@
...
+    { id: 'cost-management', name: 'Red Hat Cost Management' }
```
john-dupuy, line 19: @ryelo Stepping though the `registerProduct()` it looks like this one is actually a `BUNDLE_PRODUCT`

PR #962, opened 2020-09-14
Rework JWT failure flow + update sentry logging

ryelo:
The call happens before `updateToken` so it generates a lot of noise


karelhala (dismissed):

cben:
LGTM (though I'm not very familiar with this area) :+1:

The PR title may be inaccurate now — besides moving a noisy Sentry call to a better place, this will also change behavior: it'll redirect to login when previously it'd simly fail, right?

cc @elad661 — I think redirecting to login that's a positive change for openshift clould manager, but you have more context...

elad661:
@cben yeah, this makes sense from my perspective.

Redirecting to login if an exception occurred (this should be a rare case anyway) is the right approach, and having less Sentry noise will allow us to more easily find actual issues.

LGTM.

src/js/jwt/jwt.js
```diff
@@ -288,21 +288,29 @@ function refreshTokens() {
...
+
+      if (refreshed) {
+        log('Token was successfully refreshed');
+      } else {
+        log('Token is still valid, not updating');
+      }
+    })
+    .catch((err) => {
+      log(err);
+      Sentry.captureException(new Error('Update token failed'));
```
elad661, line 310: maybe it'd be useful to log the actual exception in Sentry so we'll have more context if it ever happens?

karelhala (approved):

karelhala (approved):

PR #964, opened 2020-09-15
Make all workloads sticky with no remove option

karelhala:
animation

### DIFF
--- a/src/js/App/GlobalFilter/constants.js
+++ b/src/js/App/GlobalFilter/constants.js
@@ -1,4 +1,5 @@
 import { deleteLocalStorageItems } from '../../utils';
+import omit from 'lodash/omit';
 export const GLOBAL_FILTER_KEY = 'chrome:global-filter';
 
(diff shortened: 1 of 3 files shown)


karelhala:
ping @terezanovotna

terezanovotna:
Nice work 👏👏

src/js/App/GlobalFilter/constants.js
```diff
@@ -15,6 +16,14 @@ export const workloads = [
     }
 ];
 
+export const selectWorkloads = () => ({
+    [workloads?.[0]?.tags?.[0]?.tag?.key]: {
```
Hyperkid123, line 20: So it is acceptable to have an `undefined` key? Should we have some fallback like `default` or something?

karelhala: Good point! I'll add `All workloads` as default.

Hyperkid123 (approved):

PR #968, opened 2020-09-15
Perform full login if token expires when getToken is called

karelhala:


Hyperkid123:
@karelhala is this still relevant?

karelhala:
We heaven't heard from the app that was experiencing issues with stale tokens for a while. Let's close this one, we can always reopen it should we need it.

PR #970, opened 2020-09-16
Depracate navigation function and console log erro

karelhala:
This PR fixes issues in apps that are still using this function.

### DIFF
--- a/src/js/entry.js
+++ b/src/js/entry.js
@@ -51,7 +51,7 @@ export function chromeInit(libjwt) {
     const { store, middlewareListener, actions } = spinUpStore();
 
     // public API actions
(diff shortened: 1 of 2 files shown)


src/js/entry.js
```diff
@@ -84,7 +84,7 @@ export function chromeInit(libjwt) {
         identifyApp: (data) => Promise.all([jwtResolver, navResolver]).then(
             () => identifyApp(data, store.getState().chrome.globalNav)
         ),
-        navigation: appNav,
+        navigation: () => console.error('Don\'t use this function, it has been deprecated!'),
```
ryelo, line 87: Can you include the name of the function here?

`Don't use insights.chrome.navigation, it has been deprecated`?

ryelo (approved):

PR #975, opened 2020-09-17
Guard and cache fetchPermission chrome API.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-9200

### Changes
- adds a guard for `fetchPermission` chrome API call
  - invoking fetchPermission x > 1 times will always result only in one API call.
- use maximum page size for the `fetchPermission` config
  - previously the page size was 25, now it's 100
  - this change alone reduces the number of calls 4x
- cache the `fetchPermission` call
  - this will remove many rbac API calls when switching between apps
  - the cache is invalidated after 10 minutes

### DIFF
--- a/src/js/entry.js
+++ b/src/js/entry.js
@@ -16,7 +16,7 @@ import { visibilityFunctions } from './consts';
 import Cookies from 'js-cookie';
 import logger from './jwt/logger';
 import sourceOfTruth from './nav/sourceOfTruth';
-import { fetchPermissions } from './rbac/fetchPermissions';
+import { createFetchPermissionsWatcher } from './rbac/fetchPermissions';
 import { getUrl } from './utils';
 import flatMap from 'lodash/flatMap';
 import { headerLoader } from './App/Header';
@@ -54,7 +54,6 @@ export function chromeInit(libjwt) {
     const { identifyApp, appNavClick, clearActive, appAction, appObjectId, chromeNavUpdate } = actions;
 
     let chromeCache;
-
     // Init JWT first.
     const jwtResolver = libjwt.initPromise
     .then(async () => {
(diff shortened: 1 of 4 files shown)


src/js/rbac/fetchPermissions.js
```diff
@@ -23,3 +22,28 @@ export const fetchPermissions = (userToken, app = '') => {
         }})
     .catch(error => log(error));
 };
+
+export const createFetchPermissionsWatcher = (getUser, libjwt) => {
+    let currentCall = undefined;
+    const cache = new CacheAdapter(
```
karelhala, line 28: I wouldn't create new cache, I'd rather share chrome cache from all over places, meaning passing it to `createFetchPermissionsWatcher`
```JS
const createFetchPermissionsWatcher = (cache) => {
  let currentCall = undefined;
    return async (userToken, app = '') => {
        if (insights.chrome.getBundle() === 'openshift') {
            return Promise.resolve([]);
        }
        const permissions = await cache.getItem('permissions');
        if (permissions) {
            return permissions;
        }
        if (typeof currentCall === 'undefined') {
            currentCall = getUser().then(() => fetchPermissions(userToken, app).then((data) => {
                currentCall = undefined;
                cache.setItem('permissions', data);
                return data;
            }));
        }
        return currentCall;
    };
}
```

src/js/entry.js
```diff
@@ -128,27 +128,29 @@ export function chromeInit(libjwt) {
...
+
+        qe.init();
+
+        return libjwt.initPromise
+        .then(libjwt.jwt.getUserInfo)
+        .catch(() => {
+            libjwt.jwt.logoutAllTabs();
+        });
+    };
+    const fetchPermissions = createFetchPermissionsWatcher(getUser, libjwt);
```
karelhala, line 146: I'd change this to consume only cahce
```JS
const fetchPermissions = getUser().then(() => createFetchPermissionsWatcher(cache));
```

karelhala: In order to make it work correctly you'd have to moce `let chromeCache` on line 56 move above chrome bootstrap.

src/js/rbac/fetchPermissions.js
Hyperkid123: @kate that would be perfect however that will take a little bit more effort than just lifting to a new file which will delay this fix. I would like to do that in a different PR.

karelhala (dismissed):

karelhala (approved):

Hyperkid123:
@karelhala I have forgotten to update the tests.

Hyperkid123:
@karelhala also using the global cache does not work :cry: The endpoint is called before the cache is created... Let's use the local cache for now and I will come up with a solution to make the chrome cache global in a different PR.
This is the error. The cache is, unfortunately `undefined`.
```
fetchPermissions.js:69 Uncaught (in promise) TypeError: Cannot read property 'getItem' of undefined
    at _callee$ (fetchPermissions.js:69)
    at tryCatch (runtime.js:63)
    at Generator.invoke [as _invoke] (runtime.js:293)
    at Generator.eval [as next] (runtime.js:118)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)
    at eval (asyncToGenerator.js:32)
    at new Promise (<anonymous>)
    at eval (asyncToGenerator.js:21)
    at eval (fetchPermissions.js:101)
```

karelhala (approved):

PR #978, opened 2020-09-21
Fix navigation after refactoring

karelhala:


Hyperkid123 (commented):
@karelhala  I see that the nav is s still generating incorrect URL:

Screenshot from 2020-09-21 17-23-17

Hyperkid123 (commented):
@karelhala I am still getting redirected back to the "top level" app. See the gif. The URL is correct though. Did the navigation API change at all? Maybe we need to fix it inside the app itself.
approval redirect

karelhala:
@Hyperkid123 I've tried it with app deployed on cloud instead of locally and it works as expected
animation

However with local catalog running it forced ridirect to approval, I think it's caused by the app running locally

Hyperkid123:
@karelhala yeah that is probably it.

Hyperkid123 (approved):

PR #983, opened 2020-09-23
Make chrome a closure

Hyperkid123:
- added one singular chrome cache instance
- add async call buffer that will wait with promise execution until
chrome cache is initialized
  - we can now wrap async chrome functions to with a cache buffer that will only execute these functions after chrome cache was created
- permissions are now cached in chrome cache

### TODO
- [x] tests


src/js/chrome.js
```diff
@@ -28,14 +29,4 @@ libjwt.initPromise.then(() => {
...
-        ...window.insights.chrome,
-        ...chromeInit(libjwt)
-    };
-};
-
-window.insights = {
-    ...window.insights,
-    ...bootstrap(libjwt, init)
-};
+window.insights = createChromeIntance(libjwt, window.insights);
```
ryelo, line 32: should this be `createChromeInstance`?

Hyperkid123: This is basically a constructor which will return the object. Its technically a class but without the unnecessary class boilerplate that JS ads to classes.

ryelo: I just meant the spelling 😄

Hyperkid123: :facepalm:

Hyperkid123: OK its fixed :D sorry about that my eyes don't see that :unicorn: :rainbow:

src/js/chrome/create-chrome.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param {object} insights existing insights instance
+ */
+const createChromeInstance = (jwt, insights) => {
+    const { actions: { chromeNavUpdate } } = spinUpStore();
+    const libjwt = jwt;
+    const chromeInstance = {
+        cache: undefined
+    };
+
+    const jwtResolver = initializeJWT(libjwt, chromeInstance);
```
karelhala, line 21: Might be a better to use then instead of mutating `chromeInstance.cache` when init is done
```JS
let chromeCache;
const jwtResolver = initializeJWT(libjwt);
jwtResolver.finally(() => {
  chromeCache = new CacheAdapter(
    'chrome-store',
    `${decodeToken(libjwt.jwt.getEncodedToken())?.session_state}-chrome-store`
  );
})
```

src/js/chrome/create-chrome.js
```diff
@@ -0,0 +1,81 @@
...
+         * Wait for JWT initialization to happen and cache initialization in chrome instance
+         */
+        return (...args) => jwtResolver.then(() => fn(...args));
+    };
+
+    const fetchPermissions = bufferAsyncFunction(createFetchPermissionsWatcher(chromeInstance));
+
+    const chromeFunctions = bootstrap(libjwt, init, getUser);
+
+    chromeFunctions.chrome.getUserPermissions = (app = '') => getUser().then(() => fetchPermissions(libjwt.jwt.getEncodedToken(), app));
```
karelhala, line 74: It would be cleaner with `async/await`
```JS
chromeFunctions.chrome.getUserPermissions = async (app = '') => {
  await getUser();
  return (await fetchPermissions)?.(libjwt.jwt.getEncodedToken(), app));
}
```

src/js/chrome/create-chrome.js
karelhala: Or going full promise
```JS
const chromeCache = jwtResolver.finally(() => new CacheAdapter(
    'chrome-store',
    `${decodeToken(libjwt.jwt.getEncodedToken())?.session_state}-chrome-store`
  )
);
```

And whenever `chromeCache` is called wrap it with either `then` or `await`

Hyperkid123: This won't work the chromeCache will not mutate in functions where the cache is passed as a parameter. It will create a copy instead of keeping the pointer.

src/js/chrome/create-chrome.js
```diff
@@ -0,0 +1,81 @@
...
+        if (chromeInstance.cache) {
+            return fn;
+        }
+        /**
+         * Wait for JWT initialization to happen and cache initialization in chrome instance
+         */
+        return (...args) => jwtResolver.then(() => fn(...args));
+    };
+
+    const fetchPermissions = bufferAsyncFunction(createFetchPermissionsWatcher(chromeInstance));
```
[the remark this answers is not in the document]
Hyperkid123: I will try it. There is some funky stuff with the concurrency.

src/js/chrome/create-chrome.js
karelhala: Ah, yeah. The problem with pointers and values of JS :disappointed: I don't think promisifying it won't help us with readability either... Keep it as is

src/js/chrome/create-chrome.js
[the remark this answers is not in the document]
Hyperkid123: @karelhala nope this won work either. It would have to be like:
```jsx
const fetchPermissions = () => jwtResolver.then(() => createFetchPermissionsWatcher(chromeInstance));
```
But then there is no way to pass arguments to the actuall `fetchPermissions` function.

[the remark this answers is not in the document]
karelhala: And how about
```JS
const fetchPermissions = (...props) => jwtResolver.then(() => createFetchPermissionsWatcher(chromeInstance)(...props));
```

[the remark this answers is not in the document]
Hyperkid123: You would create multiple instances of the `fetchPermissions` which will remove the guard that is watching simultaneous permission calls and that would cause duplicate API calls.

src/js/chrome/initialize-jwt.js
```diff
@@ -0,0 +1,34 @@
...
+        chromeCache = new CacheAdapter(
+            'chrome-store',
+    `${decodeToken(libjwt.jwt.getEncodedToken())?.session_state}-chrome-store`
+        );
+        chromeInstance.cache = chromeCache;
+    } catch (_e) {
+        if (allowUnauthed()) {
+            actions.userLogIn(false);
+        }
+        chromeInstance.cache = chromeCache;
```
karelhala, line 29: This could lead to `chromeInstance.cache = undefined` perhaps moving cache init bellow `catch` would be better.

karelhala (commented):
Just one final question, but overall looks good!

src/js/chrome/initialize-jwt.js
Hyperkid123: Yeah.

karelhala (approved):

PR #987, opened 2020-09-25
Wrap remediations with react-dom renderer

karelhala:
### Remediations wizard loading
Change how remediations is loaded, since we want to use mounted components in all details we also have to mount Remediations button to new React element. Otherwise there are problems with mismatch React version.

### Change remediations wizard open
In order to simplify Remediations use ref instead of generated instance in component, this is more React like

### Deps
* [x] merge after https://github.com/RedHatInsights/frontend-components/pull/751

Had to downgrade version of PF-react core because they had an error in Paginations component with missing component breaking Patch UI once all details were enabled


src/js/remediations/index.js
```diff
@@ -0,0 +1,29 @@
...
+        openWizard: async (data, basePath) => {
+            const wizardRef = await deferred.promise;
+            remediationsData.openWizard(data, basePath, wizardRef);
+        },
+        // eslint-disable-next-line react/display-name
+        RemediationWizard: () => {
+            return <RenderWrapper.default
+                cmp={remediationsData.RemediationWizard}
+                onAppRender={(wizardRef) => {
+                    console.log(wizardRef);
```
Hyperkid123, line 23: is this log required?

karelhala: Good point!

Hyperkid123 (approved):

PR #1001, opened 2020-10-06
Disable permission cache

Hyperkid123:
Without proper cache invalidation in place, there is no way of reliably return required data. The cache will be disabled for now until we can figure out how to invalidate the data properly.


john-dupuy (approved):

karelhala (dismissed):
You can remove these comments about cahce, we have github to track the history anyways.

karelhala (approved):

PR #912, opened 2020-07-30
capitalize skeleton and spinner import names

jctanner:
image

These files are Sentence cased ...

```
[jtanner@corsair test]$ npm install @redhat-cloud-services/frontend-components
npm WARN saveError ENOENT: no such file or directory, open '/home/jtanner/workspace/github/RedHatInsights/test/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/home/jtanner/workspace/github/RedHatInsights/test/package.json'
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of @patternfly/react-table@>=4.5.7 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of @patternfly/react-core@>=4.18.5 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of @patternfly/react-icons@>=4.3.5 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of classnames@^2.2.5 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of prop-types@^15.6.2 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of react@^16.5.1 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of react-dom@^16.5.1 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of react-redux@^5.0.7 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components@2.3.11 requires a peer of react-router-dom@^4.2.2 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components-utilities@2.1.1 requires a peer of @patternfly/react-core@>=4.18.5 but none is installed. You must install peer dependencies yourself.
npm WARN @redhat-cloud-services/frontend-components-utilities@2.1.1 requires a peer of react@^16.5.1 but none is installed. You must install peer dependencies yourself.
[...]

PR #1012, opened 2020-10-09
Fix 404 spinner, skeleton, conditional filter.

Hyperkid123:
In addition to https://github.com/RedHatInsights/insights-chrome/pull/912 we also needed to fix the actual import path.

### Changes
- Added node importer to sass. I have no clue why it was working for PF stuff, but PF is extra special in many ways that I won't even try to guess (generated styles for PF remain unchanged). Node importer is a standard way of importing node modules into sass. Normally it's handled by webpack internally, but we do not use webpack to compile our style assets.


karelhala (commented):
Good approach! I've used something similiar in our parcel build, can you change all `node_modules/` imports in chrome.scss to `~`?

Hyperkid123:
@karelhala I will try. Let's hope it won't break it all. I was honestly afraid to change it. You never know what the PF stuff will do :smile:

ryelo (approved):

PR #1013, opened 2020-10-10
List first 10 SIDs and show aplicable SAP workload

karelhala:
### List SIDs in global filter

New endpoint is available in inventory API to list all available SIDs on account. This PR uses such endpoint, passes already applied filters and shows them above tags.

Screenshot from 2020-10-12 15-14-52
Screenshot from 2020-10-12 15-15-16

### Move global filter reducers

Since global filter is getting bigger, let's split it from chrome reducers into seperate namespace.

### TODO
* [x] add some tests


src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -84,6 +107,13 @@ const GlobalFilter = () => {
         }
     }, [selectedTags, isAllowed()]);
 
+    useEffect(() => {
+        const sapTag = workloads?.[0]?.tags?.[1];
+        if (sapCount !== undefined && sapTag) {
```
Hyperkid123, line 112: I would maybe go for `typeof sapCount === 'number'` instead. This will still allow strings, arrays, etc.

src/js/App/GlobalFilter/constants.js
```diff
@@ -84,3 +88,27 @@ export const generateFilter = async () => {
 
     return [data, currToken];
 };
+
+export const flatTags = (filter, encode = false, format = false) => {
```
Hyperkid123, line 92: This one seems expensive. Is it possible to add some memoize wrapper?

src/js/App/GlobalFilter/constants.js
```diff
@@ -84,3 +88,27 @@ export const generateFilter = async () => {
...
+    const { Workloads, SID, ...tags } = filter;
+    const mappedTags = flatMap(
+        Object.entries({ ...tags, ...!format && { Workloads, SID } } || {}),
+        ([namespace, item]) => Object.entries(item || {})
```
Hyperkid123, line 96: I think it is more efficient to filter out entries with no values instead of replacing them with empty objects. You could do it inside the filter bellow and remove one iteration.

src/js/App/GlobalFilter/tagsApi.js
```diff
@@ -32,6 +22,54 @@ export function getAllTags({ search, activeTags, registeredWith } = {}, paginati
...
+export function getAllWorkloads({ activeTags, registeredWith } = {}, pagination = {}) {
+    const [workloads, SID, selectedTags] = flatTags(activeTags, false, true);
+
+    const filter = {
+        system_profile: {
+            ...workloads?.SAP?.isSelected && { sap_system: true },
+            sap_sids: SID
+        }
+    };
+    console.log(filter, generateFilter(filter), 'ffffmmm');
```
Hyperkid123, line 63: console

src/js/App/GlobalFilter/tagsApi.js
```diff
@@ -32,6 +22,54 @@ export function getAllTags({ search, activeTags, registeredWith } = {}, paginati
...
+            query: generateFilter(filter)
+        }
+    );
+}
+
+export function getAllWorkloads({ activeTags, registeredWith } = {}, pagination = {}) {
+    const [workloads, SID, selectedTags] = flatTags(activeTags, false, true);
+
+    const filter = {
+        system_profile: {
```
Hyperkid123, line 58: This object structure is repeating a lot. Can you create a function that will return this shape?

src/js/App/GlobalFilter/constants.js
karelhala: Sure thing!

src/js/App/GlobalFilter/constants.js
karelhala: Not really, this one is to check that the item is defined, otherwise it was throwing errors that `item` is not defined. Thiw was caused by default values for `workloads` and `SID`, they can be undefined and so the line 95 will expand to
```JS
Object.entries({
  SID: undefined,
  workloads: undefined
}).map(([ namespace, item ]) => Object.entries(item)).filter() // here `item` can be undefined. 
```

We could pass default values on line 93, but that'll boil down to same thing as in here. Or we could assign `Workloads` or `SID` only when defined, but that feels like more expensive than running `Object.entries` on empty object.

src/js/App/GlobalFilter/GlobalFilter.js
karelhala: Good point.

src/js/App/GlobalFilter/constants.js
Hyperkid123: Ah got it. :+1:

src/js/App/GlobalFilter/constants.js
karelhala: The memoize resolver is a bit complex, but I think it'll help us when resolving multiple values. For instance on dashboard this function will be called multiple times.

Hyperkid123:
@karelhala :test_tube: :boom:

Hyperkid123 (approved):

PR #1014, opened 2020-10-13
Use common linter settings

Hyperkid123:


ryelo (dismissed):

Hyperkid123:
:no_good: these conflicts

ryelo (commented):
I guess this is an overall comment, but it seems like there is a rule that adds a `,` at the end of a list, even if there is nothing to follow. I'm not sure if that's intentional or not, but IMO it'd be better to only add it if there's an item to follow

Hyperkid123:
That is a prettier config and it is added on multiline arrays/objects. It can be configured I think (not all prettier configurations can). O honestly don't mind either way since the IDE will fix fit for me :smile:.

karelhala (dismissed):

karelhala (approved):

src/js/App/RootApp.js
```diff
@@ -3,73 +3,58 @@ import PropTypes from 'prop-types';
...
+    <Fragment>
+      <div
+        className="pf-c-drawer__content"
+        data-ouia-page-type={activeApp || appId}
+        data-ouia-page-group={activeLocation}
+        data-ouia-safe="true"
+        {...(pageAction && { 'data-ouia-page-action': pageAction })}
+        {...(pageObjectId && { 'data-ouia-page-object-id': pageObjectId })}
+      >
+        <div>
```
jtomasek, line 21: Addition of this div breaks the PF Page + PageSection layout which uses flexbox. This currently breaks the Assisted Installer in OCM. It needs to be replaced with React.Fragment

karelhala: This PR uses new linter, you might be refering to this PR https://github.com/RedHatInsights/insights-chrome/pull/1008

PR #1018, opened 2020-10-14
Allow multiple tabs in tags modal for global filter

karelhala:
### Add SID tab to tags modal

This uses new feature from frontend components with tabs in tags modal. Will require FEC package update.

### DIFF
--- a/src/js/App/GlobalFilter/GlobalFilter.js
+++ b/src/js/App/GlobalFilter/GlobalFilter.js
@@ -172,7 +172,7 @@ const GlobalFilter = () => {
           }}
           onApplyTags={(selected, sidSelected) => {
             setValue(() =>
-              [...selected, ...sidSelected]?.reduce(
+              [...(selected || []), ...(sidSelected || [])]?.reduce(
                 (acc, { namespace, key, value }) => updateSelected(acc, namespace, key, value, true),
                 selectedTags
               )
--- a/src/js/App/GlobalFilter/GlobalFilter.js
+++ b/src/js/App/GlobalFilter/GlobalFilter.js
@@ -170,8 +170,13 @@ const GlobalFilter = () => {
             }
             setIsOpen(false);
           }}
-          onApplyTags={(selected) => {
-            setValue(() => selected.reduce((acc, { namespace, key, value }) => updateSelected(acc, namespace, key, value, true), selectedTags));
+          onApplyTags={(selected, sidSelected) => {
+            setValue(() =>
+              [...selected, ...sidSelected]?.reduce(
+                (acc, { namespace, key, value }) => updateSelected(acc, namespace, key, value, true),
+                selectedTags
+              )
+            );
           }}
         />
       )}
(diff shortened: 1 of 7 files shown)


src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -170,8 +170,13 @@ const GlobalFilter = () => {
             }
             setIsOpen(false);
           }}
-          onApplyTags={(selected) => {
-            setValue(() => selected.reduce((acc, { namespace, key, value }) => updateSelected(acc, namespace, key, value, true), selectedTags));
+          onApplyTags={(selected, sidSelected) => {
+            setValue(() =>
+              [...selected, ...sidSelected]?.reduce(
```
Hyperkid123, line 175: This spread will always produce at least an empty array. So no need for the `?`. I would be more worried about `selected` or `sidSelected` being `undefined`. If that is the case can we add a default value `[]` to the arguments?

src/js/App/GlobalFilter/TagsModal.js
```diff
@@ -1,20 +1,37 @@
...
+  );
+  const { sidLoaded, sidCount, sidPage, sidPerPage } = useSelector(
+    ({ globalFilter: { sid } }) => ({
+      sidLoaded: sid?.isLoaded,
+      sidCount: sid?.total || 0,
+      sidPage: sid?.page || 1,
+      sidPerPage: sid?.perPage || 10,
+    }),
+    shallowEqual
+  );
```
Hyperkid123, line 32: These two selectors are exactly the same. On the prefix is different. We could use a common function for it.

src/js/App/GlobalFilter/TagsModal.js
```diff
@@ -36,66 +53,113 @@ const TagsModal = ({ isOpen, filterTagsBy, onApplyTags, toggleModal, selectedTag
...
+              {
+                registeredWith: filterScope,
+                activeTags: selectedTags,
+                search: filterBy,
+              },
+              pagination
+            )
+          ),
+      ]}
+      columns={[[{ title: 'Name' }, { title: 'Value' }, { title: 'Tag sources' }], [{ title: 'Value' }]]}
```
Hyperkid123, line 158: There two columns with a `Value` title. Is that by design?

karelhala: That's actually two arrays of columns, one has 3 columns the other one has just `Value`

src/js/App/GlobalFilter/TagsModal.js
karelhala: Good point.

src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -170,8 +170,13 @@ const GlobalFilter = () => {
             }
             setIsOpen(false);
           }}
-          onApplyTags={(selected) => {
-            setValue(() => selected.reduce((acc, { namespace, key, value }) => updateSelected(acc, namespace, key, value, true), selectedTags));
+          onApplyTags={(selected, sidSelected) => {
+            setValue(() =>
+              [...(selected || []), ...(sidSelected || [])]?.reduce(
```
Hyperkid123, line 175: The `?` still lurks here :smile:

Hyperkid123 (approved):

PR #1022, opened 2020-10-19
Use correct selector for global filter hidden

karelhala:


src/js/App/RootApp.js
```diff
@@ -53,7 +53,7 @@ RootApp.propTypes = {
   globalFilterHidden: PropTypes.bool,
 };
 
-function stateToProps({ chrome: { activeApp, activeLocation, appId, pageAction, pageObjectId, globalFilterHidden } }) {
+function stateToProps({ chrome: { activeApp, activeLocation, appId, pageAction, pageObjectId }, globalFilter: { globalFilterHidden } }) {
```
Hyperkid123, line 56: This killed tests. We will probably need some default value.

karelhala: Good point!

Hyperkid123 (approved):

PR #1029, opened 2020-10-20
Enable dynamic navigation items.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-9479

I have added two there are two options:
- Each item can use the `apiRequest` API call to check its visibility
- The root item can have a dynamic attribute. That will make an API call and expects the resolved value to be a partial nav schema.
  - I don't think that this option is reasonable because it will make a mess in the navigation, but it's still an option

Even if we delete the second option I would still like to keep the nav item extracted into a separate component because it will save us some rendering cycles.

I have also changed the nav keys to be the actual id, not the array index. Again that should give react better change to prevent unnecessary renders.

### TODO
- [x] docs

### DIFF
--- a/src/js/consts.js
+++ b/src/js/consts.js
@@ -44,14 +44,12 @@ export const visibilityFunctions = {
     const userPermissions = await insights.chrome.getUserPermissions();
     return userPermissions && permissions.every((item) => userPermissions.find(({ permission }) => permission === item));
   },
-  apiRequest: async ({ url, method, ...options }) => {
-    // TODO: add caching
-    return instance.axios({
+  apiRequest: async ({ url, method, ...options }) =>
+    instance({
       url,
       method: method || 'GET',
       ...options,
-    });
-  },
+    }),
 };
 
 export const isVisible = (limitedApps, app, visibility) => {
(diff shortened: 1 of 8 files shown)


karelhala (commented):
I'll be strick on this one and would vote for not using dynamic navigaton from app, meaning pulling secondary nav items from API. I'd much rather keep every navigation in cloud-services-config and only restrict users from seing it instead of pulling arbitrary items from API. This way we know all the times what navigation is possible and we could debug it much easilly.

src/js/consts.js
```diff
@@ -44,14 +45,12 @@ export const visibilityFunctions = {
...
-    // TODO: add caching
-    return instance.axios({
+  apiRequest: async ({ url, method, accessor, ...options }) =>
+    instance({
       url,
       method: method || 'GET',
       ...options,
-    });
-  },
+    }).then((response) => (accessor ? get(response?.data || {}, accessor) : response)),
```
karelhala, line 53: Since we are using `interceptors` you should return `response.data`
```suggestion
    }).then((response) => (accessor ? get(response?.data || {}, accessor) : (response?.data || response)),
```

karelhala (approved):

Hyperkid123:
@karelhala you will have to review your changes to pas the checks. I can't give it :green_circle:

karelhala (approved):

PR #1031, opened 2020-10-21
Log insights.chrome.navigation usage

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-9190

I have found only one deprecated function.

@karelhala @ryelo are you aware of some other methods that should be deprecated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,9 @@ Insights Chrome comes with a Javacript API that allows applications to control n
     insights.chrome.identifyApp('advisor');
 ```
 
+### Deprecated functions
+- `insights.chrome.navigation` this is a legacy function and is no longer supported. Invoking it has no effect.
+
 #### Global events
 
 The following events can be observed:


karelhala (approved):
I don't think there are other functions.

Hyperkid123:
@karelhala is the sentry usage ok? I have never used it before. I assume account info is passed to it by default right?

src/js/chrome/entry.js
```diff
@@ -57,7 +58,11 @@ export function chromeInit(navResolver) {
 
   return {
     identifyApp: (data) => navResolver.then(() => identifyApp(data, store.getState().chrome.globalNav)),
-    navigation: () => console.error("Don't use insights.chrome.navigation, it has been deprecated!"),
+    navigation: () => {
+      log('Deprecated function call: insights.chrome.navigation');
+      Sentry.captureMessage('Deprecated function call: insights.chrome.navigation');
+      console.error("Don't use insights.chrome.navigation, it has been deprecated!");
```
ryelo, line 64: You don't have to do two of these. If you just throw a `console.error` it'll send to sentry automatically

ryelo (changes requested):

src/js/chrome/entry.js
Hyperkid123: Ah so just `console.error("Don't use insights.chrome.navigation, it has been deprecated!");` is enough? Ok, so I guess no changes are required for this part. It was like that before.

src/js/chrome/entry.js
```diff
@@ -57,7 +57,9 @@ export function chromeInit(navResolver) {
 
   return {
     identifyApp: (data) => navResolver.then(() => identifyApp(data, store.getState().chrome.globalNav)),
-    navigation: () => console.error("Don't use insights.chrome.navigation, it has been deprecated!"),
+    navigation: () => {
```
karelhala, line 60: No need to wrap in function body.

karelhala (commented):
I think we can just leave the documentation and sentry will handle the rest.

karelhala (approved):

PR #1033, opened 2020-10-21
App switcher updates

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-9714

Adds settings and user prefs (and a spacer) to the App Switcher

Screen Shot 2020-10-21 at 2 19 56 PM

### DIFF
--- a/src/js/App/Sidenav/AppSwitcher.js
+++ b/src/js/App/Sidenav/AppSwitcher.js
@@ -14,19 +14,29 @@ const AppSwitcher = ({ currentApp }) => {
 
   const dropdownItems = (activeApp) => {
     const appList = [
-      { title: 'Red Hat Insights', id: 'insights' },
-      { title: 'Red Hat OpenShift Cluster Manager', id: 'openshift' },
-      { title: 'Red Hat Ansible Automation Platform', id: 'ansible' },
-      { title: 'Cost Management', id: 'cost-management' },
-      { title: 'Migration Services', id: 'migrations' },
-      { title: 'Subscription Watch', id: 'subscriptions' },
+      { title: 'Red Hat Insights', id: 'insights', type: 'link' },
+      { title: 'Red Hat OpenShift Cluster Manager', id: 'openshift', type: 'link' },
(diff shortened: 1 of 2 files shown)


src/js/App/Sidenav/AppSwitcher.js
```diff
@@ -14,19 +15,29 @@ const AppSwitcher = ({ currentApp }) => {
...
-      { title: 'Cost Management', id: 'cost-management' },
-      { title: 'Migration Services', id: 'migrations' },
-      { title: 'Subscription Watch', id: 'subscriptions' },
+      { title: 'Red Hat Insights', id: 'insights', type: 'link' },
+      { title: 'Red Hat OpenShift Cluster Manager', id: 'openshift', type: 'link' },
+      { title: 'Red Hat Ansible Automation Platform', id: 'ansible', type: 'link' },
+      { title: 'Cost Management', id: 'cost-management', type: 'link' },
+      { title: 'Migration Services', id: 'migrations', type: 'link' },
+      { title: 'Subscription Watch', id: 'subscriptions', type: 'link' },
+      { type: 'separator' },
```
Hyperkid123, line 24: I think this will throw an error about missing `key` prop. We need some ID to help react with rendering an array of components.

karelhala: Good point! However no need to have a dummy id, we just have to change line 29 from `const renderNavItems = appList.map((app) => (` to `const renderNavItems = appList.map((app, key) => (` and line 41 from `key={app.id}` to `key={app.id || key}`

Hyperkid123: Well but you know you should not use array indexes as keys.

karelhala: Yeah, neither of these are perfect solutions.

epwinchell: @karelhala @Hyperkid123 I added a dummy  ID. Let me know if I should go the other direction.

karelhala (approved):

Hyperkid123 (approved):

PR #1039, opened 2020-10-23
Calculate only currently observed nav

karelhala:
### No need to calculate nav of other bundles

When user is in one bundle navigation is also calculated for other bundle, this is not correct as it can make the navigation loading slow and unresponsive. This PR calculates only currently observed navigation and further reduces rendering of navigation. This is really obvious if you directly hit any URL without any cache. Once you navigate between apps it's not so obvious as we are using cached navigation.

### Allow passing array of permissions

Some apps might require multiple checks before rendering navigation items so we have to check if permissions is array and not to wrap it in another array.

### DIFF
--- a/src/js/nav/globalNav.js
+++ b/src/js/nav/globalNav.js
@@ -3,9 +3,9 @@ import { safeLoad } from 'js-yaml';
 import flatMap from 'lodash/flatMap';
 import { getUrl } from '../utils';
 
-export let getNavFromConfig = async (masterConfig) => {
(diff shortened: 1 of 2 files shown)


rvsia (approved):

Hyperkid123 (dismissed):
:snail:

Hyperkid123:
@karelhala can you fix the failing tests?

Hyperkid123 (approved):
:space_invader:

PR #1055, opened 2020-11-03
Add apiRequests matchers.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-10469

There was a request for the API request to accept more than just true/false payloads. we can support that to a certain degree by adding some general matchers in order to re-use existing API endpoints (we should never support proprietary conditions like `foo.bar[15].name === 'pepík'`). This pr adds two new matchers: `isEmpty`, `isNotEmpty`.
### Changes
- removed the default `response.data` access
  - we are using Axios instance with an interceptor that is doing this
  - with the previous implementation, if the standard paginated response is returned it will make the additional `meta`/`links` keys inaccessible via accessor
- added `isEmpty` and `isNotEmpty` response matchers
  - in addition to boolean responses, we can now use arrays/objects and check if they are(not) empty
  - using [isEmpty function](https://lodash.com/docs/4.17.15#isEmpty)

### DIFF
--- a/src/js/consts.js
+++ b/src/js/consts.js
@@ -1,10 +1,25 @@
 import instance from '@redhat-cloud-services/frontend-components-utilities/files/interceptors';
 import get from 'lodash/get';
+import isEmpty from 'lodash/isEmpty';
 
 const obj = {
   noAuthParam: 'noauth',
   offlineToken: '2402500adeacc30eb5c5a8a5e2e0ec1f',
   allowedUnauthedPaths: ['/', '/logout', '/beta', '/security/insights', '/beta/security/insights'],
+};
+
+const matcherMapper = {
+  isEmpty,
+  isNotEmpty: (value) => !isEmpty(value),
+};
+/**
+ * returns true/false if value matches required criteria. If invalid or no matcher is provided it returns the original value.
+ * @param {any} value variable to be matched with matcher function
+ * @param {string} matcher id of matcher
+ */
+const matchValue = (value, matcher) => {
+  const match = matcherMapper[matcher];
+  return typeof match === 'function' ? match(value) : value;
 };
 
 export const visibilityFunctions = {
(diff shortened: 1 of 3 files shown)


src/js/consts.js
```diff
@@ -45,17 +60,18 @@ export const visibilityFunctions = {
...
-  apiRequest: async ({ url, method, accessor, ...options }) =>
-    instance({
+  apiRequest: async ({ url, method, accessor, matcher, ...options }) => {
+    return instance({
       url,
       method: method || 'GET',
       ...options,
     })
-      .then((response) => (accessor ? get(response?.data || response || {}, accessor) : response?.data || response))
+      .then((response) => matchValue(accessor ? get(response || {}, accessor) : response, matcher))
```
karelhala, line 69: I'd still keep in here the `response?.data || response`
```JS
.then((response) => matchValue(accessor ? get(response || {}, accessor) : response?.data || response, matcher))
```
This way, we don't have to add `accessor` for each generic API and if the response will be located deeply in the response consumers can use `accessor`

src/js/nav/globalNav.test.js
```diff
@@ -81,7 +110,7 @@ describe('global nav with API restricted sub items', () => {
     axiosSpy.mockReset();
   });
 
-  test('should display sub item with positive API response', async () => {
+  test.only('should display sub item with positive API response', async () => {
```
karelhala, line 113: You should test all tests and not this test only.

src/js/consts.js
Hyperkid123: @karelhala we actually do. Because of some nested endpoints. What if you want to access nested attribute from payload like this:
```jsx
{
  bar: {
    isEntitled: true
  }
}
```

Catalog and approval for instance have user preferences and metadata on certain API responses. I can imagine using those for this exact purpose. Here an example of how it may look.
```jsx
{
  data: [...]
  meta: {
    user_preferences: {
      canRead: false
    }
  }
}
```

src/js/nav/globalNav.test.js
Hyperkid123: yeah left it there accidentally.

karelhala (approved):

PR #1057, opened 2020-11-03
add ability to get current environment

ryelo:


Hyperkid123 (dismissed):

src/js/utils.js
```diff
@@ -127,3 +128,11 @@ export function getUrl(type) {
 
   return type === 'bundle' ? sections[1] : sections[2];
 }
+
+export function getEnv() {
+  let currentEnv = 'qa'; // Fallback in case env is not listed
+
+  Object.entries(DEFAULT_ROUTES).map(([env, values]) => values.url.map((url) => url === location.hostname && (currentEnv = env)));
```
karelhala, line 135: This seems a bit strange, how about using `find` instead 2 nested maps?
```JS
return Object.entries(DEFAULT_ROUTES).find(([, { url }]) => url.includes(location.hostname))?.[0] || 'qa';
```
Those 2 nested maps would go trough each environment and each url regardless if already found (thus eliminating couple of roudtrips) and this could rewrite already found environment.

ryelo: Works for me 💯

karelhala (approved):

PR #1067, opened 2020-11-16
Chroming 2.0

Hyperkid123:
Updates scalprum version to include app initialization check before mounting it.


config/webpack.plugins.js
```diff
@@ -1,24 +1,27 @@
...
-        flattening: true,
-        placeholders: true,
-        paths: true
-    })
+  new ModuleFederationPlugin({
+    name: 'chrome',
+    library: { type: 'var', name: 'chrome' },
+    filename: 'remoteEntry.js',
+    exposes: {
+      './foo': path.resolve(__dirname, '../src/js/chrome.js'),
```
karelhala, line 14: Shouldn't this be exposed on `./chrome` path?

Hyperkid123: This does not expose and does not consume anything at this point. It's there just to make sure the build passes. It will be replaced by the config from the hackaton.

PR #1068, opened 2020-11-16
Update chrome 2 layout to match current one.

Hyperkid123:
This should make it more "demoable" for now.

chrome-lookalike

### DIFF
--- a/src/js/chrome/entry.js
+++ b/src/js/chrome/entry.js
@@ -64,7 +64,7 @@ export function bootstrap(libjwt, initFunc, getUser) {
   };
 }
 
-const App = () => {
+const Chrome = () => {
   const [config, setConfig] = useState();
   useEffect(() => {
     sourceOfTruth('testPrefix')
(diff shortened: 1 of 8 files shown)


src/js/chrome/entry.js
```diff
@@ -106,16 +106,16 @@ const App = () => {
...
   }, []);
 
-  return config ? <RootApp config={config} /> : <div>loading...</div>;
+  return config ? <RootApp config={config} /> : null;
 };
 
 export function rootApp() {
   const { store } = spinUpStore();
-  const pageRoot = document.querySelector('.pf-c-page__drawer');
+  const pageRoot = document.getElementById('chrome-root');
```
karelhala, line 114: I wouldn't mount to a new id, but rather reuse the already used if `page` and keep the `body.pug` as is, so we keep the loading there while the JS is loaded and parsed.

src/sass/chrome.scss
```diff
@@ -189,3 +189,7 @@ section.ins-c-app-switcher--loading {
         text-align: inherit;
     }
 }
+
+.ins-chrome-content {
+    min-height: 100vh;
```
karelhala, line 194: I don't really like `vh` and `vw` units as they can sometime leads to unwanted scrollbars. Instead I'd set `height: 100%` on the `#page`

src/js/chrome/entry.js
Hyperkid123: That is not going to work. The header and the nav along with any other "part" of chrome will have to be part of the same VirtualDOM tree. We need this structure inside the actual React app.

We can add the loading state before the app is mounted later on.

Hyperkid123: We will also no longer have any need for pugs and esi tags because the apps won't have their own HTML templates. But we will need it them until every app is migrated.

src/sass/chrome.scss
Hyperkid123: I would not take this PR to reflect the finished thing. I am not even sure we will use the pf Page component because it does not work very well. I've mainly created this to make the demo look more like chrome instead of having just a blank page with 3 links.

PR #1070, opened 2020-11-18
Allow tags bookmarking trough hash value of location

karelhala:
### Update hash of location

In order to allow users to bookmark global filter values with tags and SIDs we have to update storing and retrieving value for it. This PR adds new function to build filtered by from array of strings and retrieves this value from URL hash. When no tags or SIDs are applied they are still present in the URL so we have backwards compatibility and allow sending URLs with empty SIDs/tags.

* only workloads hash = `#workloads=All+workloads&SIDs=&tags=`
* workloads and SIDs = `#workloads=All+workloads&SIDs=ABC&tags=`
* workloads and tags = `#workloads=All+workloads&SIDs=&tags=null%2Ftest-tag-total-04c40dec-58f0-4d54-a4a4-d391bc64da8e-0`
* workloads + tags + SIDs = `#workloads=All+workloads&SIDs=ABC&tags=null%2Ftest-tag-total-04c40dec-58f0-4d54-a4a4-d391bc64da8e-0`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3450,9 +3450,9 @@
       }
     },
     "@redhat-cloud-services/frontend-components": {
-      "version": "2.4.22",
-      "resolved": "https://registry.npmjs.org/@redhat-cloud-services/frontend-components/-/frontend-components-2.4.22.tgz",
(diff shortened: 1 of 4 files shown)


Hyperkid123 (commented):
The code looks good. Can you fix up the CI?

Hyperkid123 (approved):

PR #1072, opened 2020-11-18
add ability to get hash info for current app when suport case is opened

ryelo:


src/js/createCase.js
```diff
@@ -29,8 +29,22 @@ function registerProduct() {
...
 }
 
-export function createSupportCase(userInfo, fields) {
+async function getProductHash() {
+  const { store } = spinUpStore();
+
+  const currentApp = store.getState().chrome.appId;
+  const path = `${window.location.origin}${window.insights.chrome.isBeta() ? '/beta/' : '/' }apps/${currentApp}/app.info.json`;
+
+  const appData = await fetch(path)
```
karelhala, line 38: I don't realy like mixing `await` with `then`s, using just awaits should be enough.
```JS
const appData = await (await fetch(path)).json();
return `Current app: ${currentApp}, Current app hash: ${appData.src_hash}`);
```

karelhala (approved):

Hyperkid123 (approved):

PR #1075, opened 2020-11-19
send the hash to the env instead

ryelo:
Support webpage removes the "description" information on page load, so i'll just move that info over to the env (which doesn't get overwritten) instead

### DIFF
--- a/src/js/createCase.js
+++ b/src/js/createCase.js
@@ -59,7 +59,7 @@ export async function createSupportCase(userInfo, fields) {
       },
       sessionDetails: {
         createdBy: `${userInfo.user.username}`,
-        environment: `${window.insights.chrome.isBeta() ? 'Production Beta' : 'Production'}, ${currentHash}`,
+        environment: `Production ${window.insights.chrome.isBeta() ? 'Beta' : ''}, ${currentHash}`,
         ...(currentProduct && { product: currentProduct }),
         ...fields?.caseFields,
       },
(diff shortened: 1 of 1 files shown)


src/js/createCase.js
```diff
@@ -59,8 +59,7 @@ export async function createSupportCase(userInfo, fields) {
       },
       sessionDetails: {
         createdBy: `${userInfo.user.username}`,
-        environment: `${window.insights.chrome.isBeta() ? 'Production Beta' : 'Production'}`,
-        description: `${currentHash}`,
+        environment: `${window.insights.chrome.isBeta() ? 'Production Beta' : 'Production'}, ${currentHash}`,
```
karelhala, line 62: No need to repeat `Production` twice
```suggestion
        environment: `Production${window.insights.chrome.isBeta() ? ' Beta' : ''}, ${currentHash}`,
```

src/js/createCase.js
```diff
@@ -59,8 +59,7 @@ export async function createSupportCase(userInfo, fields) {
       },
       sessionDetails: {
         createdBy: `${userInfo.user.username}`,
-        environment: `${window.insights.chrome.isBeta() ? 'Production Beta' : 'Production'}`,
-        description: `${currentHash}`,
+        environment: `Production ${window.insights.chrome.isBeta() ? ' Beta' : ''}, ${currentHash}`,
```
karelhala, line 62: this will add one more space
```suggestion
        environment: `Production${window.insights.chrome.isBeta() ? ' Beta' : ''}, ${currentHash}`,
```

karelhala (approved):

PR #1084, opened 2020-11-26
Enable old apps within chrome 2.0

Hyperkid123:
Example of remote module config: https://github.com/RedHatInsights/cloud-services-config/pull/317
I fully expect it to change

### How does this work?
- thanks to webpack module sharing, chrome render is now always slower than legacy app render, therefore the app root element does not exist when the app tries to render. We need to dot his
   1. provide hidden HTML element statically to HTML template to serve as the app root
   2. once the chrome is rendered get the element
   3. remove the hidden property and inject it into the chrome scaffolding

### Chrome bootstrap split
In order to successfully load chrome, we need to
  - make sure that any there are no references to any of the shared modules before the async `import('./bootstrap')`
  - move the `createFilter` helper function to chrome codebase because the library helper file is referencing react
  - move the `loadInventory` and `loadRemediations` to the **bootstrap** file because they are referencing shared modules
  - stop sharing redux as it's crucial to have chrome store ASAP (we might change it in the future)

### Result
Chrome can now run both legacy apps and chrome 2.0 ready apps :tada:

### DIFF
--- a/src/js/App/RootApp.js
+++ b/src/js/App/RootApp.js
@@ -1,9 +1,10 @@
-import React, { Fragment } from 'react';
+import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
+import { connect, useSelector } from 'react-redux';
 import GlobalFilter from './GlobalFilter/GlobalFilter';
 import { useScalprum, ScalprumComponent } from '@scalprum/react-core';
 import { Bullseye, Page, PageHeader, PageSidebar, Spinner } from '@patternfly/react-core';
+import { BrowserRouter } from 'react-router-dom';
 import SideNav from './Sidenav/SideNav';
 import Header from './Header/Header';
 import ErrorBoundary from './ErrorBoundary';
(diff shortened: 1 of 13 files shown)


src/js/App/RootApp.js
```diff
@@ -12,8 +13,30 @@ const RootApp = ({ activeApp, activeLocation, appId, config, pageAction, pageObj
...
     (!globalFilterHidden && activeLocation === 'insights') || Boolean(localStorage.getItem('chrome:experimental:global-filter'));
   const scalprum = useScalprum(config);
+  const remoteModule = useSelector(({ chrome }) => chrome?.activeSection?.module);
+  const insightsContentRef = useRef(null);
+  const legacyMainContainerRef = useRef(null);
+  useEffect(() => {
+    const contentElement = document.getElementById('root');
+    if (contentElement) {
+      legacyMainContainerRef.current = contentElement;
+    }
```
karelhala, line 23: Is this really needed? We are not doing anything with the `legacyMainContainerRef` later on, we might want to change line 25 `insightsContentRef.current.appendChild(legacyMainContainerRef.current);` if we want to keep it.

src/js/App/RootApp.js
```diff
@@ -12,8 +13,30 @@ const RootApp = ({ activeApp, activeLocation, appId, config, pageAction, pageObj
...
+    if (!remoteModule) {
+      insightsContentRef.current.appendChild(contentElement);
```
karelhala, line 25: I'd log the usage of this in Sentry, we want to ping apps that are not using new chrome-2.0 and tell them to use it. Not needed for now, but nice to have for future.

src/js/App/RootApp.js
Hyperkid123: I think you are right.

src/js/App/RootApp.js
Hyperkid123: We can add it later. I think we might hit sentry limits very soon if we log this.

karelhala (approved):

PR #1087, opened 2020-11-27
Rewrite tests

karelhala:
### Change the test library

Enzyme does not support React 17, let's switch to react-testing library so we can pass our tests.


Hyperkid123:
@karelhala linter died

Hyperkid123 (approved):

PR #1090, opened 2020-12-02
change var name for easier debugging

ryelo:
This is actually the user's ID, changing it so it's not confusing with the actual account id


karelhala (approved):

ryelo:
closing in favor of #1091

PR #1096, opened 2020-12-07
use new fields for case creation

ryelo:
Default "product" should be "other" 

activeGroup should be used instead of activeId to get the app.info.json

### DIFF
--- a/src/js/createCase.js
+++ b/src/js/createCase.js
@@ -45,8 +45,6 @@ export async function createSupportCase(userInfo, fields) {
 
   log('Creating a support case');
 
-  console.debug(currentProduct);
-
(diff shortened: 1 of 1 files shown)


src/js/createCase.js
```diff
@@ -32,19 +32,21 @@ function registerProduct() {
...
 
 export async function createSupportCase(userInfo, fields) {
-  const currentProduct = registerProduct();
+  const currentProduct = registerProduct() || 'Other';
   const currentHash = await getProductHash();
 
   log('Creating a support case');
 
+  console.debug(currentProduct);
+
```
karelhala, line 49: No need to console.
```suggestion
```

karelhala (approved):

PR #1098, opened 2020-12-09
Add data-ouia-safe on chrome init.

Hyperkid123:
Chrome 2 compatible applications should not mutate any chrome "owned" elements. We are removing the code that does this while migrating apps.

The ouia safe attribute will be set for now inside the chrome init function, which every app has to call. In chrome 2.0 we can do this via react component lifecycle.


john-dupuy (approved):

Hyperkid123:
Ah travis :cry: Here we go again.

karelhala (approved):

PR #1102, opened 2020-12-17
Remove redhat from app switcher

karelhala:


Hyperkid123 (approved):

src/js/createCase.js
```diff
@@ -7,7 +7,7 @@ import { spinUpStore } from './redux-config';
 // Lit of products that are bundles
 const BUNDLE_PRODUCTS = [
   { id: 'openshift', name: 'Red Hat OpenShift Cluster Manager' },
-  { id: 'insights', name: 'Red Hat Insights' },
+  { id: 'insights', name: 'Insights' },
```
ryelo, line 10: None of this file needs to change. These fields are mapped to PCM's fields

ryelo (changes requested):

src/js/createCase.js
karelhala: Good point! I'll add it back.

Hyperkid123 (approved):

ryelo (approved):

PR #1117, opened 2021-01-11
Update remediations dependency.

Hyperkid123:
This update is required for the compliance migrations.

Fixes various issues with undefined remediations components (cc @fhlavac this should also fix the # 130 issue you are having with remediations.)


karelhala (approved):

fhlavac:
Thanks! @Hyperkid123

PR #1118, opened 2021-01-11
Add auth module

ryelo:


src/js/jwt/modules/useChromeAuth.js
```diff
@@ -0,0 +1,71 @@
...
+
+export function allowUnauthed() {
+  if (getAllowedUnauthedPaths().includes(getWindow().location.pathname)) {
+    return true;
+  }
+
+  return false;
+}
+
+const useChromeAuth = () => {
```
Hyperkid123, line 41: In order to truly re-use only one instance, I think we need to use React context. That would wrap the whole chrome and as a value provide the return object from this function. This implementation will run `const promise = jwt.init(options).then(bouncer);` and every other function call on each hook call. Which is something we probabylu don't want.

The `useChromeAuth` would then be just a wrapped useContext, consuming the auth context value.

Hyperkid123 (approved):

PR #1119, opened 2021-01-12
Update global filter with new UX requests

karelhala:
### New updates to global filter

UX requested new things added to the global filter
* Remove "All Workloads" option
* Change SAP from radio button to checkbox
* Change textbox words from "Search Tags" to "Filter Results"


src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -9,7 +9,7 @@ import { Chip, ChipGroup } from '@patternfly/react-core/dist/js/components/ChipG
 import { Button } from '@patternfly/react-core/dist/js/components/Button';
 import { Tooltip } from '@patternfly/react-core/dist/js/components/Tooltip';
 import TagsModal from './TagsModal';
-import { workloads, updateSelected, storeFilter, generateFilter, selectWorkloads } from './constants';
+import { workloads, updateSelected, storeFilter, generateFilter } from './constants';
```
Hyperkid123, line 12: I don't see the `selectWorkloads` to be used anywhere else. Can we delete the function declaration as well?

src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -99,22 +99,12 @@ const GlobalFilter = () => {
...
-      } else {
-        dispatch(globalFilterChange(selectedTags));
-      }
+      dispatch(globalFilterChange(selectedTags));
     }
   }, [selectedTags, isAllowed()]);
 
   useEffect(() => {
-    const sapTag = workloads?.[0]?.tags?.[1];
+    const sapTag = workloads?.[0]?.tags?.[0];
```
Hyperkid123, line 107: I am just curious about this one. The SAP tag is always the first tag? And if so can this be refactored so it's a bit nicer on the eye?

karelhala: On the initial mocks, there was SAP and some other workload, hence the array. It's a static array passed from constants. I'm thinking about exporting just the SAP tag itself and bypass the workloads.

src/js/App/GlobalFilter/GlobalFilter.js
karelhala: Already removed https://github.com/RedHatInsights/insights-chrome/pull/1119/files#diff-108371adf9416c27a728f9b30a115850331944eaf03f26ab052c7ade2447bd93L26

src/js/App/GlobalFilter/GlobalFilter.js
Hyperkid123: Yeah, it would definitely make more sense. We can do it in a different PR though.

Hyperkid123 (approved):

PR #1121, opened 2021-01-13
Do not load nav config for unauthed user.

Hyperkid123:
@ryelo I got a question. Is this going to break openshift nav? Should we make an exception for that namepsace? I recall that even unauthed users can access the OCM.

jira: https://issues.redhat.com/browse/RHCLOUD-11627

### DIFF
--- a/src/js/chrome/create-chrome.js
+++ b/src/js/chrome/create-chrome.js
@@ -26,9 +26,21 @@ const createChromeInstance = (jwt, insights) => {
    * Load navigation after login
    */
   const navResolver = jwtResolver.then(async () => {
+    const user = await libjwt.jwt.getUserInfo();
+    if (!user) {
+      /**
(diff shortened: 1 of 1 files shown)


karelhala (approved):

john-dupuy (approved):

karelhala:
@Hyperkid123 I don't think unauth users should have access to the OCM bundle. However they can access their bundle if they have no entitlements, which is still working.

ryelo:
Yeah, so @karelhala is partially correct. They have to be authed, but OCM doesn't care about an account number (which in the end is no entitlements). Whenever you get an entitlement, you get an account number. So, the OCM bundle, along with the trust page, should allow users with a 0, -1, or NaN account numbers.

Hyperkid123:
seem like we have some new years date present

PR #1122, opened 2021-01-14
track Pendo ad blocking

PanSpagetka:
Try to fetch pendo,if not successful log that pendo is blocked. By defaul `fetch`has huge timeout, so I have made it 5 sec instead.

https://issues.redhat.com/browse/RHCLOUD-9700

### DIFF
--- a/src/js/analytics.js
+++ b/src/js/analytics.js
@@ -94,25 +94,9 @@ function getPendoConf(data) {
   };
 }
 
-const tryToFetchPendo = async () => {
-  function timeout(ms, promise) {
-    return new Promise(function (resolve, reject) {
-      setTimeout(function () {
-        reject(new Error('timeout'));
-      }, ms);
-      promise.then(resolve, reject);
-    });
-  }
-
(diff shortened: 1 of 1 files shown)


Hyperkid123 (commented):
Instead of doing multiple requests for the same resource, we could also try using the `onerror` handler.
```jsx
s.onerror = function(error){console.log(error, 'error')}
```

In this case you could add that callback to here
```jsx
(v[w]);y=e.createElement(n);y.async=!0;y.src=`https://content.analytics.cloud.redhat.com/agent/static/${API_KEY}/pendo.js`
```
Targeting the "y" element.
```
y.onerror = function(error){log('Pendo blocked')}
```

PanSpagetka:
@Hyperkid123 Thanks! It seems to work and it is much nicer!

ryelo (changes requested):
Instead of doing a regular log, can you either throw a console.error or log it to Sentry that way we can gather the metrics in sentry?

ryelo (approved):

PR #1123, opened 2021-01-14
fix user menu again

ryelo:
First time something broke and didn't work. This one scopes it up one level


Hyperkid123 (commented):
Tested in UI seems to be working. One thing though. We might have to add `!important` to the display property. I've checked the current master and the layout is sometimes flex and sometimes inline-block.

It's display property overridden by the default plain dropdown styling so it's just a matter of when the CSS rule is injected into the DOM.

ryelo:
@Hyperkid123 This style should be more specific than the other selectors that could potentially try to override styles (or at least when I was looking) this should allow this style to override anything else because of the usage of the `#`. It shouldn't matter when the rule is injected

Hyperkid123 (approved):

PR #1136, opened 2021-01-22
remove all the temp nav styling

ryelo:
When we had the platform wide upgrade to the last major release of Patternfly, we also added some temp nav styling in case apps didn't upgrade and we could overwrite their styling. This removes all of that old styling and should use the actual css.

I tried to check this between different apps and different scenarios, but please double check this 😄


karelhala (approved):

karelhala:
Let's merge this and cross test this on live envs accross multiple apps.

PR #1146, opened 2021-01-28
Fix App filter

PanSpagetka:
Individual apps are not filtered correctly, this should fix that


epwinchell:
cc @jjaquish

epwinchell (approved):
Tested. Looks good.

PR #1147, opened 2021-02-01
Add link to docs

pmuir:
For now this just points to the main docs landing page until CCS get the docs site up


pmuir:
@kruai @karelhala I hope this is the right way to do this :-)

@DuncanDoyle fyi

karelhala (approved):
This the way how to add doc links! It will be shown only for apps under `application-services` main.yml partial.

PR #1150, opened 2021-02-02
Add navigation to ci-beta landing page.

Hyperkid123:
The landing page nav was moved to chrome to re-use the same data as app filter does. We can also use show/hide logic for the menu which is nice.

jira: https://issues.redhat.com/browse/RHCLOUD-11968

### Changes
- create hook from app filter data loading
  - hook can now be re-used for the landing nav
  - data is cached so we should only make one request
- move `isBeta` declaration to helper functions
- create `isFilterEnabled` constant feature flag until all these changes hits prod
- added nav to landing page
- creates new pug template for the landing page with nav

screenshot-ci foo redhat com_1337-2021 02 02-11_34_29

### DIFF
--- a/src/js/chrome/entry.js
+++ b/src/js/chrome/entry.js
@@ -13,7 +13,7 @@ import debugFunctions from '../debugFunctions';
 import { visibilityFunctions } from '../consts';
 import Cookies from 'js-cookie';
 import logger from '../jwt/logger';
-import { getUrl, getEnv } from '../utils';
+import { getUrl, getEnv, isBeta } from '../utils';
 import { createSupportCase } from '../createCase';
 import get from 'lodash/get';
 import { flatTags } from '../App/GlobalFilter/constants';
@@ -109,7 +109,7 @@ export function bootstrap(libjwt, initFunc, getUser) {
         login: () => libjwt.jwt.login(),
       },
       isProd: window.location.host === 'cloud.redhat.com',
-      isBeta: () => (window.location.pathname.split('/')[1] === 'beta' ? true : false),
+      isBeta,
       isPenTest: () => (Cookies.get('x-rh-insights-pentest') ? true : false),
       getBundle: () => getUrl('bundle'),
       getApp: () => getUrl('app'),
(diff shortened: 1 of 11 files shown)


src/js/chrome/entry.js
```diff
@@ -109,7 +109,7 @@ export function bootstrap(libjwt, initFunc, getUser) {
         login: () => libjwt.jwt.login(),
       },
       isProd: window.location.host === 'cloud.redhat.com',
-      isBeta: () => (window.location.pathname.split('/')[1] === 'beta' ? true : false),
+      isBeta: isBeta,
```
karelhala, line 112: No need to assign to same name.
```suggestion
      isBeta,
```

src/js/utils/useGlobalNav.js
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useState } from 'react';
+import { safeLoad } from 'js-yaml';
+import { getNavFromConfig } from '../nav/globalNav';
+import sourceOfTruth from '../nav/sourceOfTruth';
+
+const appIds = ['insights', 'openshift', 'cost-management', 'migrations', 'subscriptions', 'ansible', 'settings'];
+
+const useGlobalNav = () => {
```
karelhala, line 8: Can you also pass down the information that the nav is being loaded?

karelhala (commented):
Looking hella good! Just 2 minor things and we are golden.

ryelo:
Note here, we may want to expose another HTML file, similar to `landing.pug` that does not have the side nav. You can see that the trust page is using that pug file because they wanted a page without a nav (https://github.com/RedHatInsights/trust-frontend/blob/c27598617755dcd3c58f8020bd4aba1cac78dbec/src/index.html#L9). 

Or just expose something to hide the nav entirely.

Hyperkid123:
@ryelo Ah I was not aware. I think long term we should make nav-less template and use that in the trust app instead. It would make more sense to me than sharing the same partial with the landing page.

But since this is a still "temporary" change I think making special for the landing page is good for now. We still don't know if it will hit the prod or not.

karelhala (approved):

PR #1181, opened 2021-02-10
Enable landing page navigation

Hyperkid123:
Enables landing page nav in chrome 2.0

nav is part of the chrome react app.

Merge with: https://github.com/RedHatInsights/landing-page-frontend/pull/189

@epwinchell styling looks a bit strange to me. Could look at it once its merged and fix it? There are probably some global styles somewhere.

### DIFF
--- a/src/js/App/RootApp.js
+++ b/src/js/App/RootApp.js
@@ -8,7 +8,8 @@ import { BrowserRouter } from 'react-router-dom';
 import SideNav from './Sidenav/SideNav';
 import Header from './Header/Header';
 import ErrorBoundary from './ErrorBoundary';
-import { isBeta } from '../utils';
+import { getEnv, isBeta } from '../utils';
+import LandingNav from './Sidenav/LandingNav';
 
 const LoadingComponent = () => (
   <Bullseye className="pf-u-p-xl">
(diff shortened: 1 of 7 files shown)


src/js/App/RootApp.js
```diff
@@ -62,7 +65,10 @@ const RootApp = ({ activeApp, activeLocation, appId, config, pageAction, pageObj
         {...(pageAction && { 'data-ouia-page-type': pageAction })}
         {...(pageObjectId && { 'data-ouia-page-object-id': pageObjectId })}
       >
-        <Page header={<PageHeader headerTools={<Header />} />} sidebar={<PageSidebar id="ins-c-sidebar" nav={<SideNav />} isNavOpen />}>
+        <Page
+          header={<PageHeader headerTools={<Header />} />}
+          sidebar={<PageSidebar hidden={hideNav} id="ins-c-sidebar" nav={useLandingNav ? <LandingNav /> : <SideNav />} isNavOpen />}
```
karelhala, line 70: Might be better not to use the pageSidebar at all when `hideNav`

epwinchell (commented):
@Hyperkid123 The only thing I noticed is that the menu won't collapse on its own until the hamburger is clicked at least once. Then, it collapses and expands fine on its own

Hyperkid123:
> @Hyperkid123 The only thing I noticed is that the menu won't collapse on its own until the hamburger is clicked at least once. Then, it collapses and expands fine on its own

Ok. Will fix that in a follow-up PR.

karelhala (approved):

PR #1183, opened 2021-02-11
Update layout to use react PF page layout.

Hyperkid123:
### Changes
- removed custom nav toggle control
- use PF page component for the layout
- Fix landing page nav flash before appId is available in chrome
- Fix missing CSS scope for chrome 2 apps (required for safe CSS loading)

Fixes issues with navigation toggle

nav-collapse


src/js/App/Header/Header.js
```diff
@@ -1,22 +1,25 @@
...
+      {user && isFilterEnabled && <AppFilter />}
     </Fragment>
-  ) : (
-    <UnAuthtedHeader />
   );
 };
 
-export default Header;
+export const HeaderTools = () => {
+  const user = useSelector(({ chrome: { user } }) => user);
```
karelhala, line 20: Chrome could be undefined if the init was not fully loaded.
```suggestion
  const user = useSelector(({ chrome }) => chrome?.user);
```

src/js/App/Header/Login.js
```diff
@@ -1,13 +1,14 @@
 import React from 'react';
 import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
 import { login } from '../../jwt/jwt';
+import { PageHeaderTools } from '@patternfly/react-core';
```
karelhala, line 4: Shouldn't we load react core directly?
```suggestion
import { PageHeaderTools } from '@patternfly/react-core/dist/js/components/PageHeaderTools/PageHeaderTools';
```

Hyperkid123: Right. No nits not required anymore. Only when we build libraries. Webpack will tree shake correctly here.

karelhala (approved):

PR #1194, opened 2021-02-12
Select active module based on group

karelhala:


Hyperkid123 (approved):
Works. Tested with dashboard UI which was crashing before.

PR #1201, opened 2021-02-16
Add menu section component

PanSpagetka:
Test data looks like this:
```
{
  id: 'test',
  title: 'test',
  items: [
    { id: 'app1', title: 'App no. 1' },
    {
      id: 'app2',
      title: 'App no. 2',
      subItems: [
        { id: 'subapp1', title: 'SubApp no. 1' },
        { id: 'subapp2', title: 'SubApp no. 2' },
      ],
    },
  ],
}
```

Screenshot from 2021-02-16 13-16-21

This is part of https://issues.redhat.com/browse/RHCLOUD-12315


src/js/App/Sidenav/SectionNav.js
```diff
@@ -0,0 +1,44 @@
...
+              {...props}
+              {...item}
+              onClick={(event, subItem) => (item.subItems ? onClick(event, subItem, item) : onClick(event, item))}
+            />
+          ))}
+        </NavGroup>
+      </React.Fragment>
+    );
+  }
+  const item = { id: id, title: title, props };
```
karelhala, line 24: No need to assign to same name and props should be spread.
```suggestion
  const item = { id, title, ...props };
```

karelhala (commented):
Just a two changes and we are golden! I am counting that there will be a followup PR that adds logic to assign nav items to sections, right?

PanSpagetka:
@karelhala  yeah, I am working on CSC and nav loading changes right now. Hopefully today/tomorrow there will be PR. Without additional changes this PR should do nothing from user point of view.

Hyperkid123:
@PanSpagetka  @karelhala do we know the format for the main.yml config? Or is it possible to add these groups without any changes?

PanSpagetka:
@Hyperkid123 @karelhala Initially I wanted to introduce some hierarchical structure for sub_apps like:
```
 sub_apps:
      - id: my-user-access
        default: true
      - id: rbac
      - id: section1
          items
          - id: sources
          - id: applications
```

But that seems to require quite few changes in the code and might possibly broke other staff. I am not sure how much other code relies on current structure of navigation data.

However I got an idea that we could just add section to each app:
```
sources:
  title: Discovered Inventory
  api:
    versions:
      - v1
  channel: '#insights_topology_svc'
  deployment_repo: https://github.com/RedHatInsights/sources-ui-deploy
  frontend:
    title: Sources
    section: test
    paths:
      - /settings/sources
  source_repo: https://github.com/RedHatInsights/sources-ui
```

This is fairly simple to implement and it wont mess up any existing code. But on the other hand, it is not that nice...
Screenshot from 2021-02-17 14-32-07

Hyperkid123:
@PanSpagetka @karelhala  I am personally not afraid of making changes to the main.yml config. I think creating a rigid and well-specified structure of the config is better for the long term. I believe our team is the only team that actually consumes it.

karelhala:
@Hyperkid123 the problem with changing the config to represent completely the nav structure is that we'll have to change the autodeployer as well. It currently consumes this config and creates new record for each app. This was chosen because we can reference same app from within multiple places. For instance advisor can be in OCM and in insights, same for subs watch etc. Furthermore we can define entrypoints to each app (the paths section of each object). We could potentionally think of a better structure within each object, I just wouldn't break the way how things are referenced.

TL;DR @PanSpagetka I think it's wise to go with `section` for the time being and we could think of a different structure in a future.

KendraMar:
Hi, that test section looks correct (I'm assuming no one is asking you to add anything to the Settings section at this point).

@Rohoover Just triple checking that **App no. 2** should def expand to show  2 subapps in the left nav.

karelhala (approved):

PR #1210, opened 2021-02-18
Load sections from main.yaml

PanSpagetka:
Should be merged after #1201

This PR adds the ability to parse sections from `CSC. It groups together apps with same section set. But without changes to CSC this should not result in any visual changes.

Example:
```
sources:
  title: Discovered Inventory
  api:
    versions:
      - v1
  channel: '#insights_topology_svc'
  deployment_repo: https://github.com/RedHatInsights/sources-ui-deploy
  frontend:
    title: Sources
    section: test
    paths:
      - /settings/sources
  source_repo: https://github.com/RedHatInsights/sources-ui
```

### DIFF
--- a/src/js/App/Sidenav/Navigation.js
+++ b/src/js/App/Sidenav/Navigation.js
@@ -190,9 +190,9 @@ export const Navigation = () => {
   };
 
   const settingsWithSections = settings.reduce((acc, item) => {
-    const section = acc.find((i) => i.section === item.section) || { items: [], section: item.section };
+    const section = acc.find(({ section }) => section === item.section) || { items: [], section: item.section };
     return [
-      ...acc.filter((i) => i.section === undefined || i.section !== item.section),
+      ...acc.filter(({ section }) => section === undefined || section !== item.section),
       item.section ? { ...section, items: [...section.items, item] } : item,
     ];
   }, []);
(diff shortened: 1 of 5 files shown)


src/js/App/Sidenav/SectionNav.js
```diff
@@ -4,10 +4,12 @@ import { NavGroup } from '@patternfly/react-core/dist/js/components/Nav/NavGroup
 import ExpandableNav from './ExpandableNav';
 import './SectionNav.scss';
 
-const SectionNav = ({ items, title, id, onClick, ...props }) => {
+const sectionTitleMapper = (id) => id.toUpperCase();
```
karelhala, line 7: This will make the entire id upperCase, I'd imagine we have a mapper with something like
```JS
const sectionTitleMapper = (id) => ({
  insights: 'Insights'
})?.[id] || '';
```

src/js/App/Sidenav/Navigation.js
```diff
@@ -189,10 +189,18 @@ export const Navigation = () => {
     }
   };
 
+  const settingsWithSections = settings.reduce((acc, item) => {
+    const section = acc.find((i) => i.section === item.section) || { items: [], section: item.section };
```
karelhala, line 193: I would cimplify this
```suggestion
    const section = acc.find(({ section }) => section === item.section) || { items: [], section: item.section };
```

karelhala (approved):

PR #1212, opened 2021-02-18
Add back nav_toggle event

Hyperkid123:
Some apps have to use this event to **resize their PF charts**. We will re-introduce the event only until the PF components are fixed. This is something that components should do internally and not rely on some external events.

I would suggest using ResizeObserver (https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) but it's a fairly new API.

### DIFF
--- a/src/js/App/RootApp.js
+++ b/src/js/App/RootApp.js
@@ -1,6 +1,6 @@
 import React, { memo, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
-import { connect, shallowEqual, useSelector } from 'react-redux';
+import { connect, shallowEqual, useDispatch, useSelector } from 'react-redux';
 import GlobalFilter from './GlobalFilter/GlobalFilter';
 import { useScalprum, ScalprumComponent } from '@scalprum/react-core';
 import { Bullseye, Page, PageHeader, PageSidebar, Spinner } from '@patternfly/react-core';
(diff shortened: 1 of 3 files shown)


src/js/App/RootApp.js
```diff
@@ -23,32 +24,46 @@ const isModule = (key, chrome) =>
...
-          </main>
-        )}
-        <main className="pf-c-page__main" id="no-access"></main>
-      </div>
-    </Page>
-  ),
+  ({ useLandingNav, hideNav, insightsContentRef, isGlobalFilterEnabled, initialized, remoteModule, appId }) => {
+    const dispatch = useDispatch();
+    useEffect(() => {
+      const navToggleElement = document.querySelector('button#nav-toggle');
```
karelhala, line 30: Can we use `navRef` in here which would be passed to `PageHeader` to attach this `onClick` function?

karelhala (commented):
Would be awesome to use `ref` but if it's not possible I am fine with this temporary fix. We'll definetely have to talk to PF folks about charts resing so we don't have to write these hacky things.

karelhala (approved):

src/js/App/RootApp.js
Hyperkid123: No we, unfortunately, can't. No refs are propagated. I've tried.

PR #1224, opened 2021-02-23
Migrate usage of inventory to use v3

karelhala:
* [x] merge after https://github.com/RedHatInsights/frontend-components/pull/989


Hyperkid123 (commented):
@karelhala can you bump the dependency version?

config/webpack.plugins.js
```diff
@@ -28,6 +28,7 @@ const plugins = [
     shared: [
       { react: { singleton: true, requiredVersion: deps.react } },
       { 'react-dom': { singleton: true, requiredVersion: deps['react-dom'] } },
+      { 'react-redux': { singleton: true, requiredVersion: deps['react-redux'] } },
```
Hyperkid123, line 31: Is this required? It can actually break the dependencies at runtime. Hooks will be referencing to incorrect store.

karelhala: Yup, inventory is using react-redux functions and without this it trhew an error with missing dependency. We could remove the singleton here

Hyperkid123: Yeah we have to remove the singleton

Hyperkid123 (approved):

PR #1226, opened 2021-02-23
Fix ERR_UNKNOWN_URL_SCHEME source map warnings

Hyperkid123:
Removes the swarm of source maps warnings from console when running chrome locally.

### DIFF
--- a/config/webpack.plugins.js
+++ b/config/webpack.plugins.js
@@ -11,7 +11,6 @@ const plugins = [
   new WriteFileWebpackPlugin(),
   new webpack.SourceMapDevToolPlugin({
     test: /\.js/i,
-    exclude: /node_modules/i,
     filename: `sourcemaps/[name].js.map`,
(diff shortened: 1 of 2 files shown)


config/webpack.plugins.js
```diff
@@ -11,7 +11,6 @@ const plugins = [
   new WriteFileWebpackPlugin(),
   new webpack.SourceMapDevToolPlugin({
     test: /\.js/i,
-    exclude: /node_modules/i,
```
karelhala, line 14: Do we want to include node_modules as well? Sounds reasonable, just wanted to confirm.

karelhala (approved):

config/webpack.plugins.js
Hyperkid123: Yes. That is actually what fixes the problem. The warnings are about node modules source mapping.

PR #1245, opened 2021-03-03
Fix external link spacing in Navigation

PanSpagetka:
Fix css selector for external link spacing.

https://issues.redhat.com/browse/RHCLOUD-12315
https://docs.google.com/document/d/1kgLaFGeBZqSuzXRRsmLEm8ghktP8yBvMTQ0bIZLOnms/edit#


src/js/App/Sidenav/Navigation.scss
```diff
@@ -1,5 +1,5 @@
-.ins-c-sidebar#ins-c-sidebar .pf-c-nav .pf-c-nav__list .pf-c-nav__link {
-    &.ins-c-navigation__additional-links {
+.pf-c-page__sidebar#ins-c-sidebar .pf-c-page__sidebar-body .pf-c-nav .pf-c-nav__list .pf-c-nav__item {
```
Hyperkid123, line 1: Can ve make it not to use pf selector? Maybe use some wrapping component selector or add some if it's missing.

PanSpagetka: Yep, actually i think it is safe to just use the class name. It is our class and it is used only at one place.

Hyperkid123 (approved):

PR #1254, opened 2021-03-09
Add cost to global scalprum registry

Hyperkid123:
Cost needs to share modules across different bundles so we have to make it global.

Once we start working on the new nav structure, I think we should also decouple the module definition so we can have it defined in one place. Currently it "fails" because we only evaluate apps from the current bundle. We can have a separate section for modules and register them all at once.


dlabrecq (approved):
LGTM. Thank you for adding this

karelhala:
You don't really have to do this. There's already a function to pre-register module if it's not registered https://github.com/RedHatInsights/insights-chrome/blob/master/docs/api.md#register-custom-module. So before using the cost module you should call `insights.chrome.registerModule('cost-management')` it will loadd all necessary config options and register it for later use.

PR #1260, opened 2021-03-12
Block app filter loading before first open

karelhala:
### Do not fire nav filter calls

If app filter is not opened do not fire requests to filter the full navigation. This PR does that by setting `isLoaded` for global nav to 3rd state `null` in which the nav is being loaded this state is set once the filter is opened and nav has not been loaded yet. Once the nav is loaded no more requests are fired, you can open and close the app filter multipletimes while it is being processed and only one calculation will be fired.

animation

### Do not use global nav on landing nav

Since landing nav uses static list of nav items we don't have to use the globalNav there.

### DIFF
--- a/src/js/utils/useGlobalNav.js
+++ b/src/js/utils/useGlobalNav.js
@@ -5,7 +5,7 @@ import sourceOfTruth from '../nav/sourceOfTruth';
 
 const appIds = ['insights', 'openshift', 'cost-management', 'migrations', 'subscriptions', 'ansible', 'settings'];
 
-const useGlobalNav = () => {
+const useGlobalNav = (isOpen) => {
   const [state, setState] = useState({
     apps: [],
     filteredApps: [],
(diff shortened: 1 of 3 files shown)


Hyperkid123 (commented):
I think rather than passing the isOpen we should integrate it into this hook. It's now exclusively used just for the global filter so it makes more sense to me to define and control it from here rather than having an outside variable controlling it.

We will also remove one unnecessary render if we use one state variable instead of two.

Hyperkid123 (approved):

src/js/utils/useGlobalNav.js
```diff
@@ -7,20 +7,30 @@ const appIds = ['insights', 'openshift', 'cost-management', 'migrations', 'subsc
...
   const setFilteredApps = (filteredApps) => setState((prev) => ({ ...prev, filteredApps }));
+  const setIsOpen = (isOpen) => setState((prev) => ({ ...prev, isOpen }));
   useEffect(() => {
-    (async () => {
-      const navigationYml = await sourceOfTruth();
-      const appData = await getNavFromConfig(load(navigationYml), undefined);
-      setState({ apps: appIds.map((id) => appData[id]).filter((app) => !!app), filteredApps: appIds.map((id) => appData[id]), isLoaded: true });
-    })();
-  }, []);
+    if (state.isOpen === true && state.isLoaded === false) {
```
PanSpagetka, line 18: Just a one small thing. Should `isOpen` also have 3rd state? If not I would delete `=== true` so it is not confusing.

Hyperkid123: I am not sure why `x === true` would be confusing to anyone

PanSpagetka: In most of our codebase we dont use `x === true` just  `if(x)`. So, it suggest an idea that there might be third state that we care about. Exactly like `isLoaded` here. Which is probably not true for `isOpen`. it is a small thing and I don't really care about it. It was just thought that came to my mind when I was reading the code.

PR #1264, opened 2021-03-15
Optimize rendering in left navigation components

rvsia:
- better comparators for checking complex items
- useState replaced with useRef to not trigger additional render

**Known issue**

When you switch applications in the side nav, the navigation is rendered 2x - first when the click is done, second when identifyApp is triggered from the app. You can see it on the nav itself - when you switch you can see it took a while to 'select' the app:

Kapture 2021-03-15 at 12 41 11

@Hyperkid123

### DIFF
--- a/src/js/App/Sidenav/SideNav.js
+++ b/src/js/App/Sidenav/SideNav.js
@@ -12,9 +12,9 @@ import { globalNavComparator } from '../../utils/comparators';
 export const SideNav = () => {
   const dispatch = useDispatch();
 
-  const activeTechnology = useSelector(({ chrome: { activeTechnology } }) => activeTechnology);
-  const appId = useSelector(({ chrome: { appId } }) => appId);
-  const globalNav = useSelector(({ chrome: { globalNav } }) => globalNav, globalNavComparator);
+  const activeTechnology = useSelector(({ chrome }) => chrome?.activeTechnology);
+  const appId = useSelector(({ chrome }) => chrome?.appId);
+  const globalNav = useSelector(({ chrome }) => chrome?.globalNav, globalNavComparator);
   const isFirst = useRef(true);
 
   useEffect(() => {
--- a/src/js/App/Sidenav/SideNav.js
+++ b/src/js/App/Sidenav/SideNav.js
@@ -1,5 +1,4 @@
-import React, { Fragment, useEffect, useState } from 'react';
-import PropTypes from 'prop-types';
+import React, { Fragment, useEffect, useRef } from 'react';
 import Navigation from './Navigation';
 import { useDispatch, useSelector } from 'react-redux';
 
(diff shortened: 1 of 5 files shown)


Hyperkid123:
> When you switch applications in the side nav, the navigation is rendered 2x - first when the click is done, second when identifyApp is triggered from the app. You can see it on the nav itself - when you switch you can see it took a while to 'select' the app:

@rvsia you can memoize each individual nav item to prevent unnecessary renders right? Or do you mean only the top nav or each item is rendered?

src/js/App/Sidenav/SideNav.js
```diff
@@ -8,21 +7,26 @@ import { appNavClick } from '../../redux/actions';
...
+import { globalNavComparator } from '../../utils/comparators';
 
 export const SideNav = () => {
   const dispatch = useDispatch();
-  const { activeTechnology, globalNav, appId } = useSelector(({ chrome }) => chrome);
-  const [isFirst, setIsFirst] = useState(true);
+
+  const activeTechnology = useSelector(({ chrome: { activeTechnology } }) => activeTechnology);
+  const appId = useSelector(({ chrome: { appId } }) => appId);
+  const globalNav = useSelector(({ chrome: { globalNav } }) => globalNav, globalNavComparator);
```
karelhala, line 17: The chrome could be undefined, can you change it to safe accessor?
```suggestion
  const activeTechnology = useSelector(({ chrome }) => chrome?.activeTechnology);
  const appId = useSelector(({ chrome }) => chrome?.appId);
  const globalNav = useSelector(({ chrome }) => chrome?.globalNav, globalNavComparator);
```

rvsia: I am not sure that it could be undefined, when it's defined on redux init.

rvsia:
@Hyperkid123 memoizing nav items does not help it - there is onClick function that is using all the props from redux so the function is changed every time any of the used redux values is changed - so each item has to be re-rerendered.

src/js/App/Sidenav/SideNav.js
karelhala: The page with not logged in account has empty chrome and it caused a lot of problems in the past. I'd rather avoid it with the safe accessors.

karelhala (approved):

Hyperkid123:
@rvsia can we lift the data to the parent? Meaning don't even consume it in the nav item.
For example instead of calling `(myData) => onClick(parentDataA, parentDataB, myData)`, lift the function to parent like this:
```jsx
const handleClick = (myData) => callback(parentDataA, parentDataB, myData)
<Item onClick={handleClick} />
```

Or only consume the necessary item and memoize the component?

rvsia:
@Hyperkid123 I do not think this could help - the function is different on each render no matter what - even if you lift the function, the function has to be changed to consume the new parent function.

However, there is one way around this. It's a little bit ugly, but it should work well. Instead of passing values from redux, pass the getState function and get the values from it inside the function. What do you think?

Hyperkid123:
> However, there is one way around this. It's a little bit ugly, but it should work well. Instead of passing values from redux, pass the getState function and get the values from it inside the function. What do you think?

That is very ugly. Let's keep it as it is then.

PR #1270, opened 2021-03-17
When user navigates in chrome 2 enable global filer

karelhala:
### Global filter disabled when switching apps

When global filter is used in chrome 2 it can stay disabled when switching between apps, because it now remembers the state. This PR fixes such issue by enabling on new navigation (only for chrome 2 navigation)


src/js/App/Sidenav/Navigation.js
```diff
@@ -183,6 +183,7 @@ export const Navigation = () => {
          * Between chrome 2.0 apps navigation
          */
         !parent?.active && dispatch(clearActive());
+        dispatch(toggleGlobalFilter(false));
```
Hyperkid123, line 186: Won't this enable the global filter in other bundles than insights? For example the settings namespace has SPA enabled for multiple apps.

karelhala: Not really, the global filter is hidden behind `activeLocation === 'insights'`, this just sets the enable/disable state. Not the visibility of global filter.

Hyperkid123 (approved):

PR #1273, opened 2021-03-17
Remove old inventory async loader and add deprecated message

karelhala:
### Remove A/B testing

Since we can start using fed modules for inventory details the A/B testing is not really needed and we can remove it and start using the fed modules.

### Remove old inventory loader

Since we no longer support old way of loading inventory in chrome2 let's remove it and add deprecated message about it.


package.json
```diff
@@ -165,6 +161,7 @@
     "@scalprum/core": "0.0.10",
     "@scalprum/react-core": "0.0.15",
     "@sentry/browser": "^6.1.0",
+    "abortcontroller-polyfill": "^1.7.1",
```
Hyperkid123, line 164: Any reason for this dependency? I don't see it used.

karelhala: Inventory is using the abort for stopping some requests and it's not part of dependencies list.

Hyperkid123: Oh. I think we should add it to inventory and remove it from here (in a future PR)

Hyperkid123 (approved):

PR #1276, opened 2021-03-18
Optimize global filter rendering.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-12932

### Changes
- added memo wrapper around the root global filter component
- added `shallowEqual` to nonliteral selectors
- updated global filter reducer default state and use it
- split GlobalFilter dropdown into stand-alone component
  - this is for future updates, in order to decrease the render numbers even lower, we need to update the use `useTagsFilter` hook. It does several unnecessary renders but will require quite a significant refactoring (post-summit work)

We should also create a single redux action instead of 3 here: https://github.com/RedHatInsights/insights-chrome/blob/a7f8749a486269a9fc0bc8b290a4fcb15cc57ed1/src/js/App/GlobalFilter/GlobalFilter.js#L157 True it uses `batch` operation, but these are async actions that won't be executed in batch. Only the `_PENDING` actions will, the rest is async. This alone creates 3 renders instead of just one. I don't want to change it now before the summit to avoid breaking changes so close to the release. This is a common pattern around chrome and FEC so we should definitely re-visit it post-summit.

The number of renders on the mount went from 19 to 10. With future changes to the `useTagsFilter` and redux, we can go down to around 5 I think.

### DIFF
--- a/src/js/App/GlobalFilter/GlobalFilter.js
+++ b/src/js/App/GlobalFilter/GlobalFilter.js
@@ -1,144 +1,55 @@
-import React, { useEffect, Fragment, useState, useCallback } from 'react';
+import React, { useEffect, Fragment, useState, useCallback, memo, useMemo } from 'react';
+import PropTypes from 'prop-types';
 import { useSelector, useDispatch, batch, shallowEqual } from 'react-redux';
-import { GroupFilter } from '@redhat-cloud-services/frontend-components/components/cjs/ConditionalFilter';
-import { useTagsFilter } from '@redhat-cloud-services/frontend-components/components/cjs/FilterHooks';
-import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-components/components/cjs/Skeleton';
+import { GroupFilter } from '@redhat-cloud-services/frontend-components/ConditionalFilter';
+import { useTagsFilter } from '@redhat-cloud-services/frontend-components/FilterHooks';
+import { Skeleton, SkeletonSize } from '@redhat-cloud-services/frontend-components/Skeleton';
 import { fetchAllSIDs, fetchAllTags, fetchAllWorkloads, globalFilterChange } from '../../redux/actions';
-import { Split, SplitItem } from '@patternfly/react-core/dist/js/layouts/Split';
-import { Chip, ChipGroup } from '@patternfly/react-core/dist/js/components/ChipGroup';
-import { Button } from '@patternfly/react-core/dist/js/components/Button';
-import { Tooltip } from '@patternfly/react-core/dist/js/components/Tooltip';
+import { Button, Chip, ChipGroup, Split, SplitItem, Tooltip } from '@patternfly/react-core';
 import TagsModal from './TagsModal';
 import { workloads, updateSelected, storeFilter, generateFilter } from './constants';
 import debounce from 'lodash/debounce';
 import { useHistory } from 'react-router-dom';
 
-const GlobalFilter = () => {
-  const [hasAccess, setHasAccess] = useState(undefined);
-
-  const isAllowed = () => hasAccess;
-  const history = useHistory();
-
-  const [isOpen, setIsOpen] = useState(false);
-  const [token, setToken] = useState();
+const GlobalFilterDropdown = ({
+  isAllowed,
+  isDisabled,
+  userLoaded,
+  filter,
+  chips,
+  setValue,
+  selectedTags,
+  isOpen,
+  filterTagsBy,
+  filterScope,
+  setIsOpen,
+}) => {
+  const allowed = isAllowed();
   const dispatch = useDispatch();
-  const { isLoaded, count, total, sapCount, isDisabled } = useSelector(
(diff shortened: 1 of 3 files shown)


src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -23,19 +136,22 @@ const GlobalFilter = () => {
...
   const dispatch = useDispatch();
   const { isLoaded, count, total, sapCount, isDisabled } = useSelector(
-    ({ globalFilter: { tags, sid, workloads, globalFilterHidden }, chrome: { appId } }) => ({
-      isLoaded: tags?.isLoaded && sid?.isLoaded && workloads?.isLoaded,
-      count: tags?.count || 0 + sid?.count || 0 + workloads?.count || 0,
-      total: tags?.total || 0 + sid?.total || 0 + workloads?.total || 0,
-      sapCount: workloads?.hasSap,
-      isDisabled: globalFilterHidden || !appId,
-    }),
+    ({ globalFilter: { tags, sid, workloads, globalFilterHidden }, chrome: { appId } }) => {
```
rvsia, line 139: According to @karelhala chrome can be empty and we should add safe accessors.

src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -23,19 +136,22 @@ const GlobalFilter = () => {
...
+        sapCount: workloads.hasSap,
+        isDisabled: globalFilterHidden || !appId,
+      };
+    },
     shallowEqual
   );
-  const tags = useSelector(({ globalFilter: { tags } }) => tags?.items || []);
-  const sid = useSelector(({ globalFilter: { sid } }) => sid?.items || []);
+  const tags = useSelector(({ globalFilter: { tags } }) => tags.items || [], shallowEqual);
+  const sid = useSelector(({ globalFilter: { sid } }) => sid.items || [], shallowEqual);
```
rvsia, line 151: Because items contain arrays I am not sure if shallowEqual will do the trick.

src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -1,18 +1,131 @@
...
+  chips,
+  setValue,
+  selectedTags,
+  isOpen,
+  filterTagsBy,
+  filterScope,
+  setIsOpen,
+}) => {
+  const dispatch = useDispatch();
+  const GroupFilterWrapper = !isAllowed() || isDisabled ? Tooltip : ({ children }) => <Fragment>{children}</Fragment>;
```
karelhala, line 28: Wouldn't it be better to use state or useCallback for this? The `disabled` state won't change that much and this is calculated on each render.

src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -23,19 +136,22 @@ const GlobalFilter = () => {
...
   const { isLoaded, count, total, sapCount, isDisabled } = useSelector(
-    ({ globalFilter: { tags, sid, workloads, globalFilterHidden }, chrome: { appId } }) => ({
-      isLoaded: tags?.isLoaded && sid?.isLoaded && workloads?.isLoaded,
-      count: tags?.count || 0 + sid?.count || 0 + workloads?.count || 0,
-      total: tags?.total || 0 + sid?.total || 0 + workloads?.total || 0,
-      sapCount: workloads?.hasSap,
-      isDisabled: globalFilterHidden || !appId,
-    }),
+    ({ globalFilter: { tags, sid, workloads, globalFilterHidden }, chrome: { appId } }) => {
+      return {
```
karelhala, line 140: Unnecessary return statement.

src/js/App/GlobalFilter/GlobalFilter.js
karelhala: Good point, altough I would let this pass since the global filter can't be visible on unauth pages. But better be safe than sorry.

src/js/App/GlobalFilter/GlobalFilter.js
Hyperkid123: It will, and here is why. The only time when the filter should not update is when it's initializing. meaning when comparing `[] === []`. Any change to global filter results in tags and sid re-fetch and we will have to update no matter what. Plus, unfortunately, these arrays are circular structures (containing react components and functions) so not event `lodash.isEqaul` gives you the correct information.

I want to address that in post-summit refactoring and get rid of any components lurking in state management

src/js/App/GlobalFilter/GlobalFilter.js
Hyperkid123: This actually can't happen because the global filter is not visible when chrome is `undefined`. The visibility of the global filter depends on `chrome.activeLocation`. If chrome is undefined, the global filter will not be rendered.

src/js/App/GlobalFilter/GlobalFilter.js
Hyperkid123: Yup we need memo not callback here

rvsia (approved):

karelhala (approved):

PR #1280, opened 2021-03-22
Add a few tests

Hyperkid123:
### Changes
- removed extra `(` from app filter DOM (was showing in browser)
- added some ouiaIDs and normal IDs
- added tests for AppFilter, HeaderAlert and ToolbarToggle

### DIFF
--- a/src/js/App/Header/ToolbarToggle.js
+++ b/src/js/App/Header/ToolbarToggle.js
@@ -40,7 +40,7 @@ class ToolbarToggle extends Component {
     const dropdownItems = this.props.dropdownItems.map(({ url, title, onClick, isHidden, target = '_blank', rel = 'noopener noreferrer', ...rest }) =>
       !isHidden ? (
         <DropdownItem
-          key="item"
+          key={title}
           ouiaId={title}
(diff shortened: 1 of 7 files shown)


src/js/App/Header/ToolbarToggle.js
```diff
@@ -40,7 +40,7 @@ class ToolbarToggle extends Component {
     const dropdownItems = this.props.dropdownItems.map(({ url, title, onClick, isHidden, target = '_blank', rel = 'noopener noreferrer', ...rest }) =>
       !isHidden ? (
         <DropdownItem
-          key="item"
+          key={title}
```
rvsia, line 43: Just a suggestion for the future: add `key` and `oudaId` attributes to `dropdownItems` in case that title is not a string 🤔

rvsia (approved):

src/js/App/Header/ToolbarToggle.js
Hyperkid123: Yeah, this "should" always be string based on how we use this component. But we should definitely address this in the post-summit refactoring.

PR #1281, opened 2021-03-23
fix chrome heigth

Hyperkid123:
caused by: https://github.com/RedHatInsights/insights-chrome/pull/1278/files#diff-e953d69488a613b49aa81408a3595bd7161bf7547cca09d3378351b6c4552502L93

@mattnolting can you point me towards the issues you were fixing with this pr: https://github.com/RedHatInsights/insights-chrome/pull/1278? I want to check if I did not re-introduce them. If that is the case I will look for another solution.

### Before
screenshot-ci foo redhat com_1337-2021 03 23-09_15_53

### After
screenshot-ci foo redhat com_1337-2021 03 23-09_16_05

### DIFF
--- a/src/js/App/RootApp.js
+++ b/src/js/App/RootApp.js
@@ -140,6 +140,7 @@ const RootApp = ({ activeApp, activeLocation, appId, config, pageAction, pageObj
 
   return (
     <div
+      id="chrome-app-render-root"
       className="pf-c-drawer__content"
       data-ouia-subnav={activeApp}
       data-ouia-bundle={activeLocation}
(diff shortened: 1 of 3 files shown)


karelhala (approved):

mattnolting:
Screen Shot 2021-03-23 at 9 24 53 AM

Screen Shot 2021-03-23 at 9 25 17 AM

The issue with setting `height: 100%` is that, when `GlobalFilters` are present, page height becomes `100% + GlobalFilter height`, which then adds space at the end of the page, and that causes inconsistencies across apps. 

A better solution may be to revert #1278 and move `GlobalFilter` inside `pf-c-page__main`.

Hyperkid123:
@mattnolting we can't really move the global filter it will introduce other issues. However, I can create a workaround until we have chrome 2 adopted by all apps that should fix the 100% + global filter height issue.

Hyperkid123:
@mattnolting checking the dashboard and I don't see the extra space with this CSS rule. I think the problem was that before we applied the rule on multiple elements which broke the container element. Now the "extra" space is the main body padding which is fine.

screenshot-ci cloud redhat com-2021 03 25-13_44_53

PR #1285, opened 2021-03-24
Include parent for better navigation in appFilter

karelhala:
### Adjust nested app filter navigation

Since we want to surface some apps that are nested in bundles in appFilter (cost-management and subscriptions) we have to include the app ID for better navigation. This PR also allows using `reload` when navigating to each app in appFilter.


Hyperkid123 (approved):
Just a future post-summit refactoring work. I think we can share the URL generating function between the app filter and the left nav.

PR #1290, opened 2021-03-25
Populate entitlements on landing page as well

karelhala:
### No entitlements on landing page

Since we want to check if user has entitlements on landing page as well, call entitlements API on any page, if the page does not support entitlements continue as before. If the page supports entitlements and user has -1, show this information. If user has no entitlements for specific place bounce them to landing page.

### DIFF
--- a/src/js/jwt/insights/user.unit.test.js
+++ b/src/js/jwt/insights/user.unit.test.js
@@ -66,8 +66,8 @@ describe('User', () => {
   /* eslint-enable camelcase */
 
   describe('default', () => {
-    test('appends the entitlements data onto the user object', () => {
-      const o = user.default(token);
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@karelhala tests have failed

src/js/jwt/insights/user.unit.test.js
```diff
@@ -66,8 +66,8 @@ describe('User', () => {
   /* eslint-enable camelcase */
 
   describe('default', () => {
-    test('appends the entitlements data onto the user object', () => {
-      const o = user.default(token);
+    test.only('appends the entitlements data onto the user object', async () => {
```
Hyperkid123, line 69: leftover `only`

Hyperkid123 (approved):

PR #1298, opened 2021-03-29
Remove styles for frontend components

rvsia:
cc @Hyperkid123 @karelhala


src/sass/chrome.scss
```diff
@@ -25,7 +25,7 @@ $pf-global--image-path: './assets/images/';
 
 @import '~@redhat-cloud-services/frontend-components/Spinner/spinner.css';
 @import '~@redhat-cloud-services/frontend-components/Skeleton/skeleton.css';
-@import '~@redhat-cloud-services/frontend-components/ConditionalFilter/onditionalFilter.css';
+@import '~@redhat-cloud-services/frontend-components/ConditionalFilter/conditional-filter.css';
```
Hyperkid123, line 28: This css should not be required anymore. Was the style broken without this?

rvsia: Nope, the link just throws 404.. should I remove it?

Hyperkid123: Yeah, try to remove it. The v3 components inject styling with components. This is theoretically rendundant. So is the stuff above.

rvsia: Should I remove styles for all frontend components?

rvsia: Yeah, I should. Let me try it.

rvsia: Looks OK locally.

Hyperkid123 (approved):

PR #1299, opened 2021-03-29
Menu restyling: take 4

epwinchell:
Change external links icons to 12px

Screen Shot 2021-03-29 at 9 54 31 AM

### DIFF
--- a/src/js/App/Sidenav/SectionNav.scss
+++ b/src/js/App/Sidenav/SectionNav.scss
@@ -40,6 +40,10 @@
             .pf-c-nav__item {
               .pf-c-nav__link{
                 padding-left: var(--pf-global--spacer--sm);
+                svg {
+                  width: 12px;
+                  height: 12px;
+                }
               }
             }
           }
@@ -55,6 +59,10 @@
       --pf-c-nav__link--hover--after--BorderColor:  var(--pf-global--primary-color--light-100);
       a {
         padding-left: var(--pf-global--spacer--sm)
+      }
+      svg {
+        width: 12px;
+        height: 12px;
       }
     }
   }


Hyperkid123:
@epwinchell just curious here. Is it possible to honor PF pre-defined icon sizes? You can configure react icon with the `size` prop. The closest one is sm which is 10px. https://www.patternfly.org/v4/guidelines/icons/#icon-sizes

KendraMar:
Hey @Hyperkid123 -- the icon sizes PatternFly recommends (in our sketch file -- 10px is not promoted) are 54, 12, 24, 16.

karelhala (commented):
I don't think you have to repeat the size of svg in both places, but rather add it on line 30-32 https://github.com/RedHatInsights/insights-chrome/blob/master/src/js/App/Sidenav/SectionNav.scss#L30-L32 also, no need to use `width` and `height` but instead use `font-size: var(--pf-global--FontSize--xs)`

epwinchell:
`I don't think you have to repeat the size of svg in both places`

@karelhala I left it in because this is a WIP and might change again.

Hyperkid123:
Oh @KendraMar so the sketch library and the implementation is out of sync eh? That's unfortunate. Can you forward it to the PF team? And also what is the correct size then?

KendraMar:
Hey, @Hyperkid123 investigating with the pfly team...more later. Goal is to choose one size for the external icon that works visually for both the subsection and those category subsections.

src/js/App/Sidenav/SectionNav.scss
```diff
@@ -40,6 +40,10 @@
             .pf-c-nav__item {
               .pf-c-nav__link{
                 padding-left: var(--pf-global--spacer--sm);
+                svg {
+                  width: 12px;
```
Hyperkid123, line 44: Ok so rather than adding size to the SVG element itself, it's better to add font-size to the element or its parent. The SVG size is by default `1em` which is based on the font-size. Using height/width can distort icons that are not rectangles.

ryelo (approved):
Approving this for the time being, we can go back and fix later if there's a real issue.

PR #1303, opened 2021-03-30
Add module ID exception rules for sub apps.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-13280

This is a dumb solution but won't break existing module identification... Nothing more we can do before the nav re-work.


karelhala (approved):

Hyperkid123:
@karelhala sorry I forgot to delete the duplicate approval check. Should be ok now.

PR #1308, opened 2021-04-05
Masthead dropdown styling updates

epwinchell:
https://marvelapp.com/prototype/fgbch7a/screen/78188492

Old
Screen Shot 2021-04-05 at 11 07 50 AM

New
Screen Shot 2021-04-05 at 11 07 12 AM


karelhala (changes requested):
Don't do it like this, both of these dropdowns have `isPlain` prop set, you can remove it to achieve the same thing http://patternfly-react.surge.sh/components/dropdown#props

epwinchell:
`Don't do it like this, both of these dropdowns have isPlain prop set, you can remove it to achieve the same thing http://patternfly-react.surge.sh/components/dropdown#props`

@karelhala I'm not sure what you are suggesting. I remove"isPlain" with this PR.

karelhala:
@epwinchell oh I see! Can you remove those CSS styles (https://github.com/RedHatInsights/insights-chrome/blob/12c4f4af5826dd1b2a6e6b04d5b1b862b1c1e34a/src/js/App/Header/AppFilter.scss#L19-L22) and show the picture without them?

@mmenestr the borders around dropdowns are not the usual ones? This feels a bit off, I'd rather use default dropdown variations. If we want to change the borders, might be a better idea to propose this to PF team.

epwinchell:
`@epwinchell oh I see! Can you remove those CSS styles (`

@karelhala This was already reviewed with Kendra. The borders without the overrides are white, and we don't want that.

karelhala:
@epwinchell @KendraMar do you think you could drive this with PF? I don't like seing `important` in CSS and especially on cases like this. I guess PF should allow different variants of borders, either trough more specific CSS var, something like `pf-c-dropdown__toggle--BorderColor` or trough React prop.

epwinchell:
`I guess PF should allow different variants of borders, either trough more specific CSS var, something like pf-c-dropdown__toggle--BorderColor or trough React prop.`

@karel I'm using the border variables found under "css variables" here: https://www.patternfly.org/v4/components/dropdown

karelhala:
@epwinchell yeah, that's totally fine. I'd just like to see more straight forward sollution. Because this looks like a problem in our design system, that if the dropdown is used within darker area the borders don't look that good. I'd much rather see PF take care of it, because it might be a problem to others as well.

KendraMar:
@karelhala I'll ask Margot to take a look at PatternFly as you suggested.

ryelo (approved):
Good for now, this should be fixed in PF too, but that's not going to happen and be released before summit.

PR #1309, opened 2021-04-05
TEAMNADO-2098 Removing Subscription Central Nav Exception

joeldmyers:
After conversation with @karelhala, we changed the app name to be `manifests` to match the route.  As a result, the hotfix mapping subscription-central to manifests can be removed, since the bundle is now also called manifests.

/cc @Hyperkid123


karelhala (approved):
This change is reasonable, bu I am afraid it won't really fix the problem with loading the app.

Hyperkid123:
We will need a similar "exception" to approval. We plan to re-do the navigation both in chrome and CRC config post-summit. The current system was not "ready" for so many edge cases like this one. It became difficult to map between routes and apps.

PR #1316, opened 2021-04-08
Fix hiding/not hiding context switcher menu after click

PanSpagetka:
I found few more bugs, so I fixed them as well.
- Don't hide menu after clicking on search (original bug)
- Hide menu after clicking on other items
- Enable user to write and actually see text into search.

I am assuming that actual search functionality is not available yet, so search doesn't do anything.

https://issues.redhat.com/browse/RHCLOUD-12988

### DIFF
--- a/src/js/App/Header/ContextSwitcher.scss
+++ b/src/js/App/Header/ContextSwitcher.scss
@@ -11,7 +11,7 @@
       .pf-c-search-input {
         margin: 0 10px 0 10px;
         .pf-c-search-input__text .pf-c-search-input__text-input{
-          color: black;
+          color: var(--pf-global--Color--100);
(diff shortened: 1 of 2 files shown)


src/js/App/Header/ContextSwitcher.scss
```diff
@@ -10,6 +10,9 @@
       z-index: unset;
       .pf-c-search-input {
         margin: 0 10px 0 10px;
+        .pf-c-search-input__text .pf-c-search-input__text-input{
+          color: black;
```
Hyperkid123, line 14: pf variable please :slightly_smiling_face:

Hyperkid123 (approved):

PR #1319, opened 2021-04-09
Ugly fix for cost management active nav

karelhala:
### Two nav items active

When cost-management is activated it also shows as active openshift application. This PR is quick and dirty solution, I don't thing we can make it more pretty on such short notice. It will ignore the first nav item and render active just the nested nav.

##### Before

Screenshot from 2021-04-09 15-50-08

##### After
Screenshot from 2021-04-09 15-50-47


Hyperkid123 (approved):
:vomiting_face: The post-summit nav system redesign can't come soon enough!

PR #1320, opened 2021-04-09
Fix manifest link in menu

joeldmyers:
In the menu, the link for manifests is showing as /beta/subscriptions/manifests, when it should be /beta/insights/subscriptions/manifests.  

This is a proposed solution; open to other ideas.  

Screenshot from 2021-04-09 11-18-54

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5414,24 +5414,24 @@
       }
     },
     "@webpack-cli/configtest": {
-      "version": "1.0.1",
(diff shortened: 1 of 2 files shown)


karelhala (changes requested):
This is not really good approach, can you change it so it respects other apps as well?
```JS
const redirectUrl = `${parent ? `${parent.id}/` : ''}${subApp.reload || `${id}/${subApp.id}`}`;
```

PR #1327, opened 2021-04-12
Update prod-beta to newest master

karelhala:


ryelo:
@karelhala just double checking here that we're hiding application services in the app switcher and the console navigation here

karelhala:
@ryelo forgot to add it, this PR re-adds it https://github.com/RedHatInsights/insights-chrome/pull/1328

PR #1333, opened 2021-04-13
remove app filter icons

epwinchell:
Removing the bundle icons. I'm leaving the pf-split in for the time being, in case icons return...

Screen Shot 2021-04-13 at 12 28 52 PM


karelhala (approved):
:skull: RIP icons

PR #1338, opened 2021-04-14
Use new IA in App filter

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-13559


src/js/utils/useGlobalNav.js
```diff
@@ -1,16 +1,30 @@
...
+    ],
+  },
+  { id: 'ansible', routes: [{ id: 'automation-hub' }, { id: 'catalog' }, { id: 'automation-analytics' }] },
   {
-    title: 'Insights for SAP',
-    id: 'SAP',
-    routes: [{ id: 'sap-dashboard', title: 'Dashboard' }],
-    isEntitled: async () => await insights.chrome.auth.getUser().entitlements?.insights?.is_entitled,
+    id: 'settings',
+    routes: [{ id: 'my-user-access' }, { id: 'rbac' }, { id: 'sources' }, { id: 'integrations' }, { id: 'notifications' }, { id: 'applications' }],
```
Hyperkid123, line 27: As per the latest CRC changes, I think we should also exclude `notifications` from the list.

cc @karelhala

Hyperkid123 (approved):

PR #1343, opened 2021-04-15
append the app filter menu to document body

jschuler:
We have an issue here where the app filter dropdown ("Find an app or service") doesn't properly display when a quick start is open. Example before:
https://qaprodauth.cloud.redhat.com/beta/application-services/streams/resources?quickstart=kafka-bin-scripts
Screen Shot 2021-04-15 at 5 57 41 PM
Fix is to render the dropdown menu to document body using React.portal

After the change:
Screen Shot 2021-04-15 at 6 15 04 PM


Hyperkid123:
@jschuler I've briefly tested it in UI and the only thing I saw was the backdrop is also overlying the page header.

src/js/App/Header/AppFilter.js
```diff
@@ -66,43 +67,53 @@ const AppFilterDropdown = ({ isLoaded, setIsOpen, isOpen, filterValue, setFilter
...
-          </Flex>
-          {filteredApps?.length > 0 ? (
-            <div className="gallery">
-              {filteredApps.map((app) => (
-                <App key={app.id} {...app} />
-              ))}
+    {ReactDOM.createPortal(
+      <>
+        <div className="pf-c-dropdown ins-c-page__app-filter-dropdown-menu">
+          <div className="pf-c-dropdown__menu">
```
karelhala, line 73: Why not using the direct `Dropdown` component?

src/js/App/Header/AppFilter.js
```diff
@@ -66,43 +67,53 @@ const AppFilterDropdown = ({ isLoaded, setIsOpen, isOpen, filterValue, setFilter
...
-              onClear={() => setFilterValue('')}
-            />
-          </Flex>
-          {filteredApps?.length > 0 ? (
-            <div className="gallery">
-              {filteredApps.map((app) => (
-                <App key={app.id} {...app} />
-              ))}
+    {ReactDOM.createPortal(
+      <>
```
karelhala, line 71: I don't think you have to use Fragment here. But if that's true please use the full `React.Fragment` component. I tent to overlook these `<>`.

karelhala (commented):
While I don't mind where the component is rendered we should discuss the usage of drawer. We don't have custom layouts out as of yet so we should discuss how drawers are used. We should have a Drawer component in chrome so we don't have to hack our way around current layout and let apps choose what they render in said Drawer. With the help of FED modules this shouldn't prove difficuls as we can render the layout and let app tell us where the component for said drawer is located and render it inside of this Drawer.

I don't think we can do that now as summit is approaching, so once the comments are resolved and tests passed we can merge it and discuss more straight forward and cleaner approach to drawers.

jschuler:
@Hyperkid123 the problem with the drawer being in the DOM tree, is that it is then difficult to open up modals that don't cover up the quick start, because there is no container that spans the side nav and the main content area.
See this screenshot:
Screen Shot 2021-04-16 at 7 56 21 AM
That is why the drawer has to live outside of the root.
This is solved after moving the drawer out and appending the modals to a specific container:
Screen Shot 2021-04-15 at 10 54 43 AM

src/js/App/Header/AppFilter.js
jschuler: I wanted to create a really lightweight cloned structure to append to the document body

Hyperkid123:
> @Hyperkid123 the problem with the drawer being in the DOM tree, is that it is then difficult to open up modals that don't cover up the quick start, because there is no container that spans the side nav and the main content area.

@jschuler did you try using `appendTo` prop in the modal component? You should be able to find propper DOM element to append the modal to.

jschuler:
@Hyperkid123 are you able to access https://issues.redhat.com/browse/MGDSTRM-2855? The most recent comment there lists the 3 PRs that were merged to make this all work.
Feel free to take a look, but I am afraid there won;t be a solution that works when the drawer is inside the root :)
I think in general you would want the quick start to be outside the main app area, like the way OpenShift has it. Ideally it can remain open and will not be unmounted between route changes.
Screen Shot 2021-04-16 at 8 59 14 AM

Hyperkid123:
@jschuler yep I got access. @karelhala should we merge it for now? It won't kill any other app for now and should not be an issue once we are able to have multiple layouts.

karelhala:
@jschuler can you resolve the conflicts? I'll merge it after that.

karelhala (approved):

PR #1349, opened 2021-04-20
Fix spacing for external links

PanSpagetka:
Use same css class for external links from main.yaml as for those defined as extra links.

EDIT:  There was also issue about external link having bad left padding causing then not to be aligned with other items. That should be also fixed by this PR.

The first documentation link is from main.yaml and the seoond one from main.yaml.
Screenshot:
Screenshot from 2021-04-20 15-18-07


src/js/App/Sidenav/__snapshots__/NavigationItem.test.js.snap
```diff
@@ -3,13 +3,13 @@
...
   <li
-    class="pf-c-nav__item"
+    class="pf-c-nav__item undefined "
     data-ouia-component-id="OUIA-Generated-NavItem-3"
     data-ouia-component-type="PF4/NavItem"
     data-ouia-safe="true"
   >
     <a
-      class="pf-c-nav__link"
+      class="pf-c-nav__link undefined "
```
Hyperkid123, line 12: We are adding some undefined classnames here. You could use the `classNames` lib to avoid these.

Hyperkid123:
@PanSpagetka can you provide a screen? I am too lazy to check it locally.

ryelo (approved):

src/js/App/Sidenav/SectionNav.scss
```diff
@@ -43,6 +43,7 @@
                 svg {
                   width: 12px;
                   height: 12px;
+                  margin-left: var(--pf-global--spacer--sm)
```
ryelo, line 46: Missing `;`

src/js/App/Sidenav/SectionNav.scss
```diff
@@ -58,7 +59,7 @@
     .ins-c-navigation__additional-links{
       --pf-c-nav__link--hover--after--BorderColor:  var(--pf-global--primary-color--light-100);
       a {
-        padding-left: var(--pf-global--spacer--sm)
+        padding-left: var(--pf-global--spacer--lg)
```
ryelo, line 62: missing `;`

ryelo (changes requested):

karelhala (approved):

ryelo (approved):

PR #1361, opened 2021-04-22
Add conditional access to user identity.

Hyperkid123:
This will fix the runtime error but I am not sure if it fixes all the 401s we are getting in the landing page content. Maybe we are supposed to get them? I dunno. Stage is wierd.

### DIFF
--- a/src/js/rbac/fetchPermissions.js
+++ b/src/js/rbac/fetchPermissions.js
@@ -29,7 +29,7 @@ export const createFetchPermissionsWatcher = () => {
   let currentCall = {};
   return async (userToken, app = '') => {
     const user = await insights.chrome.auth.getUser();
-    if ([undefined, -1].includes(user.identity.account_number)) {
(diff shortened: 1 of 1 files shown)


src/js/rbac/fetchPermissions.js
```diff
@@ -29,7 +29,7 @@ export const createFetchPermissionsWatcher = () => {
   let currentCall = {};
   return async (userToken, app = '') => {
     const user = await insights.chrome.auth.getUser();
-    if ([undefined, -1].includes(user.identity.account_number)) {
+    if ([undefined, -1].includes(user?.identity?.account_number)) {
```
karelhala, line 32: This would always pass, what do you think about this one
```suggestion
    if (user?.identity && [undefined, -1].includes(user.identity.account_number)) {
```

Hyperkid123: Ah yeah I was not sure what is the actual condition. This makes sense

karelhala (approved):

PR #1366, opened 2021-04-26
Enable appfilter and landing nav on all environments

karelhala:


src/js/App/Header/Header.js
```diff
@@ -20,7 +18,7 @@ export const Header = ({ logoClassName }) => {
       <a href="./" className={`ins-c-header-link${logoClassName ? ` ${logoClassName}` : ''}`}>
         <Logo />
       </a>
-      {user && isFilterEnabled && <AppFilter />}
+      <AppFilter />
```
PanSpagetka, line 21: Do we want to show it without `user`?

PanSpagetka (approved):

src/js/App/Header/Header.js
Hyperkid123: @karelhala I don't think so. Maybe its for some special openshift use case?

karelhala:
Restarting travis

src/js/App/Header/Header.js
karelhala: It was a mistake, the filter should be visible only for logged in users.

PR #1382, opened 2021-04-27
Fix router POP action.

Hyperkid123:
When pressing the router back button, apps get full page to reload to the previous app, even if the history stack was in the nested app route.

We were using `!==` to compare **next browser location** that will become the current location with `prevLocation`. This has the opposite effect. Example flow

```jsx
prevLocationChrome = '/inventory'
currentHistoryLocationStack = ['/patch/advisories/nested', '/patch/advisories', '/inventory']

// POP event occurs
currentHistoryLocationStack = ['/patch/advisories', '/inventory']
nextLocation = /patch/advisories

/*
* Wrong result
* prevLocationChrome !== nextLocation
*/
'/inventory' !== '/patch/advisories'  // true
// Evaluated as true, we have incorrect redirect to prevLocation and routing error

/*
* Expected result
* prevLocationChrome === nextLocation 
*/
'/inventory' === '/patch/advisories' //false
/*
* Stack is not yet in prev app, no navigation trigger
* will occur on next POP event which is expected
*/
```

### DIFF
--- a/src/js/App/Sidenav/Navigation.js
+++ b/src/js/App/Sidenav/Navigation.js
@@ -117,7 +117,10 @@ export const Navigation = () => {
   const settings = useSelector(({ chrome }) => chrome?.globalNav, globalNavComparator);
   const dispatch = useDispatch();
   const history = useHistory();
-  const prevLocation = useRef(window.location.pathname);
+  /**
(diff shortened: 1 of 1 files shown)


karelhala (commented):
Let's test this in CI and prod-beta before fast forwarding to prod-stable.

karelhala (approved):

PR #1385, opened 2021-04-27
Use full page redirect to beta environment if user clicks on use beta feature

karelhala:


src/js/App/Sidenav/Navigation.js
```diff
@@ -240,10 +241,13 @@ export const Navigation = () => {
...
         isOpen={showBetaModal}
-        onClick={() => {
-          onClick(...deferedOnClickArgs);
-          isBeta() || (window.location = switchRelease(false, window.location.pathname));
-          setShowBetaModal(false);
+        onClick={(event) => {
+          if (!isBeta()) {
+            const [origEvent, item, parent] = deferedOnClickArgs;
+            const isMetaKey = event.ctrlKey || event.metaKey || event.which === 2 || origEvent.ctrlKey || origEvent.metaKey || origEvent.which === 2;
+            const url = `${basepath}'beta/${activeLocation || ''}/${item.reload || (parent ? `${parent.id}/${item.id}` : item.id)}`;
```
Hyperkid123, line 248:
```jsx
const url = `${basepath}'beta/${activeLocation || ''}/${item.reload || (parent ? `${parent.id}/${item.id}` : item.id)}`;
```

There is an extra `'`after the `${basepath}'`. Is that wanted?

karelhala: Yeah, good catch!

Hyperkid123 (approved):

PR #1400, opened 2021-05-03
Fix 'hasPendingAjax' checking fetchResults

john-dupuy:
We use this in iqe-core to check when a page is ready for selenium interaction. I noticed that it was returning `False` when XHR requests were still pending.

### DIFF
--- a/src/js/chrome/iqeEnablement.js
+++ b/src/js/chrome/iqeEnablement.js
@@ -73,7 +73,7 @@ export default {
     xhrResults.map((e) => console.log(`[iqe] xhr incomplete: ${e._url}`)); // eslint-disable-line no-console
     Object.values(fetchResults).map((e) => console.log(`[iqe] fetch incomplete: ${e}`)); // eslint-disable-line no-console
 
-    return xhrResults.length > 0 || fetchResults.hasOwnProperty('length') ? fetchResults.length : Object.values(fetchResults).length > 0; // eslint-disable-line no-prototype-builtins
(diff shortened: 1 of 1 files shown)


src/js/chrome/iqeEnablement.js
```diff
@@ -73,7 +73,7 @@ export default {
     xhrResults.map((e) => console.log(`[iqe] xhr incomplete: ${e._url}`)); // eslint-disable-line no-console
     Object.values(fetchResults).map((e) => console.log(`[iqe] fetch incomplete: ${e}`)); // eslint-disable-line no-console
 
-    return xhrResults.length > 0 || fetchResults.length > 0;
+    return xhrResults.length > 0 || fetchResults.hasOwnProperty('length') ? fetchResults.length : Object.values(fetchResults).length > 0; // eslint-disable-line no-prototype-builtins
```
karelhala, line 76: You don't really have to check for the `length` property. The `Array` is object as well so you can simplify this to
```suggestion
    return xhrResults.length > 0 || Object.values(fetchResults).length > 0;
```

Hyperkid123: It's my fault for suggesting that. I have seen this piece of code for the first time and did not know what was the original intent.

john-dupuy: @karelhala thanks! that sounds good to me, suggestion applied

karelhala (approved):

PR #1401, opened 2021-05-04
Update App filter items

PanSpagetka:
There is Edge bonus item that is not on the mocks. But I assume that it should be there?

#### Screenshot
Screenshot from 2021-05-04 12-04-10

https://issues.redhat.com/browse/RHCLOUD-13888

### DIFF
--- a/src/js/utils/useGlobalNav.js
+++ b/src/js/utils/useGlobalNav.js
@@ -3,7 +3,7 @@ import { load } from 'js-yaml';
 import { getNavFromConfig } from '../nav/globalNav';
 import sourceOfTruth from '../nav/sourceOfTruth';
 
-const appIds = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'settings'];
+const appIds = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'subscriptions', 'cost-management', 'settings'];
 const useGlobalNav = () => {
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

src/js/utils/useGlobalNav.js
```diff
@@ -3,7 +3,7 @@ import { load } from 'js-yaml';
 import { getNavFromConfig } from '../nav/globalNav';
 import sourceOfTruth from '../nav/sourceOfTruth';
 
-const appIds = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'settings'];
+const appIds = ['application-services', 'openshift', 'insights', 'ansible', 'subscriptions', 'cost-management', 'edge', 'settings'];
```
karelhala, line 6: Can you put the `edge` before `ansible`?
```suggestion
const appIds = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'subscriptions', 'cost-management', 'settings'];
```

karelhala (approved):

ryelo:
Do we want to show edge here? @karelhala do we need to gate that behind something?

karelhala:
@ryelo we are not deploying it to prod for now. I'll ask around and check if it is fine to surface on prod-beta.

PR #1419, opened 2021-05-07
use the right function for feedback

ryelo:


src/js/App/Feedback.js
```diff
@@ -24,7 +24,7 @@ const Feedback = ({ user }) => {
         },
         body: JSON.stringify({
           description: `Feedback: ${textAreaValue}, Username: ${user.identity.user.username}, Account ID: ${user.identity.account_number}, Email: ${user.identity.user.email}, URL: ${window.location.href}`, //eslint-disable-line
-          summary: `${!window.insights.isProd && '[PRE-PROD]'} Insights Feedback`,
+          summary: `${!window.insights.chrome.isProd && '[PRE-PROD]'} Insights Feedback`,
```
Hyperkid123, line 27: Hm i think that `${false && "bla"}X` will result in `falseX`. It was probably there the whole time?

ryelo: This was reporting everything as 

[PRE-PROD] Insights Feedback

It shouldn't show anything other than `Insights Feedback` on prod.

Hyperkid123: Huh was it? When you try running the template string in the console it will prefix the string with the condition result and calls `toString` no matter the outcome. So you can end up with an `undefined Insights Feedback` string.

Magic?

Hyperkid123: Not magic the `!` confused me :slightly_smiling_face:

Hyperkid123 (approved):

coderbydesign (approved):

PR #1423, opened 2021-05-11
Remove label from OS column

rvsia:
see https://github.com/RedHatInsights/frontend-components/pull/1129


rvsia:
cc @karelhala @Hyperkid123

PR #1426, opened 2021-05-12
add ability to hide nav item

ryelo:


src/js/App/Sidenav/NavigationItem.js
```diff
@@ -8,11 +8,11 @@ import classnames from 'classnames';
 import './Navigation.scss';
 
 const basepath = document.baseURI;
-const NavigationItem = ({ itemID, title, parent, navigate, ignoreCase, className, isBeta, isLoading, ...props }) => (
+const NavigationItem = ({ itemID, title, parent, navigate, ignoreCase, className, isBeta, isLoading, isHidden, ...props }) => (
   <NavItem
     {...props}
     itemId={itemID}
-    className={classnames(className, { 'ins-c-navigation__additional-links': !isLoading && navigate })}
+    className={classnames(className, { 'ins-c-navigation__additional-links': !isLoading && navigate }, { 'ins-c-navigation__isHidden': isHidden })}
```
Hyperkid123, line 15: I am not sure if we want to even add the link to the dom. I would just skip the rendering of the link altogether. Jus return `null` in the render function. It should still work.

Hyperkid123 (approved):

karelhala (approved):

PR #1438, opened 2021-05-25
Supress header from hotjar recordings

karelhala:


Hyperkid123 (approved):
Based purely on the hotjar docs this seems good to me.

PR #1439, opened 2021-05-26
Add jira labels to feedback form

PanSpagetka:
Submitting form doesn't work in my dev setup. I am getting `Bad request` response, however this change should match API specs, bud it needs to be verified after this PR is merged.

https://issues.redhat.com/browse/RHCLOUD-14051


Hyperkid123:
@PanSpagetka can you try it in prod env? It should work there.

Hyperkid123:
We will just have to try it once deployed

karelhala (approved):

PR #1441, opened 2021-05-27
Upgrade to npm7

Hyperkid123:
### Changes
- enforce node 15+ and npm 7+
- remove unused dependencies
- update all dependencies
- stop building FontAwesome (it's not used)
- remove style lint

Please make sure that I have not removed any critical dependency used only in deployment.


src/js/inventory/inventoryStyles.js
```diff
@@ -1,9 +0,0 @@
-import '@patternfly/patternfly/components/Toolbar/toolbar.css';
```
karelhala, line 1: I am a bit sceptical about these changes, but we can only see it fail once we deploy that. Let's see on ci-beta, afterall this env is supposed to be for testing integrations.

karelhala (approved):

src/js/inventory/inventoryStyles.js
Hyperkid123: All of this should be imported with the components :crossed_fingers: . We will see, we can always put it back.

PR #1444, opened 2021-06-01
Do not mix PF cjs/esm imports.

Hyperkid123:
Fixes nav context toggle issues.

### Changes
- stop using absolute PF imports
  - it was causing some context issue
  - it's not really necessary because federated modules will pull the whole library because of nested dependencies anyway
- removed some legacy imports from FEC modules

Was outputting a console error but the navigation still worked. @john-dupuy could you check if this helps with the highlight issues? I would guess it does not.


john-dupuy:
@Hyperkid123 yes, post-merge I can check this out.

karelhala (approved):

PR #1461, opened 2021-06-08
fix login on unauthed pages

ryelo:


src/js/jwt/jwt.js
```diff
@@ -287,6 +287,19 @@ function refreshTokens() {
   authChannel.postMessage({ type: 'refresh' });
 }
 
+const shouldPageAuth = (path) => {
+  if(path === '/') {
+    return false
+  }
+
+  const adjustedPath = path.replace(/\/$/, "");
+  if(allowedUnauthedPaths.includes(adjustedPath)) {
```
Hyperkid123, line 296: I think we could just use this instead of the whole `if` statement.
```jsx
return allowedUnauthedPaths.includes(path.replace(/\/$/, ""))
```

ryelo: I don't think this will work for the '/' page?

Hyperkid123: I meant for the second if

Hyperkid123: to get rid of the `if(x) {return false} else {return true}` we can just `return !x`.

Hyperkid123 (approved):

PR #1474, opened 2021-06-18
Fix feedback summary string.

Hyperkid123:
If the `!window.insights.chrome.isProd` condition was false, it will output that `false` result into the template string.

~~@ryelo is it supposed to show the `[PROD]` string in the summary or do we want the flag only for pre-prod environments?~~

OK, the `[PROD]` prefix should exist.


ryelo (approved):

ryelo:
Only show it for pre-prod

Hyperkid123:
> Only show it for pre-prod

@ryelo I have received an executive order to add the `[PROD]` to prod :guard:

PR #1483, opened 2021-06-23
Update token in cookie on token refresh.

Hyperkid123:
maybe fixes: https://issues.redhat.com/browse/RHCLOUD-14530

The most important part is this commit: https://github.com/RedHatInsights/insights-chrome/pull/1483/commits/2b205ccc545f46874eae51505d2d9aa3ef5b5097

I've taken the opportunity and made a class from the `priv` object to make it a bit safer.

~~Needs some testing in prod.~~

### DIFF
--- a/src/js/jwt/modules/useChromeAuth.js
+++ b/src/js/jwt/modules/useChromeAuth.js
@@ -40,14 +40,15 @@ export const initChromeAuth = () => {
 
   wipePostbackParamsThatAreNotForUs();
   const token = cookie.get(options.cookieName);
+  const refreshToken = cookie.get('cs_jwt_refresh');
 
   // If we find an existing token, use it
   // so that we dont auth even when a valid token is present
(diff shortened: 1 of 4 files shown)


Hyperkid123:
I can't test this directly in openshoft. The API does not include `prod.foo` in their CORS policy. We will just have to wait and see.

src/js/jwt/modules/useChromeAuth.js
```diff
@@ -40,16 +40,18 @@ export const initChromeAuth = () => {
...
   // otherwise its quick, but we bounce around and get a new token
   // on every page load
-  if (token && token.length > 10) {
+  if (token && token.length > 10 && refreshToken && refreshToken.length > 10) {
+    options.refreshToken = refreshToken;
     options.token = token;
-    options.refreshToken = cookie.get('cs_jwt_refresh');
   }
 
+  console.log(options);
```
rvsia, line 54: Console log here

PR #1492, opened 2021-06-29
Proper savings planner redirect

karelhala:
### Saving planner redirect

Since ansible tower analytics routes lives under `ansible/insights` and new route has been added https://github.com/RedHatInsights/cloud-services-config/pull/649 we have to update the list of routes in chrome as well, otherwise we'd still be redirected to `ansbile/savings-planner` instead of `ansible/insights/savings-planner`

### DIFF
--- a/src/js/App/Sidenav/SectionNav.js
+++ b/src/js/App/Sidenav/SectionNav.js
@@ -1,0 +1,89 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { NavGroup } from '@patternfly/react-core';
+import ExpandableNav from './ExpandableNav';
+import WrenchIcon from '@patternfly/react-icons/dist/js/icons/wrench-icon';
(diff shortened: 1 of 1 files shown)


Hyperkid123 (commented):
@karelhala I don't think these changes will have any impact. I think that adding new routes to the module's config will do the trick (at least in ci-beta). You probably want to add this to prod instead?

ryelo (approved):

ryelo:
This will probably go out before the rest of the nav stuff, so we can just cherry pick it or merge it straight to all the other envs

Hyperkid123:
@ryelo yeah but this won't have absolutely any effect in ci-beta. The file was deleted and is not used. Can we just change the target branch?

ryelo:
works for me

karelhala:
@Hyperkid123 good point!

PR #1494, opened 2021-06-30
Quickstarts provider

jschuler:
Add API for consumers to set and toggle quick starts. Also consumer apps could pull in the Catalog, although I am wondering if we should instead add the catalog globally within insights-chrome?

Example consumer (RBAC)
https://github.com/jschuler/insights-rbac-ui/pull/1/files


src/js/App/RootApp.js
```diff
@@ -109,10 +111,33 @@ RootApp.propTypes = {
...
-      <RootApp {...props} />
-    </ScalprumProvider>
+    <QuickStartContext.Provider value={valuesForQuickstartContext}>
+      <QuickStartDrawer>
+        <ScalprumProvider
+          config={config}
+          api={{
+            chrome: {
+              experimentalApi: true,
+              setQuickStarts: valuesForQuickstartContext.setAllQuickStarts,
```
jschuler, line 131: thoughts on API? Should we nest them? i.e. could be something like
```
chrome: {
  experimentalApi: true,
  quickStarts: {
    set: valuesForQuickstartContext.setAllQuickStarts,
    toggle: valuesForQuickstartContext.setActiveQuickStart.
    Catalog: QuickStartCatalogPage
  }
}
```

Hyperkid123: Definitely nest the API, we can create specific hooks like `useQuickstarts` to logically separate them. I would even suggest having it under a separate key (move it out of the chrome object).

Hyperkid123:
> Also consumer apps could pull in the Catalog, although I am wondering if we should instead add the catalog globally within insights-chrome?

@jschuler that depends on what the "Catalog" is. Anything that is common for every application/consumer should be within the chrome. Anything that is application-specific should be in the application code base and exported as a federated module.

Hyperkid123:
@jschuler I was able to run the quickstart locally with the RBAC UI. So I would say it's a success.

However, the tests are failing because you have direct imports to the ESM version of the PF icons in the quickstart build (for example: https://github.com/patternfly/patternfly-quickstarts/blob/0d438746c48614a77ef08752787fa45afbc32166/packages/module/src/ConsoleShared/src/components/status/icons.tsx)

That will either have to change to relative imports or change the absolute path to `/js` or make sure that the CJS build output transforms that to the CJS modules. We can force jest to transpile the modules for now, but I would not want that in production.

src/js/App/RootApp.js
```diff
@@ -14,6 +14,9 @@ import isEqual from 'lodash/isEqual';
 import { onToggle } from '../redux/actions';
 import Routes from './Routes';
 import useOuiaTags from '../utils/useOuiaTags';
+import { QuickStartDrawer, QuickStartContext, useValuesForQuickStartContext, useLocalStorage, QuickStartCatalogPage } from '@patternfly/quickstarts';
+import '@patternfly/quickstarts/dist/quickstarts.css';
```
Hyperkid123, line 18: Can you import these in the `chrome.scss` file please?

src/js/App/RootApp.js
```diff
@@ -109,16 +112,42 @@ RootApp.propTypes = {
...
+        <ScalprumProvider
+          config={config}
+          api={{
+            chrome: {
+              experimentalApi: true,
+              ...window.insights.chrome,
+              quickStarts: {
+                set: valuesForQuickstartContext.setAllQuickStarts,
+                toggle: valuesForQuickstartContext.setActiveQuickStart,
+                Catalog: QuickStartCatalogPage,
```
Hyperkid123, line 142: Can you lazy load this component and add a suspense wrapper around it?

src/js/App/Sidenav/Navigation/ChromeLink.js
```diff
@@ -60,8 +60,9 @@ const LinkWrapper = ({ href, isBeta, onLinkClick, className, currAppId, appId, c
 
     dispatch(appNavClick({ id: actionId }, domEvent));
   };
+  const quickStartHighlightId = href.split('/').slice(-1);
```
Hyperkid123, line 63: I am thinking that maybe we should just use the full `href` string to avoid duplicate ids.

src/js/App/RootApp.js
jschuler: Yes, also I am currently cleaning up the global styles. PR will be on hold until that is fixed

src/js/App/Sidenav/Navigation/ChromeLink.js
jschuler: i can't use `/` in the highlight id so i turn them into underscores

src/js/App/RootApp.js
```diff
@@ -17,6 +18,24 @@ import useOuiaTags from '../utils/useOuiaTags';
 import Banner from './Banners/Banner';
 import cookie from 'js-cookie';
 
+const QuickStartCatalog = lazy(() =>
+  import(/* webpackChunkName: "quick-start" */ './QuickStartCatalog').then(({ QuickStartCatalog }) => ({ default: QuickStartCatalog }))
```
Hyperkid123, line 22: We usually use default exports. Could you change the `QuickStartCatalog.js` file to use default export and remove the `then` branch?

src/js/App/RootApp.js
```diff
@@ -17,6 +18,24 @@ import useOuiaTags from '../utils/useOuiaTags';
...
+);
+
+const LazyQuickStartCatalog = ({ ...props }) => {
```
Hyperkid123, line 25: Can we move this component to a standalone file. The RootApp is getting a bit cluttered.

Hyperkid123 (commented):
I tested the quickstarts in RBAC UI and it's working great. I got only a few nitpicks for the chrome changes.

Hyperkid123 (approved):

PR #1500, opened 2021-07-01
Allow first level apps by checking active app Id and nav Id

karelhala:
### Incorrect navId

When in app that lives on first level: 
- /edge/fleet-management
- /edge/manage-images
- /openshift/overview
- /openshift/releases
- /openshift/downloads

Clicking on any nav item will result in sending nav event with `navId: '/'`, this is incorrect as it should check for active appId in chrome, if it's the same as the one user clicked in navigation we shouldn't navigate user to root. This PR fixes such issue by checking if current navId is same as appId. And if so do not assume we want to navigate to root.

### DIFF
--- a/src/js/App/Sidenav/Navigation/ChromeLink.js
+++ b/src/js/App/Sidenav/Navigation/ChromeLink.js
@@ -26,14 +26,15 @@ const useDynamicModule = (appId) => {
   return isDynamic;
 };
 
-const LinkWrapper = ({ href, isBeta, onLinkClick, className, children }) => {
+const LinkWrapper = ({ href, isBeta, onLinkClick, className, currAppId, appId, children }) => {
   let actionId = href.split('/').slice(2).join('/');
   if (actionId.includes('/')) {
     actionId = actionId.split('/').pop();
   }
-  if (href.split('/').length === 3) {
+  if (currAppId !== appId && href.split('/').length === 3) {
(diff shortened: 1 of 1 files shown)


src/js/App/Sidenav/Navigation/ChromeLink.js
```diff
@@ -116,15 +119,17 @@ RefreshLink.propTypes = {
 
 const ChromeLink = ({ appId, children, ...rest }) => {
   const { onLinkClick } = useContext(NavContext);
+  const currAppId = useSelector(({ chrome }) => chrome?.appId);
   const isDynamic = useDynamicModule(appId);
+  console.log(appId, 'this is appId!');
```
rvsia, line 124: This is a console log!

karelhala, line 124:
```suggestion
```

Hyperkid123 (approved):
Thanks, @karelhala. Seems OK.

I was thinking about long-term solutions and how we can completely avoid any splitting, conditions, etc. I think the best course of action for the future is to deprecate all this `actionId` `appId` stuff and use the only thing that is always 100% accurate which is the `href` attribute. We will never have an issue since it always points to the correct route. 

Apps will be able to use it by replacing the app router basename in the href. I will start pushing it after we have the nav changes in all environments.

PR #1501, opened 2021-07-02
Add ErrorComponent for Scalprum

Hyperkid123:
### Changes
- Log application uncaught exceptions to sentry
- Make the error component a little bit nicer

(the HAHA error title is expected that is what I've thrown in the application.)

screenshot-ci foo redhat com_1337-2021 07 02-11_46_43

screenshot-ci foo redhat com_1337-2021 07 02-11_46_54


src/js/App/ErrorComponent/ErrorComponent.js
```diff
@@ -0,0 +1,80 @@
...
+  FlexItem,
+} from '@patternfly/react-core';
+import ErrorCircleOIcon from '@patternfly/react-icons/dist/js/icons/error-circle-o-icon';
+import { getUrl } from '../../utils';
+
+import './ErrorComponent.scss';
+
+const ErrorComponent = (props) => {
+  useEffect(() => {
+    Sentry.captureException(new Error('Unhandled UI runtime error'), {
```
Hyperkid123, line 25: @ryelo is correct? Do we have different sentry instances for different apps? ideally, it should log the error for the specific application.

karelhala: The sentry logs it based on the bundle etc, AFAIK we don't have to check which app is currently active.

karelhala (approved):

src/js/App/ErrorComponent/ErrorComponent.js
ryelo: You can do this above and you're fine. You don't have to check anything in particular.

PR #1504, opened 2021-07-06
Add global index.html to be used as loader for all apps

karelhala:


src/pug/templates/index.pug
```diff
@@ -0,0 +1,9 @@
+doctype html
+html(lang='en-US')
+    head
+        meta(charset='UTF-8')
+        title
+            Chrome | console.redhat.com
+        include ../head.pug
+    body
+        include ../body.pug
```
Hyperkid123, line 9: Once this takes an effect, we would only really need a simple `<div id="foo" />` and the chrome script tags. Everything else should be then injected as a React component.

karelhala: Yup, I kept this in here just for the simplicity.

src/pug/templates/head-block.pug
```diff
@@ -1,7 +1,8 @@
 meta(http-equiv="x-ua-compatible" content="ie=edge")
 meta(name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no")
-base(href=`${ release === 'insights' ? '/' : '/beta/'}`)
+base(href=`${ release === 'insightsbeta' ? '/beta' : '/'}`)
```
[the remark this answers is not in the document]
karelhala: Good question, it probably is required there.

Hyperkid123 (approved):
@karelhala I know we need this for testing. So we can merge this. But we should update the chrome index with the beta base path ASAP.

PR #1505, opened 2021-07-06
replace fileMapper with `${name}-navigation.json`

redallen:
When adding a nav item in https://github.com/RedHatInsights/cloud-services-config I find it unreasonable to also have to update insight-chrome's `src/js/utils/useNavigation.js`.

It looks like the only name not in accordance is `insights` -> `rhel` so I've moved the file in https://github.com/RedHatInsights/cloud-services-config/pull/735.

If a mapper is really necessary, I believe it should come from cloud-services-config and I can update https://github.com/RedHatInsights/cloud-services-config/pull/735.

### DIFF
--- a/src/js/App/Header/AppFilter/useAppFilter.js
+++ b/src/js/App/Header/AppFilter/useAppFilter.js
@@ -3,7 +3,6 @@ import { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { isBeta } from '../../../utils';
 import { evaluateVisibility } from '../../../utils/isNavItemVisible';
-import { navigationFileMapper } from '../../../utils/useNavigation';
 
 export const requiredBundles = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'settings'];
 const bundlesOrder = ['application-services', 'openshift', 'rhel', 'edge', 'ansible', 'settings', 'cost-management', 'subscriptions'];
@@ -127,7 +126,7 @@ const useAppFilter = () => {
         ...prev,
         isLoading: true,
       }));
-      let bundles = requiredBundles.filter((app) => !existingSchemas.includes(app)).map((app) => navigationFileMapper[app]);
+      let bundles = requiredBundles.filter((app) => !existingSchemas.includes(app));
       bundles.map((fragment) =>
         axios
           .get(`${isBetaEnv ? '/beta' : ''}/config/chrome/${fragment}`)
--- a/src/js/App/Header/AppFilter/useAppFilter.js
+++ b/src/js/App/Header/AppFilter/useAppFilter.js
@@ -129,7 +129,7 @@ const useAppFilter = () => {
       let bundles = requiredBundles.filter((app) => !existingSchemas.includes(app));
       bundles.map((fragment) =>
         axios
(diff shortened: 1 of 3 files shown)


src/js/App/Header/AppFilter/useAppFilter.js
```diff
@@ -3,7 +3,6 @@ import { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { isBeta } from '../../../utils';
 import { evaluateVisibility } from '../../../utils/isNavItemVisible';
-import { navigationFileMapper } from '../../../utils/useNavigation';
 
 export const requiredBundles = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'settings'];
 const bundlesOrder = ['application-services', 'openshift', 'rhel', 'edge', 'ansible', 'settings', 'cost-management', 'subscriptions'];
```
karelhala, line 8: Just FYI if you are not going to add the `cloud-tutorials` in here it won't be shown in the app filter.

src/js/App/Header/AppFilter/useAppFilter.js
```diff
@@ -3,7 +3,6 @@ import { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { isBeta } from '../../../utils';
 import { evaluateVisibility } from '../../../utils/isNavItemVisible';
-import { navigationFileMapper } from '../../../utils/useNavigation';
 
 export const requiredBundles = ['application-services', 'openshift', 'insights', 'edge', 'ansible', 'settings'];
```
karelhala, line 7: The `cloud-tutorials` bundle should be added here so chrome knows about it.

src/js/App/Header/AppFilter/useAppFilter.js
```diff
@@ -127,7 +126,7 @@ const useAppFilter = () => {
         ...prev,
         isLoading: true,
       }));
-      let bundles = requiredBundles.filter((app) => !existingSchemas.includes(app)).map((app) => navigationFileMapper[app]);
+      let bundles = requiredBundles.filter((app) => !existingSchemas.includes(app));
```
karelhala, line 129: This still requires the `-navigation.json` partial, otherwise we'll fetch `/config/chrome/insights` which would be 404.

karelhala (commented):
While I am not against automatic mapping we've specified the nav in a way that users have to strictly define the nav and which bundle matches which navigation. New bundles are not that common thing, however renaming bundles is a very common thing and we need a better way that just renaming a file - it is more maintanable to change the mapper key here based on the URL and we don't have to synch the URL change with config file rename.

src/js/App/Header/AppFilter/useAppFilter.js
redallen: That's okay! We'd like to keep it hidden for now.

src/js/App/Header/AppFilter/useAppFilter.js
redallen: Good catch, fixed!

karelhala (approved):

Hyperkid123:
@redallen thanks for generalizing it a bit.

PR #1511, opened 2021-07-12
remove get/set ROS specific cookie search

kgaikwad:
With [JIRA card](https://issues.redhat.com/browse/RHIROS-201), removing ROS specific cookie check from respective repositories.
+adding @karelhala into loop.


src/js/chrome/entry.js
```diff
@@ -84,8 +84,6 @@ export function chromeInit() {
 }
 
 export function bootstrap(libjwt, initFunc, getUser) {
-  const searchParams = new URLSearchParams(location.search);
```
kgaikwad, line 87: @karelhala - I removed whole logic here as only ROS specific cookie exists. is it ok?

Hyperkid123 (approved):

PR #1525, opened 2021-07-19
Update logo and landing nav title

ryelo:
Screen Shot 2021-07-19 at 3 05 52 PM


karelhala (approved):
Let's keep it here, but we should also add it to the static assets repo as well.

ryelo:
@karelhala there's a bunch of different ones that I will be adding there too

PR #1528, opened 2021-07-20
Deploy nav changes to prod beta

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15395

@karelhala @ryelo  I had to do a manual merge commit it does the following
- removes the Section nav file which also contained savings planner "hack": https://github.com/RedHatInsights/insights-chrome/pull/1493

As far as I know, the issues should be fixed by the new nav so there should be no need for them.

We aim for Wednesday, July 21st
cc @john-dupuy


karelhala (approved):
I've created a new tag v2.2.1 so we have point in time to fallback in case of any bugs.

Hyperkid123:
@karelhala I think we can go ahead and merge it.

PR #1529, opened 2021-07-20
Disable unauthed pages.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15295

Merge on July, 29th

### Changes
- remove landing page from the list of allowed un authed pages
- use spinner in routes instead of `null` before initialization is complete


karelhala (approved):

Hyperkid123:
@karelhala there was a slight update of the plan. Chrome will no longer allow any unauthed pages. This will hit the prod when the migration to the console domain will happen. Until then we can disable it in all pre-prod environments to get people used to it.

src/js/auth.js
```diff
@@ -33,6 +34,12 @@ function getAllowedUnauthedPaths() {
 }
 
 export function allowUnauthed() {
+  /**
+   * We can delete the whole logic after console.dot migration is over
+   */
+  if (getEnv() !== 'prod') {
```
karelhala, line 40: I would check here for the `console` in URL. If that is present we will return `false` all the time. I would presonally add new helper function exposed on window to check this as well.

Hyperkid123: Yeah, it struck my mind too but we basically have to stop the chrome code promotion again until the console switch happens. I don't think we want to that ATM.

Hyperkid123:
@karelhala @ryelo I've updated the PR to just simply disable any un authed content. This can sit here until 29th.

ryelo (approved):

Issue #1531, opened 2021-07-20
The chrome.auth.getToken() should wait for the jwt initPromise like the getUser()


b1zzu:
Hi, 
let's start by saying that this is my opinion and we can actually work around this issue in different ways, but in the application-service-ui we use the getToken()[1] function very early to retrieve the user token. If the getToken() function is executed before the initPromise is completed, the getToken() will restart the SSO authentication because it thinks we are unauthenticated and this could go into an endless loop. 

The cleanest solution is to concatenate the initPromise to the getToken like for the getUser()[2], but before working on the PR, it would like to understand if this sounds like a good solution and if it could have drawbacks.

1. https://github.com/RedHatInsights/insights-chrome/blob/e5b15492de45fa37d7831b5557ba70aae6744dc7/src/js/chrome/entry.js#L92
2. https://github.com/RedHatInsights/insights-chrome/blob/e5b15492de45fa37d7831b5557ba70aae6744dc7/src/js/chrome/create-chrome.js#L42

b1zzu:
cc: @pmuir @ajaypratap003

Hyperkid123:
@b1zzu This issue is not possible to happen if the UI has been migrated and leveraging chrome 2. Could you send us a link to your repo to check when that is the case or not?

b1zzu:
@Hyperkid123 wrote you a private message

b1zzu:
Hi @Hyperkid123, I've opened a PR with my personal solution, wdyt?

Hyperkid123:
Yeah, I wanted to look at it tomorrow. I need to test that it won't break unauthenticated pages. That's the only issue I can imagine that could possibly happen.

Hyperkid123:
I will close the issue once we get the fix in all envs and @b1zzu confirms the issue is solved.

b1zzu:
@Hyperkid123 can you ping me when I can test it on the qaprodauth environment!?

Hyperkid123:
@b1zzu yep

Hyperkid123:
@b1zzu it should be deployed

b1zzu:
@Hyperkid123 I confirm the issue is fixed, thanks

PR #1532, opened 2021-07-20
Wait for jwt initPromise to be completed before retrieving the token

b1zzu:
**Why**
Avoid endless loop if the getToken is called before the jwt initPromise is completed

fix: #1531


Hyperkid123 (approved):
Works good. :100:

PR #1533, opened 2021-07-21
Fix chrome link error before modules are loaded.

Hyperkid123:
Fixes crashing app filter on un authed pages.


karelhala (approved):

Hyperkid123:
This will have to make its way to prod beta.

PR #1550, opened 2021-07-26
Use inventory fed modules

karelhala:
### Inventory fed modules fallback

This is just a temporary fallback, before apps fully migrate to the new loader trough frontend components.


Hyperkid123 (approved):

Hyperkid123:
Adding label until we can merge this.

PR #1571, opened 2021-08-03
Add 404 route.

Hyperkid123:
Because the un authed pages are no longer with us, we now handle the 404 route as well. This PR adds a 404 page as a fallback to the react-router (even non-chrome 2 apps are handled via react-router.).

### Before
Plain text saying "Not found" 

### After
screenshot-ci foo redhat com_1337-2021 08 03-10_00_50

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13917,8 +13917,7 @@
     "node_modules/require-main-filename": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/require-main-filename/-/require-main-filename-2.0.0.tgz",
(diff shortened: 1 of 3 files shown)


rvsia (approved):

ryelo (approved):

ryelo:
Any reason the lockfile is changing here?

Hyperkid123:
> Any reason the lockfile is changing here?

:thinking: you are right. I just ran npm install and I guess some stuff changed.. Do you want me to remove it?

ryelo:
It looks like integrity stuff, so I don't think it's an issue

PR #1574, opened 2021-08-05
add flag to cookie

ryelo:


ryelo:
this doesn't actually work, would have to move cookie logic to a server and rewrite this implementation because httponly cookies can't be read by js

PR #1589, opened 2021-08-24
Introduce custom global filter menu component.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15651

### Changes
- use "custom" select component for global filter

Why? Because in a new version of PF the select children are always one render cycle behind. For example, if you select a value, the checkbox will be set only after you select another one and so on.

This might be also used a base for the new conditional filter implementation. This no longer uses a lot of hacks required before.

### DIFF
--- a/src/js/App/GlobalFilter/GlobalFilterMenu.js
+++ b/src/js/App/GlobalFilter/GlobalFilterMenu.js
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { TextInput, MenuList, MenuItem, Select, SelectVariant, MenuGroup, Checkbox } from '@patternfly/react-core';
@@ -16,37 +16,28 @@ const getMenuItems = (groups, filterValueRegex, onChange, calculateSelected) => 
   const result = groups.map(({ value, label, id, type, items, ...group }) => ({
     label,
     value,
-    items: items
-      .filter(
-        (item) =>
-          group.noFilter ||
-          (value && filterValueRegex.test(value)) ||
-          (label && filterValueRegex.test(label)) ||
-          (item.value && filterValueRegex.test(item.value)) ||
-          (item.label && filterValueRegex.test(item.label))
-      )
-      .map((item, index) => ({
-        ...item,
-        key: item.id || item.value || index,
-        value: String(item.value || item.id || index),
-        onClick: (event) => {
-          onChange(
-            event,
-            calculateSelected(type, value, item.value),
-            {
-              value,
-              label,
-              id,
-              type,
-              items,
-              ...group,
-            },
-            item,
+    items: items.map((item, index) => ({
+      ...item,
+      key: item.id || item.value || index,
+      value: String(item.value || item.id || index),
+      onClick: (event) => {
+        onChange(
+          event,
+          calculateSelected(type, value, item.value),
+          {
             value,
-            item.value
-          );
-        },
-      })),
+            label,
+            id,
+            type,
+            items,
+            ...group,
+          },
+          item,
+          value,
+          item.value
+        );
+      },
+    })),
   }));
   return result.filter(({ noFilter, items = [] }) => noFilter || items.length > 0);
 };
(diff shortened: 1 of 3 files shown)


src/js/App/GlobalFilter/GlobalFilterMenu.js
```diff
@@ -0,0 +1,169 @@
...
+      })),
+  }));
+  return result.filter(({ noFilter, items = [] }) => noFilter || items.length > 0);
+};
+
+const GlobalFilterMenu = (props) => {
+  const { filterBy, onFilter, groups = [], onChange, selectedTags } = props;
+  const [isOpen, setIsOpen] = useState(false);
+
+  const calculateSelected = (type, groupKey, itemKey) => {
```
karelhala, line 58: I'd move this function out of `GlobalFilterMenu` component, this way it's recreated everytime new render cycle happens. I know you have to remeber current `selectedTags` so how about wrapping it with another function.

```JS
const calculateSelected = (selectedTags) => (type, groupKey, itemKey) => {
    const activeGroup = selectedTags[groupKey];
    if (activeGroup) {
      if (type !== groupType.radio && (activeGroup[itemKey] instanceof Object ? activeGroup[itemKey].isSelected : Boolean(activeGroup[itemKey]))) {
        return {
          ...selectedTags,
          [groupKey]: {
            ...(activeGroup || {}),
            [itemKey]: false,
          },
        };
      }

      return {
        ...selectedTags,
        [groupKey]: {
          ...(type !== groupType.radio ? activeGroup || {} : {}),
          [itemKey]: true,
        },
      };
    }

    return {
      ...selectedTags,
      [groupKey]: {
        [itemKey]: true,
      },
    };
};
```

src/js/App/GlobalFilter/GlobalFilterMenu.js
```diff
@@ -0,0 +1,169 @@
...
+  };
+  const sanitizedFilterRegex = useMemo(() => {
+    try {
+      return new RegExp(filterBy, 'i');
+    } catch (err) {
+      return new RegExp(filterBy.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'i');
+    }
+  }, [filterBy]);
+
+  const menuItems = getMenuItems(groups, sanitizedFilterRegex, onChange, calculateSelected);
```
karelhala, line 95: The usage of `calculateSelected` out of this component can be here
```suggestion
  const menuItems = getMenuItems(groups, sanitizedFilterRegex, onChange, calculateSelected(selectedTags));
```

src/js/App/GlobalFilter/GlobalFilterMenu.js
```diff
@@ -0,0 +1,169 @@
...
+  const menu = [
+    <div key="x" className="pf-c-menu ins-c-global-filter__menu" style={{ boxShadow: 'none' }}>
```
karelhala, line 97: I'd change the key to `menu`, I know it's just for the react to know when to re-render, but the `x` doesn't tell us anything

src/js/App/GlobalFilter/GlobalFilterMenu.js
```diff
@@ -0,0 +1,169 @@
...
+    value,
+    items: items
+      .filter(
+        (item) =>
+          group.noFilter ||
+          (value && filterValueRegex.test(value)) ||
+          (label && filterValueRegex.test(label)) ||
+          (item.value && filterValueRegex.test(item.value)) ||
+          (item.label && filterValueRegex.test(item.label))
+      )
```
karelhala, line 27: The `filter` from original implementation was there because Select required some filtering in the past, it wasn't working without it. Can you try it without this filter? The component will shrink even more.

src/js/App/GlobalFilter/GlobalFilterMenu.js
```diff
@@ -0,0 +1,169 @@
...
+            {items.map(({ value, label, onClick, ...props }) => (
+              <MenuItem key={value} onClick={onClick}>
+                <Checkbox
+                  className="ins-c-global-filter__checkbox"
+                  // eslint-disable-next-line react/prop-types
+                  id={props.id}
+                  // eslint-disable-next-line react/prop-types
+                  isChecked={!!Object.values(selectedTags).find((group = {}) => group[props.tagKey]?.isSelected)}
+                  label={label}
+                />
```
karelhala, line 110: This should change based on the `type`, however for the chrome's implementation it doesn't have to be here.

karelhala (commented):
Looking hella good! We can remove a bunch of hacks from the Conditional filter group implementation.

src/js/App/GlobalFilter/GlobalFilterMenu.js
Hyperkid123: Oh yeah, it was just a placeholder :D I totally wanted to use something sane.

src/js/App/GlobalFilter/GlobalFilterMenu.js
Hyperkid123: Yes, I am aware. I think we can leave the different types for the FEC version.

src/js/App/GlobalFilter/GlobalFilterMenu.js
Hyperkid123: Well, this will re-create the function as well. It's just gonna be wrapped. I think using `useCallback` is probably the best option here.

karelhala: The `useCallback` won't help us much in here as we'd have to write complex update function for the `selectedTags`, but well at least the function can be outside of the component making the component more readable.

src/js/App/GlobalFilter/GlobalFilterMenu.js
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      return {
+        ...selectedTags,
+        [groupKey]: {
+          [itemKey]: true,
+        },
+      };
+    },
+    [selectedTags]
```
karelhala, line 79: What if you select multiple tags? Does it recalculate?

Hyperkid123: Yeah, you will get a new object.

src/js/App/GlobalFilter/GlobalFilterMenu.js
Hyperkid123: > to write complex update function 

Why? The only mutable variable is the `selectedTags`. We add it as the effect argument and it should be just fine. The rest of the attributes comes as a function args

karelhala: Mhh, we are copying them to new object, so the react should re-calculate the function on each change of `selectedTags` so it should be fine.

karelhala (approved):

PR #1602, opened 2021-08-30
App filter fixes

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15897

### Changes
- allow overriding default filterable checks with `filterable: true` flag
- change app filter search input text
- fix order of app filter items with async checks (they were always resolved last)


karelhala (approved):

Hyperkid123:
@karelhala note, we need this in prod QUICKLY. I will promote it later this week.

PR #1617, opened 2021-09-02
Fix case creation app info request

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15928

Instead of location which sometimes returns incorrect application name, we now use the actual module id which has always the correct app id.

The app id sometimes has to be transformed into kebab-case due to a mismatch of the module ID and the appname. The mismatch is caused by webpack remote containers name requirements (can't have `-` character).


john-dupuy (approved):
Verified that this fixes the issue in MUA by running this PR locally, note that the info is still not pre-filled:
https://issues.redhat.com/browse/RHCLOUD-15829

PR #1620, opened 2021-09-08
Add bundle and appId to Url on user-pref redirect

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-14545

When navigating to user-preferences `bundle` name and `appId` of the previous location are passed to the URL as a hash eg.
`https://ci.foo.redhat.com:1337/user-preferences/email#bundle=settings&appId=rbac`

### DIFF
--- a/src/js/App/Header/UserToggle.js
+++ b/src/js/App/Header/UserToggle.js
@@ -3,10 +3,10 @@ import { DropdownToggle, KebabToggle, DropdownItem, DropdownSeparator, DropdownP
 import UserIcon from './UserIcon';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { isBeta } from '../../utils';
+import { isBeta, getUrl } from '../../utils';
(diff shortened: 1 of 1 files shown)


src/js/App/Header/UserToggle.js
```diff
@@ -51,7 +51,7 @@ function buildItems(username, isOrgAdmin, accountNumber = -1, isInternal, extraI
       {accountNumber > -1 && (
         <DropdownItem
           component={
-            <ChromeLink href="/user-preferences/email" isBeta={isBeta()} appId="userPreferences">
+            <ChromeLink href={`/user-preferences/email#bundle=${getUrl('bundle')}&appId=${appId}`} isBeta={isBeta()} appId="userPreferences">
```
Hyperkid123, line 54: I think we want to exclude the bundle if its `undefined` that can happen on landing page.

Hyperkid123 (approved):

PR #1621, opened 2021-09-08
Fedramp apps

ryelo:


Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-14323

There is one open question about modules that are in multiple places. Currently, we only have `isFedramp` flag which marks the whole module as fedramp viable regardless of its placement (rhel vs ansible). If it turns out, a module is viable only in one place, we will have to use a more granular config.

Hyperkid123:
I will address the Ansible variants of the rhel modules in a separate PR.

Hyperkid123 (approved):

PR #1631, opened 2021-09-14
Add bypassCache option to getUserPermissions

rvsia:
We have encountered a bug in Sources: Sources now require write permissions that users can set up in RBAC application, but because both of these application are in the same bundle, Chrome is not being reloaded and the old permissions are cached, so users have to do hard refresh. This can create confusion amongst users.

cc @karelhala @Hyperkid123 

It would be great to backport this PR to prod-stable/beta and stable branches.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "insights-chrome",
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
--- a/src/js/rbac/fetchPermissions.js
+++ b/src/js/rbac/fetchPermissions.js
@@ -27,12 +27,12 @@ const fetchPermissions = (userToken, app = '') => {
 
 export const createFetchPermissionsWatcher = () => {
   let currentCall = {};
-  return async (userToken, app = '') => {
+  return async (userToken, app = '', bypassCache) => {
     const user = await insights.chrome.auth.getUser();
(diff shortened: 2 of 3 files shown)


Hyperkid123:
@rvsia you can pass appid to the function. 
```jsx
window.insights?.chrome?.getUserPermissions('sources')
```

Can you try it before we merge this PR?

rvsia:
@Hyperkid123 https://github.com/RedHatInsights/sources-ui/blob/07c56104090f741a7a283ef2bf86e2e6047ea69f/src/redux/user/actions.js#L7

Hyperkid123 (approved):

rvsia:
@Hyperkid123 Otherwise, the option is already there - https://github.com/RedHatInsights/insights-chrome/blob/4b84084bfa98cd3d30f1a3864901289115b02a85/src/js/chrome/create-chrome.js#L66 - it's not just being passed to the function and so it does nothing.

I think we could implement some time restraints in the future to avoid more situations like this one - for example, caching results only for one minute or something like that.

Hyperkid123:
@rvsia yeah I know its there. We have disabled it to reduce the number of calls to the API intentionally. But now when we have SPA the number of calls is limited quite a lot. I don't think sources will bring the API down

rvsia:
@Hyperkid123 Sources are loading the permissions only once when users enter the app - so it should not be an issue.

PR #1641, opened 2021-09-15
Change how resolve url loader resolves fonts in single index.html

karelhala:
### Fixes incorrect font imports

When running the single html build use different approach to importing fonts as they are not placed in the buil folder, but resolved in build time. When running the default sass build we are not calculating the path for these fonts, we take the original PF value and paste the fonts in correct build folder. Once the browser requests them they will be served from this build folder.

JIRA: https://issues.redhat.com/browse/RHCLOUD-16108


Hyperkid123 (approved):
I don't really like we have to do this sort of thing. I would consider using CDN for PF styles in the future and don't even compile them.

PR #1656, opened 2021-10-05
Single html template

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15879

This prepares single HTML template build.

### Prerequisites
- all apps are migrated to chrome 2 and deployed to prod (missing automation hub)
- Akamai/Fakamai is ready to serve only one HTML template

### DIFF
--- a/config/webpack.plugins.js
+++ b/config/webpack.plugins.js
@@ -55,8 +55,8 @@ const plugins = [
   }),
   new HtmlWebpackPlugin({
     title: 'Authenticating - console.redhat.com',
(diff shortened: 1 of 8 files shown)


karelhala (approved):

karelhala:
Looking good! Can you make the selectors in a way they will work with current and future single HTML as well? This way we can merge this PR and start slowly migrate all apps to use the chrome's index instead of their own `index.html`.

karelhala (approved):

PR #1659, opened 2021-10-12
Refactor navigation to allow dynamic navigation

karelhala:
### Refactor navigation loading

When we are using `useNavigation` hook we are posting the nav schema to redux, this PR slightly adjusts such thing by merging previous schema with the new one and highlights currently visited navigation. This allows for removing some of the logic from complex `useNavigation` hook and move it to simpler function in reducer.

### Allow dynamic navigation

Dynamic navigation is still defined in CSC by adding `"dynamicNav": "$appId/$section"` where `$section` is optional. This way we can make sure the order of navigation, if some app adds additional navItem in navigation it's up to them, but we'll make sure it's positioned on the right place in the navigation.

The way how we fetch the nav is by including `Navigation` FedModule and calling `default` export off it. This way application can do any computation on their side while doing so we'd show skeleton, and once the data arives we'll show the navigation element and ping the application to show the proper route.

### TODO

* [x] Styling - skeleton is not styled properly
* [x] Tests - both refactor and new tests
* [ ] Documentation

animation

### DIFF
--- a/src/js/redux/reducers.test.js
+++ b/src/js/redux/reducers.test.js
@@ -106,4 +106,96 @@ describe('Reducers', () => {
       });
     });
   });
+
+  describe.only('loadNavigationSegmentReducer', () => {
+    const navigation = { test: { navItems: [], sortedLinks: [] } };
+    it('should create new segment', () => {
+      const newNav = {
+        navItems: [
+          {
+            href: '/something',
+          },
+        ],
+      };
+      const result = reducers.loadNavigationSegmentReducer(
+        {
+          navigation: {},
+        },
+        {
+          payload: {
+            segment: 'test',
+            schema: newNav,
+          },
+        }
+      );
+      expect(result).toEqual({
+        navigation: {
+          ...navigation,
+          test: {
+            ...navigation.test,
+            navItems: newNav.navItems,
+            sortedLinks: ['/something'],
+          },
+        },
+      });
+    });
+
+    it('should merge two schema together', () => {
+      const newNav = { navItems: [{ href: '/another' }, { href: '/different' }] };
+      const result = reducers.loadNavigationSegmentReducer(
+        {
+          navigation: {
+            ...navigation,
+            test: {
+              ...navigation.test,
+              navItems: [{ href: '/something' }],
+              sortedLinks: ['/something'],
+            },
+          },
+        },
+        {
+          payload: {
+            segment: 'test',
+            schema: newNav,
+          },
+        }
+      );
+      expect(result).toEqual({
+        navigation: {
+          ...navigation,
+          test: {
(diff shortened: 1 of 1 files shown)


src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -0,0 +1,44 @@
...
+import { loadLeftNavSegment } from '../../../redux/actions';
+
+const DynamicNav = ({ dynamicNav, ...rest }) => {
+  const [nav, setNav] = useState();
+  const {
+    location: { pathname },
+  } = useHistory();
+  const dispatch = useDispatch();
+  const [appName] = dynamicNav.split('/');
+  const currentNamespace = pathname.split('/')[1];
```
Hyperkid123, line 17: I think we have a helper function for this somewhere. And if not, the same split is definitely done somewhere in the nav already. Might be worth extracting it.

src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -0,0 +1,44 @@
...
+  const schema = useSelector(({ chrome: { navigation } }) => navigation[currentNamespace]);
+  const { default: navigation } = useModule(appName, './Navigation', {});
```
Hyperkid123, line 19: Is the empty object argument required?

src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    if (navigation) {
+      if (typeof navigation === 'function') {
+        Promise.resolve(navigation({ schema, dynamicNav, currentNamespace })).then((data) => {
+          schema.navItems = schema?.navItems?.map((item, key) => ({
+            ...item,
+            ...(key === schema.navItems.findIndex((nav) => nav.dynamicNav === dynamicNav) && data),
+          }));
+          dispatch(loadLeftNavSegment(schema, currentNamespace, window.location.pathname));
```
Hyperkid123, line 28: Is there a reason to use window location rather than router location? The only difference between those two should only ever be the `/beta` prefix.

src/js/App/Sidenav/Navigation/DynamicNav.js
karelhala: Yup, the `useModule` uses state and when we use the destruct here `default: navigation` it would throw an error of destructing `undefined`

src/js/App/Sidenav/Navigation/DynamicNav.js
karelhala: It's used on couple of places, but not direct export. I'll create one in `utils` file.

PR #1663, opened 2021-10-13
Use module route in expandbale items in app filter links.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16616

### Changes
- try to use the module route before using truncated `href` partial

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "insights-chrome",
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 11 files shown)


src/js/App/Header/AppFilter/useAppFilter.js
```diff
@@ -10,6 +10,14 @@ const bundlesOrder = ['application-services', 'openshift', 'rhel', 'edge', 'ansi
 
 const isFedrampEnv = isFedRamp();
 
+function findModuleByLink(href, { modules } = { modules: [] }) {
+  const routes = modules
+    .map(({ routes }) => routes.map((route) => (typeof route === 'string' ? route : route.pathname)))
+    .flat()
```
gkarat, line 16: can be replaced by .flatMap?

Hyperkid123: Yup it can.

PR #1671, opened 2021-10-26
Add product to support case creation 

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15829

### Changes
- add `product` to case creation query

support-case

### DIFF
--- a/src/js/createCase.js
+++ b/src/js/createCase.js
@@ -1,6 +1,7 @@
 import Cookies from 'js-cookie';
 import * as Sentry from '@sentry/browser';
 import logger from './jwt/logger';
+import URI from 'urijs';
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@john-dupuy the issue with the support tickets was resolved. The correct fields should be now pre-filled.

john-dupuy:
@Hyperkid123 Testing this out in QA, I noticed the env was set to Production Beta and the commit hash was set to null - could this be a result of running this locally?
Screenshot from 2021-12-06 06-43-34

john-dupuy (approved):

Hyperkid123:
@john-dupuy  Yeah it works only prod :slightly_frowning_face:

PR #1684, opened 2021-11-05
Update scalprum react core to newest version and use load module

karelhala:


src/js/utils/useNavigation.js
```diff
@@ -46,7 +46,7 @@ const useNavigation = () => {
   const { pathname } = location;
   const { activeQuickStartID } = useContext(QuickStartContext);
   const currentNamespace = pathname.split('/')[1];
-  const schema = useSelector(({ chrome: { navigation } }) => navigation[currentNamespace]);
+  const [schema] = useSelector(({ chrome: { navigation, activeApp } }) => [navigation[currentNamespace], activeApp]);
```
karelhala, line 49: I had to check for active app to trigger re-render. Otherwise the nav wouldn't change :disappointed:

Hyperkid123 (approved):

PR #1685, opened 2021-11-05
Allow dynamic nav to have flat structure

karelhala:
### Flat navigation structure

Since dynamic nav can be also flat let's update the code so we can consume flat structure. Also remove merging of these schemas, if the schema is not present or the actions sent merging of schema together, just replace them.

### DIFF
--- a/src/js/App/Sidenav/Navigation/DynamicNav.js
+++ b/src/js/App/Sidenav/Navigation/DynamicNav.js
@@ -19,27 +19,35 @@ const DynamicNav = ({ dynamicNav }) => {
   useEffect(() => {
     if (navigation) {
       if (typeof navigation === 'function') {
-        Promise.resolve(navigation({ schema, dynamicNav, currentNamespace })).then((data) => {
-          const indexOfDynamicNav = schema.navItems.findIndex((nav) => nav.dynamicNav === dynamicNav);
-          delete schema.navItems[indexOfDynamicNav].dynamicNav;
-          if (Array.isArray(data)) {
-            schema.navItems.splice(
-              indexOfDynamicNav,
-              1,
-              ...data.map((item) => ({
-                appId: dynamicNav.split('/')[0],
-                ...schema.navItems[indexOfDynamicNav],
-                ...item,
-              }))
+        const indexOfDynamicNav = schema.navItems.findIndex((nav) => nav.dynamicNav === dynamicNav);
+        if (indexOfDynamicNav !== -1) {
+          const { dynamicNav: _dynamicNav, ...originalNav } = schema.navItems[indexOfDynamicNav];
+          Promise.resolve(navigation({ schema, dynamicNav, currentNamespace })).then((data) => {
+            const newValue = Array.isArray(data)
+              ? data.map((item) => ({
+                  appId: dynamicNav.split('/')[0],
+                  ...originalNav,
+                  ...item,
+                }))
+              : [
+                  {
+                    ...originalNav,
+                    ...data,
+                  },
+                ];
+            dispatch(
+              loadLeftNavSegment(
+                {
+                  ...schema,
+                  navItems: schema.navItems.flatMap((item, key) => (key === indexOfDynamicNav ? newValue : item)),
+                },
+                currentNamespace,
+                window.location.pathname,
+                true
+              )
             );
-          } else {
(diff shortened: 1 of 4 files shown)


src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -20,11 +20,25 @@ const DynamicNav = ({ dynamicNav }) => {
     if (navigation) {
       if (typeof navigation === 'function') {
         Promise.resolve(navigation({ schema, dynamicNav, currentNamespace })).then((data) => {
-          schema.navItems = schema?.navItems?.map((item, key) => ({
-            ...item,
-            ...(key === schema.navItems.findIndex((nav) => nav.dynamicNav === dynamicNav) && data),
-          }));
-          dispatch(loadLeftNavSegment(schema, currentNamespace, window.location.pathname));
+          const indexOfDynamicNav = schema.navItems.findIndex((nav) => nav.dynamicNav === dynamicNav);
+          delete schema.navItems[indexOfDynamicNav].dynamicNav;
```
andrewballantyne, line 24: You do not guard against a negative index (if it's not found)... is that possible?

andrewballantyne (changes requested):

src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -20,11 +20,25 @@ const DynamicNav = ({ dynamicNav }) => {
...
+                ...item,
+              }))
+            );
+          } else {
+            schema.navItems = schema?.navItems?.map((item, key) => ({
+              ...item,
+              ...(key === indexOfDynamicNav && data),
+            }));
+          }
+          dispatch(loadLeftNavSegment(schema, currentNamespace, window.location.pathname, true));
```
andrewballantyne, line 41: `splice` and `schema.navItems =` mutate the schema (Redux) object. I think you want something that builds a new schema:

```
const start = schema.navItems.slice(0, Math.min(indexOfDynamicNav, 0));
const end = schema.navItems.slice(indexOfDynamicNav + 1);

const newContent = Array.isArray(data)
  ? data.map((item) => ({
      appId: dynamicNav.split('/')[0],
      ...schema.navItems[indexOfDynamicNav],
      ...item,
    }))
  : [data];

const updatedSchema = {
  ...schema,
  navItems: [
    ...start,
    ...newContent,
    ...end,
  ],
}

dispatch(loadLeftNavSegment(updatedSchema, currentNamespace, window.location.pathname, true));
```

andrewballantyne (changes requested):
As I was reviewing your other PR I followed the code back here and noticed one more thing. Feel free to change this code up a bit if you want... I just think it's best we avoid doing any Redux object mutations.

src/js/App/Sidenav/Navigation/DynamicNav.js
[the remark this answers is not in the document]
karelhala: Good point, it is a leftover.

karelhala: Well the dynamic nav brought us here, but I like defensive programming, so let's add it here!

src/js/App/Sidenav/Navigation/DynamicNav.js
karelhala: Yeah, good point. This shouldn't affect the schema, since we are mutating it right away. But it's better to do it this way.

andrewballantyne (approved):

src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -19,13 +19,35 @@ const DynamicNav = ({ dynamicNav }) => {
...
+                  },
+                ];
+            dispatch(
+              loadLeftNavSegment(
+                {
+                  ...schema,
+                  navItems: schema.navItems.flatMap((item, key) => (key === indexOfDynamicNav ? newValue : item)),
+                },
+                currentNamespace,
+                window.location.pathname,
```
Hyperkid123, line 45: Can we use router location instead? The window will use the `/beta` prefix but we have that as a `basepath` prop in the router so it may cause some issues.

karelhala: Good point! Updated.

Hyperkid123 (approved):

PR #1686, opened 2021-11-09
Chrome local development

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16882

### Changes
- added `dev` script to enable local chrome development without using host apps

Please check it out locally.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -4,7 +4,7 @@ const plugins = require('./webpack.plugins.js');
 const TerserPlugin = require('terser-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const { createJoinFunction, createJoinImplementation, asGenerator, defaultJoinGenerator } = require('resolve-url-loader');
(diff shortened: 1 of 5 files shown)


config/webpack.config.js
```diff
@@ -99,14 +100,32 @@ const commonConfig = ({ publicPath, noHash }) => ({
...
+    },
+    historyApiFallback: {
+      index: `${publicPath}index.html`,
+    },
+    https: true,
+    port: 1337,
+    ...proxy({
+      env: 'stage-beta',
+      port: 1337,
+      appUrl: [/^\/*$/, /^\/beta\/*$/],
```
fhlavac, line 118: image
the `appUrl` makes the terminal output a bit weird, does it accept the regular expressions?

fhlavac (approved):
Tested it locally and it works! Just one note

config/webpack.config.js
Hyperkid123: Yeah, that is an unfortunate output. I am not sure if there is a better way of representing regexp.

PR #1700, opened 2021-11-25
Update ErrorComponent.js

cben:
Sufficiently old browsers can fail to load JS and show this screen.  
I don't believe users actively seek out documentation on which browsers we support, so I want some hint here that an old browser might be the problem.

Example seen on console of ancient Chrome 77 choking on `?.` syntax in apps/openshift:

    Uncaught SyntaxError: Unexpected token '.'
    ChunkLoadError: Loading chunk 677 failed.

### DIFF
--- a/src/js/App/ErrorComponent/ErrorComponent.js
+++ b/src/js/App/ErrorComponent/ErrorComponent.js
@@ -47,7 +47,7 @@ const ErrorComponent = (props) => {
             <a href="https://status.redhat.com/" target="_blank" rel="noreferrer">
               status page
             </a>{' '}
-            for known outages.
+            for known outages. If your browser is old, try updating to a recent version.
           </p>
           <Flex alignContent={{ default: 'alignContentCenter' }} direction={{ default: 'column' }}>
             <FlexItem>


Hyperkid123:
> Example seen on console of ancient Chrome 77 choking on ?. syntax in apps/openshift:

@cben this has nothing to do with browser support.

This means that the code is probably not transpiled to ES5 if a syntax like this makes it to the browser. I would suggest revisiting the config in openshift application.

Hyperkid123:
@cben all applications in CRC should ES5. If you were not doing that I would please ask you to do that.

We don't really want to deal with browser versioning and we would like to support as many browsers as possible to include as many users as possible.

cben:
ack, thanks.  what babel config should we use?

currently we have
```
browsers: [
        'last 2 firefox versions',
        'last 2 chrome versions',
        'last 2 edge versions',
        'last 2 safari versions',
      ],
```
which I agree takes the official stance too literally.

I see this repo .babelrc has `"targets": "> 0.25%, not dead"`.  Does that guarantee ES5?

Hyperkid123:
You should use the env preset https://babeljs.io/docs/en/babel-preset-env

PR #1701, opened 2021-11-26
Change how the dynamic nav partial is calculated to use hooks

karelhala:
### Dynamic nav will be hook

Since we want to tap into React's lifecyle we have to change how dynamic nav is calculated. This PR changes this calculation to use hooks. This somewhat simplifies the calculation as we are splitting the loading of dynamic module into wrapper component, which once mounted will use the exported hook.


src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -1,57 +1,58 @@
...
+      }));
+      if (!isEqual(newValue, currNav)) {
+        const currNavIndex = schema.navItems.findIndex((item) => item.dynamicNav === dynamicNav);
+        dispatch(
+          loadLeftNavSegment(
+            {
+              ...schema,
+              navItems: mergeArrays(
+                schema.navItems.filter((item) => !(item.dynamicNav && item.dynamicNav === dynamicNav)),
+                currNavIndex,
```
andrewballantyne, line 39: If `currNavIndex` is `-1`, your array splicing is going to get weird. Did you want to safeguard against the unable to find `findIndex` call above?

andrewballantyne (changes requested):
As asked on the other PR... how do I test this code with that code?

https://github.com/openshift/hac-core/pull/7#issuecomment-986837592

src/js/App/Sidenav/Navigation/DynamicNav.js
karelhala: Yeah, good point. It shouldn't happen since we took the dynamic nav from the navigation itself. Let's be defensive and safeguard it!

andrewballantyne (commented):
Looks good to me... I'd like to test it with the other PR though. Just need to know how to setup things so I can get the full e2e flow tested

src/js/App/Sidenav/Navigation/DynamicNav.js
```diff
@@ -1,57 +1,60 @@
...
-                  ...schema,
-                  navItems: schema.navItems.flatMap((item, key) => (key === indexOfDynamicNav ? newValue : item)),
-                },
-                currentNamespace,
-                pathname,
-                true
-              )
-            );
-          });
+    if (newNav?.length > 0) {
```
andrewballantyne, line 25: So Navigation can still return an object... why do we only want arrays here?

karelhala: Ah, yeah! Good point. It should check just for the `newNav` not if it's an array.

andrewballantyne (approved):
/lgtm

Hyperkid123 (approved):

PR #1707, opened 2021-12-13
Use webpack to build all assets

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15878

In order to merge this PR, we first need the single HTML template to be applied in production.

### Changes
- build all assets trough webpack
- remove PUG partials
- remove many scripts which have no longer any use
- remove unused sass files and utilities (were used in pug templates)
- remove PF images from build


karelhala (approved):

Hyperkid123:
@karelhala I have updated the config and tested it in UI. Can you take another look? I would like to move it in stage ASAP

PR #1742, opened 2022-01-26
Remove replace.sh and add option to dynamically load SSO url from JS file

karelhala:
### Remove the window.SSO

Since we don't want to expose publis SSO url to any user and since we don't want to do any complicated image build with post run script execution let's replace the injection of SSO with dynamic import of JS file.

Let's turn off the hash for sso-url file, we expect this file won't change (at least the content of it) that often.

### Return callback from `getOfflineToken`

Since we are waiting for the calculation of URL now, let's return a callback that does the actual request from `getOfflineToken`

### DIFF
--- a/src/js/jwt/insights/offline.js
+++ b/src/js/jwt/insights/offline.js
@@ -36,7 +36,7 @@ export function wipePostbackParamsThatAreNotForUs() {
   }
 }
 
-export function getOfflineToken(realm, clientId) {
+export async function getOfflineToken(realm, clientId) {
   const postbackUrl = getPostbackUrl();
 
   if (priv.response) {
@@ -50,10 +50,12 @@ export function getOfflineToken(realm, clientId) {
     return Promise.reject('not available');
   }
 
-  return Promise.resolve(insightsUrl(DEFAULT_ROUTES)).then((ssoUrl) => {
-    const tokenURL = `${ssoUrl}/realms/${realm}/protocol/openid-connect/token`;
-    const params = parseHashString(postbackUrl);
+  const ssoUrl = await insightsUrl(DEFAULT_ROUTES);
 
+  const tokenURL = `${ssoUrl}/realms/${realm}/protocol/openid-connect/token`;
+  const params = parseHashString(postbackUrl);
+
+  return () => {
     return axios({
       method: 'post',
       url: tokenURL,
@@ -63,7 +65,7 @@ export function getOfflineToken(realm, clientId) {
       priv.response = response;
       return response;
     });
-  });
+  };
 }
 
 function getWindow() {
--- a/src/js/jwt/insights/offline.js
+++ b/src/js/jwt/insights/offline.js
@@ -55,17 +55,14 @@ export async function getOfflineToken(realm, clientId) {
   const tokenURL = `${ssoUrl}/realms/${realm}/protocol/openid-connect/token`;
   const params = parseHashString(postbackUrl);
 
-  return () => {
-    return axios({
-      method: 'post',
-      url: tokenURL,
-      data: getPostDataString(getPostDataObject(postbackUrl, clientId, params.code)),
-      config: { headers: { 'Content-Type': 'application/x-www-form-urlencoded' } },
-    }).then((response) => {
+  return axios
+    .post(tokenURL, getPostDataObject(postbackUrl, clientId, params.code), {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+    })
+    .then((response) => {
       priv.response = response;
       return response;
     });
-  };
 }
 
 function getWindow() {
@@ -97,12 +94,4 @@ function parseHashString(str) {
       result[parts[0]] = parts[1];
       return result;
     }, {});
-}
-
-function getPostDataString(obj) {
-  return Object.entries(obj)
-    .map((entry) => {
-      return `${entry[0]}=${entry[1]}`;
-    })
-    .join('&');
 }
--- a/src/js/jwt/insights/offline.js
+++ b/src/js/jwt/insights/offline.js
@@ -50,17 +50,19 @@ export function getOfflineToken(realm, clientId) {
     return Promise.reject('not available');
   }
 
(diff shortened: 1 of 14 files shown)


src/js/jwt/insights/offline.js
```diff
@@ -50,18 +50,22 @@ export function getOfflineToken(realm, clientId) {
...
-  return axios({
-    method: 'post',
-    url: tokenURL,
-    data: getPostDataString(getPostDataObject(postbackUrl, clientId, params.code)),
-    config: { headers: { 'Content-Type': 'application/x-www-form-urlencoded' } },
-  }).then((response) => {
-    priv.response = response;
-    return response;
-  });
+  return () => {
```
Hyperkid123, line 58: Are you sure about the double `return`?

karelhala: Ah, yeah! We don't have to have function body, we can just return arrow function.

Hyperkid123: I am more curious about why we now return a function instead of a promise directly.

src/js/jwt/jwt.unit.test.js
```diff
@@ -296,7 +296,7 @@ describe('JWT', () => {
         });
       });
 
-      describe('token update passes', () => {
+      xdescribe('token update passes', () => {
```
karelhala, line 299:
```suggestion
      describe('token update passes', () => {
```

src/js/jwt/insights/offline.js
karelhala: Previously it was just a function that triggered async call, now it's an async function that returns a function that triggers promise. The biggest problem I had was with tests tbh. I was trying to wait for the function to finish, but since the axios never really finishes the tests were failing. This was the problematic part https://github.com/RedHatInsights/insights-chrome/pull/1742/files#diff-c3e05911df83f21e072e4d2ec2eb95fe152cf7fde3152926822a14fd0f33e558R44. Without the init to finish the axios is never fired.

There are no changes to the public API, this returned promise as well, now it returns promise as well.

Hyperkid123: Hmm. You could mock the whole `axios` module and its implementation in the tests to fix it. I don't necessarily like doing extra magic just to fix the tests. I can easily confuse anyone who is reading the code and does not have the context.

src/js/jwt/jwt.unit.test.js
```diff
@@ -264,7 +264,7 @@ describe('JWT', () => {
         });
       });
 
-      describe('token update fails', () => {
+      xdescribe('token update fails', () => {
```
Hyperkid123, line 267: typo?

karelhala: More like forgotten `xdescribe`. Thank you!

Hyperkid123 (approved):

src/js/jwt/insights/offline.js
```diff
@@ -50,18 +50,19 @@ export function getOfflineToken(realm, clientId) {
...
-    method: 'post',
-    url: tokenURL,
-    data: getPostDataString(getPostDataObject(postbackUrl, clientId, params.code)),
-    config: { headers: { 'Content-Type': 'application/x-www-form-urlencoded' } },
-  }).then((response) => {
-    priv.response = response;
-    return response;
-  });
+  return axios
+    .post(tokenURL, getPostDataObject(postbackUrl, clientId, params.code), {
```
cben, line 59: @karelhala Looks like axios.post() with object doesn't do the right thing for form-urlencoded, it sends JSON.  
I'm seeing errors from SSO for `insights.chrome.auth.getOfflineToken()` in [qaprodauth](https://qaprodauth.console.redhat.com/openshift/token/show), [qaprodauth beta](https://qaprodauth.console.redhat.com/beta/openshift/token/show), and [prod-beta](https://console.redhat.com/beta/openshift/token/show);
[prod-stable](https://console.redhat.com/openshift/token/show) works.  

old working request has payload looking like `code=...&grant_type=authorization_code&client_id=cloud-services&redirect_uri=https%3A%2F%[2Fconsole.redhat.com](http://2Fconsole.redhat.com)%2Fopenshift%2Ftoken%2Fshow%3Fnoauth%3D2402500adeacc30eb5c5a8a5e2e0ec1f`.
new failing request has JSON payload `{"code":"...","grant_type":"authorization_code","client_id":"cloud-services","redirect_uri":"https%3A%2F%[2Fconsole.redhat.com](http://2Fconsole.redhat.com)%2Fbeta%2Fopenshift%2Ftoken%2Fshow%3Fnoauth%3D2402500adeacc30eb5c5a8a5e2e0ec1f"}`
This causes server to respond 400 Bad Request, body `{"error":"invalid_request","error_description":"Missing form parameter: grant_type"}` plus omit Access-Control-Allow-Origin header so what browser sees is CORS error.

Per https://github.com/axios/axios#using-applicationx-www-form-urlencoded-format, passing an object is not supposed to work, see there for alternatives (or bring back `getPostDataString()`).

Hyperkid123: @cben  This change was not pushed to prod-stable yet. @karelhala can you please check this and fix the issue?

karelhala: Sure thing!

karelhala: @cben thank you for the investigation! I've overlooked the form url encoded format and just blindly removed the encoding part. This PR brings back the encoding function and sends correct data https://github.com/RedHatInsights/insights-chrome/pull/1759

cben: Thanks.
- [x] https://qaprodauth.console.redhat.com/beta/openshift/token/show now works again (and we haven't pushed any app changes there, so confirms the insights-chrome fix :+1:)
- [ ] https://qaprodauth.console.redhat.com/openshift/token/show still sends JSON.
- [ ] https://console.redhat.com/beta/openshift/token/show still sends JSON.  I assume those just weren't deployed yet.

Hyperkid123: We will promote the changes to other envs soon.

PR #1743, opened 2022-01-26
[HAC-509] add FE template, turn off Jest cache

BlakeHolifield:
* Turn off cache for Jest testing (breaks on file permissions currently)
* Add FE template for bonfire to deploy to eph environments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "fixjs": "eslint src --fix",
     "lint": "npm-run-all lint:*",
     "lint:js": "eslint src",
-    "test": "jest --forceExit --detectOpenHandles",
+    "test": "jest --forceExit --detectOpenHandles --no-cache",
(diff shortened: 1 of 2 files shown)


karelhala (approved):

package.json
```diff
@@ -26,7 +26,7 @@
     "fixjs": "eslint src --fix",
     "lint": "npm-run-all lint:*",
     "lint:js": "eslint src",
-    "test": "jest --forceExit --detectOpenHandles",
+    "test": "jest --forceExit --detectOpenHandles --no-cache",
```
Hyperkid123, line 29: Can we create a new command just for the CI? This will significantly increase the duration of the test during local development.

karelhala: Good point, created new issue for it - https://github.com/RedHatInsights/insights-chrome/issues/1745

PR #1763, opened 2022-02-07
rescope classes & general cleanup #2

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-17881


src/sass/chrome.scss
```diff
@@ -47,145 +41,67 @@ aside {
...
-            .pf-c-select__menu li button:hover label .pf-c-badge.pf-m-read {
-                background: var(--pf-global--Color--light-100);
-            }
-
-            .ins-c-tagfilter__option-value {
-                max-width: 360px;
-            }
-        }
-    }
+  --pf-c-page__sidebar--Transition: all 500ms cubic-bezier(.42, 0, .58, 1);
```
karelhala, line 44: Mhh, I know the timing is a bit different, but do you think we could use `--pf-global--TimingFunction` for the animation function? At least we'll be aligned with PF.

karelhala (commented):
Looking good! Just a few minor changes and we are golden.

src/sass/chrome.scss
epwinchell: @karelhala Actually, we didn't need that line at all. We were overriding the default timing.

karelhala (approved):

src/sass/chrome.scss
```diff
@@ -47,145 +41,66 @@ aside {
 
 // Page temp styles
 .pf-c-page {
-    --pf-c-page__sidebar--Transition: all 500ms cubic-bezier(.42, 0, .58, 1);
-    --pf-m-emphasis--FontSize: var(--pf-global--FontSize--xl);
-
-    .ins-c-chrome__global-filter {
-        padding: var(--pf-global--spacer--md) var(--pf-global--spacer--lg);
-        background: var(--pf-global--BackgroundColor--100);
```
Hyperkid123, line 55: @epwinchell the white background is missing from the global filter:

Screenshot from 2022-02-14 11-49-53

The issue is present in stage beta.

PR #1774, opened 2022-02-15
Add unleash feature flags provider.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15343

### Changes
- create FeatureFlagsProvider wrapper
- mark @unleash/proxy-client-react as a singleton dependency


src/js/App/FeatureFlags/FeatureFlagsProvider.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import UnleasFlagProvider from '@unleash/proxy-client-react';
+
+const config = {
+  url: `${document.location.origin}/api/featureflags/v0`,
+  clientKey: 'proxy-123',
```
karelhala, line 7: Shouldn't the `clientKey` be changable? Otherwise it looks like a magic value.

src/js/App/FeatureFlags/FeatureFlagsProvider.js
```diff
@@ -0,0 +1,22 @@
...
+  appName: 'web',
+  headerName: 'X-Unleash-Auth',
+  refreshInterval: 60000,
+  metrcisInterval: 120000,
+};
+
+const FeatureFlagsProvider = ({ children }) => {
+  return <UnleasFlagProvider config={config}>{children}</UnleasFlagProvider>;
+};
```
karelhala, line 16: Extra return
```suggestion
const FeatureFlagsProvider = ({ children }) => <UnleasFlagProvider config={config}>{children}</UnleasFlagProvider>;
```

src/js/App/RootApp/RootApp.js
```diff
@@ -2,15 +2,18 @@ import React from 'react';
...
 import IDPChecker from './IDPChecker';
 import ScalprumRoot from './ScalprumRoot';
 
 const RootApp = (props) => (
   <Router history={chromeHistory} basename={isBeta() ? '/beta' : '/'}>
-    <IDPChecker>
-      {/* <CrossRequestNotifier /> */}
-      <ScalprumRoot {...props} />
-    </IDPChecker>
+    <FeatureFlagsProvider>
```
karelhala, line 11: What's the benefit of having a extra component? We could directly import `UnleasFlagProvider` and have the `config` value be in constant file. Unless there's expected more props to be passed to this component in future (className, conditional tun on/off, etc.) I would rather keep it simple.

karelhala (commented):
Looking good, I was kinda suprised with the amount of changes in `package-lock.json` but I guess unleash brings in a lot of dependencies.

src/js/App/FeatureFlags/FeatureFlagsProvider.js
Hyperkid123: We have a custom unleash proxy backed that is not using the key, but the client requires it. I think eventually the client key will be our user number. But we are not there yet.

src/js/App/RootApp/RootApp.js
Hyperkid123: I expect that more configuration will be required in a near future.

Hyperkid123: I also want to keep the RootApp as simple as possible. We don't want file to have 500 lines like it used to have in chrome 1

karelhala: Then it would be better to have a component called `ChromeWrapper` with `UnleasFlagProvider`, `IDPChecker` and `ScalprumRoot`.

That way `RootApp` would set the Router and `ChromeWrapper` the other providers.

karelhala (approved):

src/js/App/RootApp/RootApp.js
Hyperkid123: The `RootApp` is meant to be such a wrapper.

PR #1778, opened 2022-02-17
Render feedback button in a portal

fhlavac:
Made feedback button render in a ReactDOM portal to prevent it from interrupting in DOM. It caused unwanted rendering in the top navbar in some apps (eg. Registration) - see image before.

Before:
beforefeedback

Now:
afterfeedback

cc @Hyperkid123

### DIFF
--- a/src/js/App/Feedback/Feedback.scss
+++ b/src/js/App/Feedback/Feedback.scss
@@ -1,4 +1,4 @@
-.chr-c-button-feedback {
+button.chr-c-button-feedback {
   svg { margin-right: var(--pf-global--spacer--sm); }
   position: fixed;
   z-index: 19000;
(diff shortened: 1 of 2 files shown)


karelhala (approved):

Hyperkid123:
@fhlavac @karelhala I think the color of the button should not be blue. You can check prod. I think we are missing some styles.

@epwinchell I can also see that the button is white in pre-prod envs. Is that a planned change? Or is this a regression?

fhlavac:
@Hyperkid123 ah, good catch, some stylings are really not being applied. I'll fix that in a bit

fhlavac:
@Hyperkid123 
This should help: https://github.com/RedHatInsights/insights-chrome/pull/1779

PR #1781, opened 2022-02-22
Fix missing quickstarts state object.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-18159

In non-stage environment, the quickstarts state object is not initialized properly and can cause a runtime crash when opening the quickstart drawer.

### DIFF
--- a/src/js/App/QuickStart/useQuickstartsStates.js
+++ b/src/js/App/QuickStart/useQuickstartsStates.js
@@ -7,7 +7,7 @@ import { getEnv } from '../../utils';
 const isStage = getEnv() === 'stage';
 
 const statePersistor = isStage ? useState : useLocalStorage;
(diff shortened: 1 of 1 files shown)


karelhala (approved):

src/js/App/QuickStart/useQuickstartsStates.js
```diff
@@ -7,7 +7,7 @@ import { getEnv } from '../../utils';
 const isStage = getEnv() === 'stage';
 
 const statePersistor = isStage ? useState : useLocalStorage;
-const initiStatesArgs = isStage ? ['insights-quickstarts', {}] : [{}];
+const initiStatesArgs = isStage ? ['insights-quickstarts', {}] : ['insights-quickstarts', {}];
```
aneelac22, line 10: Is this even needed, looks like it is set to the same arg either way

arivepr (approved):

PR #1783, opened 2022-02-23
Update AAP filter usage

kruai:
The current AAP filter does not work, as it will only fetch hosts which have _all_ of ansible's child fields populated
On the API side, we're adding the ability to filter on whether a parent field has any child field with a value. This PR should not be merged until the API change is ready.

[This PR](https://github.com/RedHatInsights/insights-inventory-frontend/pull/1502) updates the implementation on the HBI front-end.

### DIFF
--- a/src/js/App/GlobalFilter/tagsApi.js
+++ b/src/js/App/GlobalFilter/tagsApi.js
@@ -28,12 +28,7 @@ const buildFilter = (workloads, SID) => ({
     ...(workloads?.SAP?.isSelected && { sap_system: true }),
     // enable once AAP filter is enabled
     ...(workloads?.[AAP_KEY]?.isSelected && {
-      ansible: {
-        controller_version: 'not_nil',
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@kruai I have changed the PR to draft so we don't accidentally hit the button.

kruai:
@karelhala @Hyperkid123 This is now ready for review! The API PR got merged and is now live in Stage.

ryandillinfelton (approved):

Hyperkid123:
@kruai when can we expect it in production?

kruai:
> @kruai when can we expect it in production?

We're going to try to promote the HBI API change to Prod on Monday.

Hyperkid123:
> > @kruai when can we expect it in production?
> 
> We're going to try to promote the HBI API change to Prod on Monday.

@kruai Cool, please ping me on slack so we can add this commit to prod as well.

PR #1795, opened 2022-03-02
Enable typescript build.

Hyperkid123:
### Changes
- use TS loader to enable typescript code to be loaded
- eagerly load all shared modules (seems to be required if we want to use ts-loader)
- move all initial chrome code to the `bootsrap.js` files

### TODO
- [x] configure build
- [x] check test
- [x] check linter
- [x] migrate simple file to TS to verify the build output


fhlavac (approved):
Tested locally, looks good to me!

PR #1798, opened 2022-03-03
TS Migration

Hyperkid123:
requires: https://github.com/RedHatInsights/insights-chrome/pull/1797


src/js/sentry.ts
```diff
@@ -1,4 +1,5 @@
 import * as Sentry from '@sentry/browser';
+import { ChromeUser } from './types';
```
karelhala, line 2: Could be better to indicate we're importing `type`. Based on the doc, `import type` will be completely removed once transpiled.
```suggestion
import type { ChromeUser } from './types';
```

src/js/types.d.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export type VisibilityFunctions = {
+  isOrgAdmin: () => Promise<boolean>;
+  isActive: () => Promise<boolean>;
+  isInternal: () => Promise<boolean>;
+  isEntitled: () => Promise<boolean>;
+  isProd: () => boolean;
+  isBeta: () => boolean;
+  isHidden: () => true;
+  withemai: (toHave: string) => Promise<boolean>;
```
karelhala, line 55: This should be `withEmail` and `toHave` should be unknon amount of attrs.
```suggestion
  withEmail: (...toHave: string[]) => Promise<boolean>;
```

src/js/utils.ts
```diff
@@ -112,9 +113,9 @@ export function lastActive(searchString, fallback) {
   }, fallback);
 }
 
-export const isAnsible = (sections) => sections.includes('ansible') && sections.includes('insights');
+export const isAnsible = (sections: string[]) => (sections.includes('ansible') && sections.includes('insights') ? 1 : 0);
```
karelhala, line 116: Any reason to switch from `boolean` to `1`|`0`?

src/js/sentry.ts
Hyperkid123: We now import it form the types package

src/js/types.d.ts
Hyperkid123: Type is now imported from the types package.

karelhala (approved):

PR #1802, opened 2022-03-07
fix-eager-loading-modules-issues

Hyperkid123:
### Changes
- keep babel-loader in addition to ts-loader, this probably fixes the problem with the lazy boostrap import that does not seem to be working for ts-loader for some reason


config/webpack.config.js
```diff
@@ -51,6 +51,11 @@ const commonConfig = ({ dev, publicPath = '/', noHash }) => ({
   },
   module: {
     rules: [
+      {
+        test: /\.jsx?$/,
+        loader: 'babel-loader',
+        exclude: /node_modules/,
+      },
       {
         test: /\.(jsx?|tsx?)$/,
```
karelhala, line 60: Shouldn't here be just `tsx`?

Hyperkid123: yes

karelhala (approved):

PR #1816, opened 2022-03-10
Updates redhat logo link based on env

aneelac22:
[RHCLOUD-18223]


karelhala (commented):
This is caused by the new index.html template, this will fix the immediate problem with logo clicks. However there are other pages that rely on document root. We should use [html-replace-webpack-plugin](https://www.npmjs.com/package/html-replace-webpack-plugin) and in index.html replace the `base href="/"`. There's also `head-block.pug` which we should use similiar approach on it.

Hyperkid123:
@karelhala even better we should use the react-router routing that should solve any issues with beta envs and we will have nice client-side routing.

aneelac22:
@Hyperkid123 Not sure about this approach, I still have to check for isBeta in ChromeLink

Hyperkid123:
> @Hyperkid123 Not sure about this approach, I still have to check for isBeta in ChromeLink

You don't have. The link uses the basename of the router. Which is `/beta` in beta enviroments. You do have to add the `appId` though.

PR #1827, opened 2022-03-21
Enable global filter on all AAP screens

karelhala:


Hyperkid123 (commented):
@karelhala can you fix up the tests?

Hyperkid123 (approved):

ZitaNemeckova:
@karelhala This affects Insights for Ansible and Hub as well. Could you please fix it to make it only AAP? 

cc: @himdel

karelhala:
@ZitaNemeckova this was requested by Jerome Marc in RHCLOUD-18330 we've stressed many times that it'll surface global filter on all Ansible pages. The thing is that when we were working with UX on this (initially just for insights bundle) we identified that users were confused where the global filter went if it wasn't enabled for some pages in bundle. So the only thing app can do to hide the global filter once it's deployed to a bundle is to disable it.

PR #1840, opened 2022-04-07
Adding docs button to about menu.

arivepr:


src/js/App/Header/Tools.js
```diff
@@ -92,6 +93,11 @@ const Tools = () => {
     });
   }, []);
 
+  useEffect(() => {
+    const section = getSection();
+    section === 'insights' ? setIsRhel(true) : setIsRhel(false);
+  }, []);
```
Hyperkid123, line 99: This is way too complicated and also will not update if you switch bundles. I would just use the condition in the `aboutMenuDropdownItems` and check on each render.

```jsx
  const aboutMenuDropdownItems = [
    {
	@@ -111,6 +117,11 @@ const Tools = () => {
      title: 'Status page',
      url: 'https://status.redhat.com/',
    },
    {
      title: 'Insights for RHEL Documentation',
      url: `https://access.redhat.com/documentation/en-us/red_hat_insights/2021`,
      isHidden: getSection() !== 'insights',
    },
```

arivepr: Nice fat trimming!

Adding now.

Hyperkid123 (approved):

PR #1854, opened 2022-04-20
Adding target_org cookie.

arivepr:
Adding target_org cookie to context switcher!

### DIFF
--- a/src/js/App/Header/ContextSwitcher.js
+++ b/src/js/App/Header/ContextSwitcher.js
@@ -17,7 +17,7 @@ import {
   CROSS_ACCESS_ACCOUNT_NUMBER,
   REQUESTS_COUNT,
   REQUESTS_DATA,
-  TARGET_ORG,
+  CROSS_ACCESS_ORG_ID,
 } from '../../consts';
 
 const ContextSwitcher = ({ user, className }) => {
@@ -31,6 +31,7 @@ const ContextSwitcher = ({ user, className }) => {
   };
 
   const handleItemClick = (target_account, request_id, end_date) => {
+    console.log('OMG HIIIIIIIIIIII');
     if (target_account === selectedAccountNumber) {
       return;
     }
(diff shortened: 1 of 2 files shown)


arivepr:
Let me know if this works for what you needed @Hyperkid123 !

src/js/App/Header/ContextSwitcher.js
```diff
@@ -32,6 +39,8 @@ const ContextSwitcher = ({ user, className }) => {
     localStorage.removeItem(REQUESTS_DATA);
     setSelectedAccountNumber(target_account);
     Cookies.set(CROSS_ACCESS_ACCOUNT_NUMBER, target_account);
+    Cookies.set(TARGET_ORG, target_account);
```
coderbydesign, line 42: Hey @arivepr! This will need to be the org id value, not the account number. The idea is that RBAC will need both the account number in the `cross_access_account_number` cookie, and the org id in the `cross_access_org_id` cookie.

arivepr: Hey Keith @coderbydesign , thanks for the clarification!

Did not realize they were two seperate cookies.
If you don't mind, does that look good for the crossid? Or is it a naming issue, given that target_org === cross_id? and then add the missing cookie/constant for cross-act-num

coderbydesign: Not sure I follow completely, but since these two cookies should be named similarly I would think. So where you have `CROSS_ACCESS_ACCOUNT_NUMBER` currently, I'd think it would make sense to have `CROSS_ACCESS_ORG_ID` instead of `TARGET_ORG`, but you'd supply it the value `target_org`, from the actual cross account request object?

The actual cookie name of `cross_access_org_id` is perfect, though! That's what we're anticipating be sent.

Hyperkid123:
@arivepr the changes look good. I will test the flow in the UI to be absolutely sure.

Hyperkid123 (approved):
Tested in UI. Cookie is set in requests. :+1:

PR #1877, opened 2022-05-02
Navigation feature flags

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-19079

### DIFF
--- a/src/js/consts.js
+++ b/src/js/consts.js
@@ -35,8 +35,6 @@ const matchValue = (value, matcher) => {
   return typeof match === 'function' ? match(value) : value;
 };
 
-const checkFeatureFLag = (flagName, expectedValue) => getFeatureFlagsError() !== true && unleashClient.isEnabled(flagName) === expectedValue;
-
 /**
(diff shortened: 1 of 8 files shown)


src/js/consts.js
```diff
@@ -112,6 +115,7 @@ export const visibilityFunctions = {
       return false;
     }
   },
+  featureFlag: checkFeatureFLag,
```
karelhala, line 118: Any reason not to define it here, but rather on line 38?

karelhala (commented):
Looking good, just one minor nit pick.

src/js/consts.js
Hyperkid123: I just don't like anonymous functions that's it.

Gotta figure out the test though

karelhala (approved):

Hyperkid123:
Converting to draft so we don't accordantly include it in the summit build

karelhala:
@Hyperkid123 I think we can mark it as redy for review.

karelhala (approved):

PR #1894, opened 2022-05-09
make jira tag for feedback more descriptive

InsaneZein:
for this jira: [RHCLOUD-18880](https://issues.redhat.com/browse/RHCLOUD-18880)

### DIFF
--- a/src/js/App/Feedback/Feedback.js
+++ b/src/js/App/Feedback/Feedback.js
@@ -20,7 +20,7 @@ const Feedback = ({ user }) => {
 
   const addFeedbackTag = () => {
     if (window.insights.chrome.isProd) {
-      return bundle;
+      return `[${bundle}]`;
     } else {
       return '[PRE-PROD]';
     }
--- a/src/js/App/Feedback/Feedback.js
+++ b/src/js/App/Feedback/Feedback.js
@@ -17,14 +17,7 @@ const Feedback = ({ user }) => {
   const bundle = window.insights.chrome.getBundle();
   const isAvailable = env === 'prod' || env === 'stage';
   const setIsModalOpen = (...args) => dispatch(toggleFeedbackModal(...args));
-
-  const addFeedbackTag = () => {
-    if (window.insights.chrome.isProd) {
-      return `[${bundle}]`;
-    } else {
-      return '[PRE-PROD]';
-    }
-  };
+  const addFeedbackTag = () => window.insights.chrome.isProd ? `[${bundle}]` : '[PRE-PROD]';
 
   const handleModalSubmission = () => {
     if (isAvailable) {
--- a/src/js/App/Feedback/Feedback.js
+++ b/src/js/App/Feedback/Feedback.js
@@ -26,7 +26,7 @@ const Feedback = ({ user }) => {
     } else {
       return '[PRE-PROD]';
     }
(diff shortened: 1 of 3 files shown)


src/js/App/Feedback/Feedback.js
```diff
@@ -29,7 +37,7 @@ const Feedback = ({ user }) => {
         },
         body: JSON.stringify({
           description: `Feedback: ${textAreaValue}, Username: ${user.identity.user.username}, Account ID: ${user.identity.account_number}, Email: ${user.identity.user.email}, URL: ${window.location.href}`, //eslint-disable-line
-          summary: `${window.insights.chrome.isProd ? '[PROD]' : '[PRE-PROD]'} App Feedback`,
+          summary: `${addFeedbackTag()} App Feedback`,
```
ryelo, line 40: can we put the brackets around this so it comes up as `[Insights] App Feedback`?

ryelo (changes requested):

src/js/App/Feedback/Feedback.js
InsaneZein: oh good catch

src/js/App/Feedback/Feedback.js
```diff
@@ -18,6 +18,14 @@ const Feedback = ({ user }) => {
   const isAvailable = env === 'prod' || env === 'stage';
   const setIsModalOpen = (...args) => dispatch(toggleFeedbackModal(...args));
 
+  const addFeedbackTag = () => {
```
Hyperkid123, line 21: Lets make it a bit less wordy: 
```js
const addFeedbackTag = () => window.insights.chrome.isProd ? `[${bundle}]` : '[PRE-PROD]'
```

src/js/App/Feedback/Feedback.js
```diff
@@ -18,6 +18,14 @@ const Feedback = ({ user }) => {
...
+    if (window.insights.chrome.isProd) {
```
Hyperkid123, line 22: Can you create a helper function that will give you the same value? We want to stop using the window API.

You can check this file to get the implementation: https://github.com/RedHatInsights/insights-chrome/blob/master/src/js/chrome/entry.js#L112

InsaneZein: Hmm I'm a bit confused by this. Looking at the implementation of that file, getApp is also using the window API.

Hyperkid123: Yup, there is a plenty of parts that are using the window. We want to migrate away from it.

InsaneZein: I've changed the implementation. Let me know if this is what you meant.

Hyperkid123 (approved):

src/js/App/Feedback/Feedback.js
```diff
@@ -6,6 +6,7 @@ import Cookies from 'js-cookie';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
 import { toggleFeedbackModal } from '../../redux/actions';
+import { isProd } from '../utils.ts';
```
Hyperkid123, line 9: This should be 
```js
import { isProd } from '../../utils.ts';
```

PR #1895, opened 2022-05-10
[POC] WS browser events

Hyperkid123:
### Description

This is a POC of the upcoming dynamic browser events service (web sockets in short)

It adds two main features
1. Notification drawer. That is something we will have to support.
2. Allow application WS listeners registration. This is something that can provide a huge potential

The WS service is not published anywhere yet but it does these things

1. Emit events based on Kafka messages. Currently, it listens on `chrome-events` channel. Tapping into existing channels might be an issue due to frequency and due to the fact that the messages probably do not follow strict structures.
2. Emit events based on a POST request. Just passes along the payload.
3. Emit events based on a client event send from the browser connection.

The WS service has 3 types of granularity setup
1. Broadcast: send message to every client connected
2. Tennant based (org): send message to every connected client in a tenant
3. User ID based: only a user that matches the ID will receive a message

Each message can have multiple targets. Meaning that we can send a message to more than tenants or users or combinations.

Do not take any code seriously it will change. The Notification drawer has a high priority as it is a future requirement.

Let me know your thoughts. You can check this RBAC POC to see advanced caching and data invalidation options: https://github.com/RedHatInsights/insights-rbac-ui/pull/1139.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,7 +6,7 @@
     <title>
         console.redhat.com
     </title>
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https: wss: data: 'unsafe-inline' 'unsafe-eval' https://*.redhat.com/ https://www.redhat.com  https://*.openshift.com/ https://api.stage.openshift.com/ https://identity.api.openshift.com/ https://www.youtube.com/ https://redhat.sc.omtrdc.net/ https://assets.adobedtm.com https://www.redhat.com https://*.storage.googleapis.com/ https://*.hotjar.com:* https://*.hotjar.io wss://*.hotjar.com;">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https: wss: data: 'unsafe-inline' 'unsafe-eval' ws://localhost:8080 https://*.redhat.com/ https://www.redhat.com  https://*.openshift.com/ https://api.stage.openshift.com/ https://identity.api.openshift.com/ https://www.youtube.com/ https://redhat.sc.omtrdc.net/ https://assets.adobedtm.com https://www.redhat.com https://*.storage.googleapis.com/ https://*.hotjar.com:* https://*.hotjar.io wss://*.hotjar.com;">
     <meta http-equiv="Pragma" content="no-cache">
     <meta http-equiv="cache-control" content="no-cache, no-store, must-revalidate">
     <meta http-equiv="x-ua-compatible" content="ie=edge">
--- a/src/js/App/RootApp/RootApp.js
+++ b/src/js/App/RootApp/RootApp.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { Router } from 'react-router-dom';
 import { HelpTopicContainer, QuickStartContainer } from '@patternfly/quickstarts';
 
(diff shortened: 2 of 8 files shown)


src/index.html
```diff
@@ -6,7 +6,7 @@
     <title>
         console.redhat.com
     </title>
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https: wss: data: 'unsafe-inline' 'unsafe-eval' https://*.redhat.com/ https://www.redhat.com  https://*.openshift.com/ https://api.stage.openshift.com/ https://identity.api.openshift.com/ https://www.youtube.com/ https://redhat.sc.omtrdc.net/ https://assets.adobedtm.com https://www.redhat.com https://*.storage.googleapis.com/ https://*.hotjar.com:* https://*.hotjar.io wss://*.hotjar.com;">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' https: wss: data: 'unsafe-inline' 'unsafe-eval' ws://localhost:8080 https://*.redhat.com/ https://www.redhat.com  https://*.openshift.com/ https://api.stage.openshift.com/ https://identity.api.openshift.com/ https://www.youtube.com/ https://redhat.sc.omtrdc.net/ https://assets.adobedtm.com https://www.redhat.com https://*.storage.googleapis.com/ https://*.hotjar.com:* https://*.hotjar.io wss://*.hotjar.com;">
```
karelhala, line 9: What about `wss:`? Also, shouldn't there be a `wss:///.redhat.com`?

src/js/redux/index.js
```diff
@@ -80,6 +82,8 @@ const reducers = {
   [POPULATE_QUICKSTARTS_CATALOG]: populateQuickstartsReducer,
   [DISABLE_QUICKSTARTS]: disableQuickstartsReducer,
   [UPDATE_DOCUMENT_TITLE_REDUCER]: documentTitleReducer,
+  'toggle-notifications-drawer': notificationsDrawerReducer,
+  'add-notification': addNewNotificationReducer,
```
karelhala, line 86: Can we make these into constants?

src/js/App/RootApp/RootApp.js
```diff
@@ -12,11 +12,73 @@ import useQuickstartsStates from '../QuickStart/useQuickstartsStates';
...
+        });
+    }
+    console.log(events);
+  }
+
+  /**
+   * Establish connection
+   */
+  useEffect(() => {
+    const x = new WebSocket('ws://localhost:8080/ws');
```
karelhala, line 40: I'd rather see a proxy config in webpack for websockets.

src/js/App/RootApp/RootApp.js
```diff
@@ -12,11 +12,73 @@ import useQuickstartsStates from '../QuickStart/useQuickstartsStates';
...
+  const activeModule = useSelector(({ chrome: { activeModule } }) => activeModule);
+  const dispatch = useDispatch();
+  function handleEvent(payload) {
+    if (payload.type === 'new-notification') {
+      dispatch({ type: 'add-notification', payload });
+    } else if (payload.type === 'invalidate') {
+      Object.values(events.current)
+        .flat()
+        .forEach((cb) => {
+          cb(payload);
```
karelhala, line 30: I'd check if the callback is defined in here
```suggestion
          cb?.(payload);
```

src/index.html
Hyperkid123: Don't take these seriously we I don't even know where the service will live

src/js/redux/index.js
Hyperkid123: Definitely. This is just a draft to demo the features. Do not take anything seriously, only the concept.

src/js/App/RootApp/RootApp.js
Hyperkid123: It's checked when a callback is registered. It will not register listener if a callback is not a function

src/js/App/RootApp/RootApp.js
Hyperkid123: I have to do some research on this. I am not sure how the webpack proxy works with WS. Not sure if our implementation will work properly.

Hyperkid123:
Obsolete

PR #1897, opened 2022-05-18
add ability to send activation request

ryelo:
this needs to be yeeted into prod after merge

https://issues.redhat.com/browse/RHCLOUD-19403
Screen Shot 2022-05-18 at 1 04 41 PM

### DIFF
--- a/src/js/App/Activation/Activation.js
+++ b/src/js/App/Activation/Activation.js
@@ -1,0 +1,59 @@
+import React, { useState, useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+import { Modal, ModalVariant } from '@patternfly/react-core';
+import Cookies from 'js-cookie';
+import PropTypes from 'prop-types';
+
+const Activation = ({ user }) => {
+
+  const [isModalOpen, setIsModalOpen] = useState(true)
+  const history = useHistory();
+  const env = window.insights.chrome.getEnvironment();
+  const isAvailable = env === 'prod';
+
+  const handleActivationRequest = () => {
+    if (isAvailable) {
+      fetch(`${window.origin}/api/platform-feedback/v1/issues`, {
+        method: 'POST',
+        headers: {
+          Accept: 'application/json',
+          Authorization: `Bearer ${Cookies.get('cs_jwt')}`,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          description: `Username: ${user.identity.user.username}, Account ID: ${user.identity.account_number}, Email: ${user.identity.user.email}, Send to rbernlei@redhat.com`, //eslint-disable-line
+          summary: `Activation Request - assign to rbernlei@redhat.com`,
+          labels: 'activation-request'
+        }),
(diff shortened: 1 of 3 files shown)


src/js/App/Activation/Activation.js
```diff
@@ -0,0 +1,55 @@
...
+    history.push('/');
+  };
+
+  useEffect(() => {
+    user && handleActivationRequest();
+  }, []);
+
+  return (
+    <Modal title="Thank you for submitting your activation request" isOpen={isModalOpen} variant={ModalVariant.medium} onClose={onModalClose}>
+      {' '}
```
fhlavac, line 45: Do we need this?

fhlavac (approved):

src/js/App/Header/Header.js
```diff
@@ -29,6 +30,8 @@ const FeedbackRoute = ({ user }) => {
 
 export const Header = () => {
   const user = useSelector(({ chrome }) => chrome?.user);
+  const isActivationPath = window.location.search === '?azure-openshift-activation';
```
Hyperkid123, line 33: What if we have more than one search parameter? Is that scenario possible? We might want to use the `URLSearchParams`
```js
const search = new URLSearchParams(window.location.search)
const isActivationPath = search.has('azure-openshift-activation')
```

src/js/App/Activation/Activation.js
```diff
@@ -0,0 +1,55 @@
+import React, { useState, useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+import { Modal, ModalVariant } from '@patternfly/react-core';
+import Cookies from 'js-cookie';
+import PropTypes from 'prop-types';
+
+const Activation = ({ user }) => {
+  const [isModalOpen, setIsModalOpen] = useState(true);
+  const history = useHistory();
+  const env = window.insights.chrome.getEnvironment();
```
Hyperkid123, line 10: can we use the `getEnv` function directly from `utils.js` instead of accessing it from window?

src/js/App/Activation/Activation.js
```diff
@@ -0,0 +1,55 @@
...
+          Accept: 'application/json',
+          Authorization: `Bearer ${Cookies.get('cs_jwt')}`,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          description: `Username: ${user.identity.user.username}, Account ID: ${user.identity.account_number}, Email: ${user.identity.user.email}, Send to rbernlei@redhat.com`, //eslint-disable-line
+          summary: `Activation Request - assign to rbernlei@redhat.com`,
+          labels: 'activation-request',
+        }),
+      }).then((response) => response.json());
```
Hyperkid123, line 27: We are not doing anything with the response. The then branch is not necessary.

src/js/App/Activation/Activation.js
ryelo: Lint adds it 🙁

Hyperkid123 (approved):

PR #1914, opened 2022-06-02
Update container testing env name

BlakeHolifield:
I'm adding this to both `stage` and `qa` since it's a stage deployment, but using qa? That's how it was setup so let me know if that's an issue.

### DIFF
--- a/src/js/jwt/constants.ts
+++ b/src/js/jwt/constants.ts
@@ -7,7 +7,7 @@ export const DEFAULT_ROUTES = {
     portal: 'https://access.redhat.com',
   },
   qa: {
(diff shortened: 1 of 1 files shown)


Hyperkid123:
So stage UI is using QA build assets. That is correct. But the SSO is different. Officially we no longer support QA environment. There is only qaprodauth. I am not sure how its gonna be with the operator.

BlakeHolifield:
> So stage UI is using QA build assets. That is correct. But the SSO is different. Officially we no longer support QA environment. There is only qaprodauth. I am not sure how its gonna be with the operator.

I think we're just going to use whatever chrome has setup. So stage sso was working before we got a full domain. I'll just try it with stage.

BlakeHolifield (commented):
Don't use qa

Hyperkid123 (approved):

PR #1918, opened 2022-06-07
Replace pendo tracking with segment IO

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-19413

### Changes
- add segment-next analytics module
- enable runtime API key switching
- disable pendo analytics
- user identification
- group identification


config/webpack.config.js
```diff
@@ -108,7 +110,7 @@ const commonConfig = ({ dev }) => {
       https: true,
       port: 1337,
       ...proxy({
-        env: 'stage-beta',
+        env: 'prod-beta',
```
rvsia, line 113: is this intended?

src/js/App/analytics/segment-analytics.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export type SegmentProviderProps = {
+  activeModule: string;
+};
+
+export const SegmentProvider: React.FC<SegmentProviderProps> = ({ activeModule, children }) => {
+  const analytics = useRef<AnalyticsBrowser>();
+  const user = useSelector(({ chrome: { user } }: AnyObject) => user);
+  useEffect(() => {
+    registerUrlObserver();
```
rvsia, line 76: shouldn't we `disconnect` it when the app is unmounted?

Hyperkid123: If chrome is unmounted it means the tab was closed or the site has changed. There is no point in unregistering. It will be lost in a shadow realm :ghost:

config/webpack.config.js
Hyperkid123: No

src/js/App/analytics/segment-analytics.tsx
rvsia: What about the concurrent mode? 😏

luciddreamz: Somehow Pendo in our previous setup seemed like it continued tracking AFTER someone left our domain. We had pageview data coming in from hundreds/thousands of other domains. A few of them could be explained by our Pendo config/keys getting copy/pasted and used on other websites, but there were so many that it seemed like the impossible was happening and our tracking was continuing to run in the browser after someone left our site. Just throwing that out there in case the shadow realm isn't shadowy enough (maybe some browsers have a bug or maybe Pendo's interface was incorrectly telling us what domain data had been captured from).

Hyperkid123: Hmm, interesting. The only thing I can think of would be some weird data in local storage. This is a simple DOM observer triggered on a DOM change that cannot exist after the tree is destroyed. I will unregister it though just to be sure.

rvsia (approved):

src/js/unfetch/index.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * This is required for the segment analytics
+ * It uses the unfetch library which is not compatbile with our fecth library and fails to initialize the
```
karelhala, line 3: A typo, missing what is failing to initialize.

src/js/App/analytics/segment-analytics.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const bodyList = document.body;
+  const observer = new MutationObserver(function (mutations) {
+    mutations.forEach(function () {
+      const newLocation = document.location.href.replace(/#.*$/, '');
+      if (oldHref !== newLocation) {
+        oldHref = newLocation;
+        window?.sendCustomEvent('pageBottom');
+        setTimeout(() => {
+          window.segment?.page(window?._segment?.pageOptions);
```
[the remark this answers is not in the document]
Hyperkid123: Not here. If the segment is defined page is defined it cannot be empty.

src/js/unfetch/index.js
Hyperkid123: it's on the line above. Th segment analytics

src/js/App/analytics/segment-analytics.tsx
```diff
@@ -0,0 +1,122 @@
...
+    if (activeModule && user) {
+      const newKey = getAPIKey(isProd() ? 'prod' : 'dev', activeModule as SegmentModules);
+      const identityOptions = { cloud_user_id: user.identity.internal.account_id, adobe_cloud_visitor_id: getAdobeVisitorId() };
+      const groupOptions = {
+        account_number: user.identity.account_number,
+        account_id: user.identity.internal.org_id,
+      };
+      if (!analytics.current) {
+        analytics.current = AnalyticsBrowser.load({ writeKey: newKey }, { initialPageview: false });
+        window.segment = analytics.current;
```
luciddreamz, line 88: We might want to set this as `window.analytics` or any apps trying to call `analytics.track` that are following the docs on Segment's website are probably going to be confused.

src/js/App/analytics/segment-analytics.tsx
```diff
@@ -0,0 +1,122 @@
...
+  useEffect(() => {
+    const disconnect = registerUrlObserver();
+    return () => disconnect();
+  }, []);
+
+  useEffect(() => {
+    if (activeModule && user) {
+      const newKey = getAPIKey(isProd() ? 'prod' : 'dev', activeModule as SegmentModules);
+      const identityOptions = { cloud_user_id: user.identity.internal.account_id, adobe_cloud_visitor_id: getAdobeVisitorId() };
+      const groupOptions = {
```
luciddreamz, line 82: Here's a fairly typical output from the current group call:

```
"account": {
  "id": "5583404",
  "account_number": "5583404",
  "account_id": "7378958",
  "cloud_org_id": "7378958",
  "cloud_ebs_id": "5583404"
}
```

Again, unfortunately trying to preserve that <sigh>, but also switching to `org_id` as the primary key for Group ID:

```
const groupTraits = {
  account_number: user.identity.account_number,
  account_id: user.identity.internal.org_id,
  cloud_org_id: user.identity.internal.org_id,
  cloud_ebs_id: user.identity.account_number
};
//...
analytics.current.group(
  user.identity.internal.org_id // switching to org_id as the Group ID
  groupTraits
);
```

src/js/App/analytics/segment-analytics.tsx
```diff
@@ -0,0 +1,122 @@
...
+        account_number: user.identity.account_number,
+        account_id: user.identity.internal.org_id,
+      };
+      if (!analytics.current) {
+        analytics.current = AnalyticsBrowser.load({ writeKey: newKey }, { initialPageview: false });
+        window.segment = analytics.current;
+        analytics.current.identify(user.identity.internal.account_id, identityOptions);
+        analytics.current.group(user.identity.account_number, groupOptions);
+        analytics.current.page();
```
luciddreamz, line 91: We should strip "/beta" out of the URL's so we're tracking the canonical pages (since most pages are available at both URLs), and just track `beta` as metadata about the page. Plus `module` would probably be pretty useful metadata. Finally, we also need to add `groupId` as context to pretty much all the calls that aren't the group call itself just to be safe and ensure compatibility with destinations that might expect group ID and user ID in every call.

```
function getPagePath() {
  return window.location.pathname.replace(/^\/beta(\/)?/i, '/');
}
function getPagePathSegment(n) {
  const pathSegment = getPagePath().split('/');
  return pathSegment[n] ? pathSegment[n] : '';
}
function getPageURL() {
  const url = window.location.href;
  if (isBeta()) {
    url = url.replace(/\/beta(\/)?/);
  }
  const i = url.indexOf('#');
  return i === -1 ? url : url.slice(0, i);
}
//...
const pageProperties = {
  path: getPagePath(), // strip beta from path
  url: getPageURL(), // strip beta and hash from url
  beta: isBeta(), // add beta to page metadata
  module: activeModule
};
const pageContext = {
  context: {
    groupId: user.identity.internal.org_id
  }
};
//...
analytics.current.page(pageProperties, pageContext);
```

luciddreamz (changes requested):

src/js/App/analytics/segment-analytics.tsx
Hyperkid123: The official API for applications is not going to be the window object. The `segmentAnalytics` object instance is exposed when using the `useChrome` hook. It is exposed here: https://github.com/RedHatInsights/insights-chrome/pull/1918/files#diff-90cf814c0a22d1dbb9311fa142c3e931efafebf0a19c9d58851cbdfcd9fd028fR107

I don't mind renaming the object key.

Hyperkid123:
@luciddreamz I have incorporated all the changes you requested. You can check the last 4 commits. Let me know if there is anything else we have to change.

PR #1921, opened 2022-06-10
Remove loadInventory and loadRemediations

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-19831

Removed deprecated functions `loadInventory` and `loadRemediations` 

Instead of `loadInventory` we have async inventory https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/src/Inventory/InventoryTable.js

Instead of `loadRemediations` we have FED modules for it - https://github.com/RedHatInsights/frontend-components/blob/master/packages/remediations/doc/remediations.md#hot-loading-the-wizard-directly

@Hyperkid123


karelhala (approved):
Let's see if anybody is still using this functions

PR #1926, opened 2022-06-15
Replace custom KC with newest implementation

karelhala:
### Start using official KC implementation

Since KC has released new version of KC javascript adapter and looks like a lot of changes fixed our hacks we can start using official implementation of KC JS adapter.

With such addoption a few changes needed to be done, like using `new Keycloak` instead of static `Keycloak` (which is now deprecated).


Hyperkid123 (commented):
@karelhala everything seems to be OK in the UI. There are some type failures during tests tho

Hyperkid123 (approved):

Hyperkid123:
:fire:

PR #1931, opened 2022-06-20
Refactoring chrome entry point

victoria-dos:
https://issues.redhat.com/browse/RHCLOUD-15880


src/js/bootstrap.js
```diff
@@ -1,63 +1,121 @@
...
+        dispatch(loadModuesSchema(response.data));
+      });
+  }, []);
+
+  useEffect(() => {
+    if (typeof documentTitle === 'string') {
+      document.title = `${documentTitle} | console.redhat.com`;
+    } else {
+      document.title = `console.redhat.com`;
+    }
```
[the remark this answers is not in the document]
victoria-dos: that is so elegant

fhlavac (approved):
Looks good! :tada:

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
-  trustarcScript.src = '//static.redhat.com/libs/redhat/marketing/latest/trustarc/trustarc.js';
-} else {
-  trustarcScript.src = '//static.redhat.com/libs/redhat/marketing/latest/trustarc/trustarc.stage.js';
-}
+  return modules && scalprumConfig && jwtState ? <RootApp config={scalprumConfig} /> : null;
+};
 
-document.body.appendChild(trustarcScript);
+const { store } = spinUpStore();
+const pageRoot = document.getElementById('chrome-entry');
```
karelhala, line 107: No need to assign these to constants (moreover if it's not in a function).

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
 
-import renderChrome from './chrome/render-chrome';
-// render root app
-renderChrome();
+ReactDOM.render(
+  <Provider store={store}>
+    <App />
+  </Provider>,
+  pageRoot
```
karelhala, line 113: I would call directly the getElementById
```suggestion
  document.getElementById('chrome-entry')
```

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
 
-document.body.appendChild(trustarcScript);
+const { store } = spinUpStore();
+const pageRoot = document.getElementById('chrome-entry');
 
-import renderChrome from './chrome/render-chrome';
-// render root app
-renderChrome();
+ReactDOM.render(
+  <Provider store={store}>
```
karelhala, line 110: You can call `spinUpStore` and access the store out of it.
```suggestion
  <Provider store={spinUpStore()?.store}>
```

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
-    console.log('Unable to parse initial account. Using default account');
-    Cookies.remove(CROSS_ACCESS_ACCOUNT_NUMBER);
   }
-}
+};
+
+const libjwtSetup = () => {
+  const libjwt = auth();
+
+  function noop() {}
```
karelhala, line 37: I'd move this function to constant file, this could be usefull on other places as well.

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
+      .then((...data) => {
+        sentry(...data);
+      })
+      .catch(noop);
+  });
+
+  return libjwt;
+};
+
+const trustarcScriptSetup = () => {
```
karelhala, line 51: Could be usefull to move this to a utils file so we clean this file a bit.

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
+    axios
+      .get(`${window.location.origin}${isBeta() ? '/beta' : ''}/config/chrome/fed-modules.json?ts=${Date.now()}`, {
+        headers: {
+          'Cache-Control': 'no-cache',
+          Pragma: 'no-cache',
+          Expires: '0',
+        },
+      })
+      .then((response) => {
+        dispatch(loadModuesSchema(response.data));
```
karelhala, line 94: Since we're refactoring this file I'd fix the type on this function
```suggestion
        dispatch(loadModulesSchema(response.data));
```

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
-  registerUrlObserver(window._satellite.pageBottom);
-}
+    axios
+      .get(`${window.location.origin}${isBeta() ? '/beta' : ''}/config/chrome/fed-modules.json?ts=${Date.now()}`, {
+        headers: {
+          'Cache-Control': 'no-cache',
+          Pragma: 'no-cache',
+          Expires: '0',
+        },
+      })
```
karelhala, line 92: Could be usefull to move this to a utils file.

src/js/bootstrap.js
```diff
@@ -1,63 +1,114 @@
...
+  libjwt.initPromise.then(() => {
+    libjwt.jwt
+      .getUserInfo()
+      .then((...data) => {
+        sentry(...data);
+      })
+      .catch(noop);
+  });
+
+  return libjwt;
```
karelhala, line 48: What if the user is not logged in? Do we want to break the flow and login the user or continue with rest of the calculations? I'd guess breaking the cycle. Let's return directly the result from `libjwt.initPromise`

karelhala (commented):
Looking good! Just a few questions, mostly about code cleanup and auth cycle.

src/js/bootstrap.js
victoria-dos: But I use libjwt later on the line 76 in **createChromeInstance**, isn't it a problem?

src/js/bootstrap.js
```diff
@@ -1,63 +1,89 @@
...
-  window._satellite.pageBottom();
-  registerUrlObserver(window._satellite.pageBottom);
-}
+    trustarcScriptSetup();
 
-const trustarcScript = document.createElement('script');
-trustarcScript.id = 'trustarc';
+    loadFedModules().then((response) => {
+      dispatch(loadModulesSchema(response.data));
+    });
```
karelhala, line 73: Since we're using just `data` from the response, let's break it. Also no need to have function expression, just call the dispatch directly.
```suggestion
    loadFedModules().then(({ data }) => dispatch(loadModulesSchema(data)));
```

karelhala (commented):
Looking good! One final thing and we are golden!

karelhala (approved):

PR #1932, opened 2022-06-21
alphabetize imports

epwinchell:


Hyperkid123 (changes requested):
Instead of doing this manually we can use the eslint rule and its automated fixes: https://eslint.org/docs/rules/sort-imports

fhlavac (approved):

PR #1939, opened 2022-06-24
Ensure the `org_id` is on the top-level of the user object

coderbydesign:
The identity header now adds `identity.org_id`, and the UI user object should also
reflect this change.


karelhala (approved):
@coderbydesign does this mean we can remove the `org_id` from internal object?

PR #1940, opened 2022-06-24
Adjust expectation in user spec

coderbydesign:
The test for the user object structure was still passing after https://github.com/RedHatInsights/insights-chrome/pull/1939
which was due to the `toMatchObject` matcher, which matches a subject. This updates
the test data and the expectation to match the shape completely.


karelhala (approved):

romanblanco (approved):
LGTM :+1:

PR #1945, opened 2022-06-30
add org_id for entitlements call

ryelo:
entitlements can use org_id instead of account number now. Add org_id as another restriction so customers could have one or the other.

### DIFF
--- a/src/js/jwt/insights/user.ts
+++ b/src/js/jwt/insights/user.ts
@@ -116,13 +116,16 @@ export default async (token: SSOParsedToken) => {
   }
 
   if (user) {
-    log(`Account Number: ${user.identity.account_number}`);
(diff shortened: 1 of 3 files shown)


ryelo:
Splitting this into two PRs

src/js/jwt/insights/user.ts
```diff
@@ -116,13 +116,16 @@ export default async (token: SSOParsedToken) => {
...
+    const accNumber =  user.identity.account_number;
+    const orgId = user.identity.org_id;
+
     try {
-      if (user.identity.account_number) {
+      if (accNumber || orgId) {
         data = await servicesApi(token.jti).servicesGet();
       } else {
-        console.log('Cannot call entitlements API, no account number');
+        console.log(`Cannot call entitlements API. Account number: ${accNumber}, Org ID: ${orgId}`);
```
aneelac22, line 128: How can they be logged if they don't exist?

PR #1951, opened 2022-07-04
Add option to use global config SSO url

karelhala:
### Custom SSO url

There's a request to change SSO url when app starts in eph environment. This PR adds option to consume chrome config from global fed-mods.json where `ssoUrl` will be located.

### Pass global config to global chrome

With addition of said SSO url we want to also add option to observe this global config from within running app. To do so this PR adds `globalConfig` entry to global chrome object which contains every available app and its config (including fed mods stuff).

### DIFF
--- a/src/js/auth.ts
+++ b/src/js/auth.ts
@@ -24,7 +24,7 @@ export function crossAccountBouncer() {
   window.location.reload();
 }
 
-export default () => {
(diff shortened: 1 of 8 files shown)


karelhala:
Ping @psav btw this adds the global ssoUrl config we've been talking a while back.

Hyperkid123 (approved):

PR #1962, opened 2022-07-18
Wait for JWT validation before rendering modules.

Hyperkid123:
Some apps were throwing errors because the USER was not ready when the apps were rendering. We now postpone any module rendering only after the JWR initialization was successfull

### DIFF
--- a/src/js/bootstrap.js
+++ b/src/js/bootstrap.js
@@ -30,14 +30,16 @@ const initializeAccessRequestCookies = () => {
   }
 };
 
-const libjwtSetup = (chromeConfig) => {
+const libjwtSetup = (chromeConfig, setReadyState) => {
   const libjwt = auth(chromeConfig || {});
 
   libjwt.initPromise.then(() => {
-    libjwt.jwt
+    return libjwt.jwt
       .getUserInfo()
       .then((...data) => {
+        console.log('sentry step');
         sentry(...data);
+        setReadyState(true);
       })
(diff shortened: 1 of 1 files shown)


radekkaluzik (approved):

src/js/bootstrap.js
```diff
@@ -57,8 +59,7 @@ const App = () => {
       const { chrome: chromeConfig } = data;
       dispatch(loadModulesSchema(data));
       initializeAccessRequestCookies();
-      const libjwt = libjwtSetup(chromeConfig);
-      libjwt.initPromise.then(() => setJwtState(true));
+      const libjwt = libjwtSetup(chromeConfig, setJwtState);
```
karelhala, line 62: I am not a big fan of sending callback state set function to a rather complex multi promise function. How about we'd change the `libjwtSetup` function to be
```JS
const libjwtSetup = (libjwt) => {
  return libjwt.initPromise.then(() => {
    return libjwt.jwt
      .getUserInfo()
      .then((...data) => {
        console.log('sentry step');
        sentry(...data);
      })
      .catch(noop);
  });
};
```

And were to change the `useEffect` to
```JS
const libjwt = auth(chromeConfig || {});
libjwtSetup(libjwt).then(() => setReadyState(true));
```

karelhala (commented):
Just a bit of suggestion on cleaner `libjwtSetup` function.

PR #1976, opened 2022-07-28
Add segment dev debug function

### LINKED ISSUE #1972: Analytics: isProd check includes local dev env URL
When we launch our local dev env at `prod.foo.redhat.com` and send analytics track events, it uses the prod write key
https://github.com/RedHatInsights/insights-chrome/blob/883f5ce806dc8f8de5fade7f7ba255fff0a09e04/src/js/utils.ts#L143

I would expect to use the dev write key from a local dev env.

Hyperkid123:
fixes #1972 

jira: https://issues.redhat.com/browse/RHCLOUD-20597

Forces segment to use DEV key

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -10,3 +10,4 @@ There are some localStorage values for you to enable debuging information or ena
 * `remediationsDebug` - to enable debug buttons in remediations app
 * `shortSession` - to enable short session in order to test automatic logouts
 * `forcePendo` - to force Pendo initializtion
+* `segmentDev` - force usage of common segment dev API key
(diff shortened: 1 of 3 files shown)


docs/debugging.md
```diff
@@ -10,3 +10,4 @@ There are some localStorage values for you to enable debuging information or ena
 * `remediationsDebug` - to enable debug buttons in remediations app
 * `shortSession` - to enable short session in order to test automatic logouts
 * `forcePendo` - to force Pendo initializtion
+* `segmentDev` - forece usage of common segment dev API key
```
aneelac22, line 13:
```suggestion
* `segmentDev` - force usage of common segment dev API key
```

karelhala (approved):

aneelac22 (approved):

PR #1979, opened 2022-08-01
Ts migration

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-20636

### Changes

- Expose complete redux store state type
- Migrate Active component to TS
- Migrate Banner component to TS
- Migrate CrossRequest notifier and its dependencies to TS


src/js/App/CrossRequestNotifier/CrossRequestNotifier.tsx
```diff
@@ -1,52 +1,47 @@
...
 );
 
-DescriptionComponent.propTypes = {
-  id: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.string), PropTypes.string]).isRequired,
-  markRead: PropTypes.func.isRequired,
-};
-
 const CrossRequestNotifier = () => {
-  const [, forceRender] = useReducer((state = false) => !state);
+  const [, forceRender] = useReducer((state) => state + 1, 0);
```
karelhala, line 40: Any reason for the `+1`?

karelhala (approved):
Looking good just a confirmation question.

src/js/App/CrossRequestNotifier/CrossRequestNotifier.tsx
Hyperkid123: Yes. Actually using the `true/false` can sometimes run through multiple updates and will not resolve in a forced re-render. We had this issue in DDF and figured we can't use a simple boolean flip. Hence the counter.

karelhala: Okay, works for me

PR #1981, opened 2022-08-01
add icon with hover message to account number

InsaneZein:
Part of [Jira 18660](https://issues.redhat.com/browse/RHCLOUD-18660?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=true)

account-num-hint

~Still need to add case where Account Number is not visible~


Hyperkid123 (changes requested):
Something is wrong with the tooltip placement. I have tested din in a chrome browser.
tooltip

InsaneZein:
That's really strange. I'm seeing that now, but I wasn't seeing it yesterday. I'll take a look.

InsaneZein:
> Something is wrong with the tooltip placement. I have tested din in a chrome browser.

I changed the icon placement based on rereading the JIRA, but I've done a lot of testing with different browsers and screen sizes and the tooltip is in the correct place. I can't reproduce the issue I was having yesterday. 

account-num-hint_2

src/js/App/Header/UserToggle.js
```diff
@@ -20,9 +23,14 @@ function buildItems(username, isOrgAdmin, accountNumber = -1, isInternal, extraI
     </DropdownItem>,
     <React.Fragment key="account wrapper">
       {accountNumber > -1 && (
-        <DropdownItem key="Account" isDisabled>
```
Hyperkid123, line 23: This will cause issues with a cursor. See the attached recording:
Peek 2022-08-04 10-41

I get that the tooltip does not work if the item is disabled. But we have to change the styling if the child elements to fix the cursor.

src/js/App/Header/UserToggle.js
```diff
@@ -20,9 +23,14 @@ function buildItems(username, isOrgAdmin, accountNumber = -1, isInternal, extraI
...
+        <DropdownItem key="Account" isPlainText>
           <dl className="chr-c-dropdown-item__stack">
-            <dt className="chr-c-dropdown-item__stack--header">Account number:</dt>
+            <dt className="chr-c-dropdown-item__stack--header">
+              Account number:
+              <Tooltip id="accountNumber-tooltip" content={accountNumberTooltip}>
```
Hyperkid123, line 30: I think there should be some spacing between the text and the icon. Probably `pf-u-ml-sm` class on the tooltip.

InsaneZein: the `pf-u-ml-sm` wasn't working but I've added some spacing between the icon and title

src/js/App/Header/UserToggle.js
InsaneZein: I've fixed this issue in the latest commit. I used mouse-events. I'm not sure if there's a cleaner way. If there is, please let me know.

Hyperkid123 (commented):
Almost there

src/js/App/Header/UserToggle.js
Hyperkid123: @InsaneZein ml is marging left. There are equivalent classes for other directions. If that does not work either, you can use the PF CSS variables: https://www.patternfly.org/v4/utilities/spacing

InsaneZein: :sweat_smile: the styling worked this time. I don't why it didn't last time. 

image

Hyperkid123 (approved):

PR #1999, opened 2022-08-09
Use JWT token only on API routes

karelhala:
### Description

Since we don't need to include JWT token to each static assets request we can attach the `cs_jwt` token only to `/api` paths. This way we'll remove huge amount of request header.

Only place where this is currently failing is landing page, however that's because landing page is touching cookies directly. This PR fixes such issue https://github.com/RedHatInsights/landing-page-frontend/pull/415

### DIFF
--- a/src/js/jwt/jwt.ts
+++ b/src/js/jwt/jwt.ts
@@ -347,7 +347,7 @@ export function setCookie(token?: string) {
   if (token && token.length > 10) {
     const cookieName = priv.getCookie()?.cookieName;
     if (cookieName) {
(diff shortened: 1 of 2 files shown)


Hyperkid123 (approved):

Hyperkid123:
@karelhala the test is missig the `/api` path can you add it please?
https://app.travis-ci.com/github/RedHatInsights/insights-chrome/builds/254188937

karelhala:
@Hyperkid123 aaah, good point!

PR #2001, opened 2022-08-10
Add flex-grow: 1 to default layout

karelhala:
### Description

The default layout does not span over entire height of the content. This is caused by flex display set on the parent component and if the content does not occupy entire screen it will be cutt off.

#### Current

Screenshot from 2022-08-10 09-15-45

#### New

Screenshot from 2022-08-10 09-15-51

### DIFF
--- a/src/js/App/RootApp/DefaultLayout.scss
+++ b/src/js/App/RootApp/DefaultLayout.scss
@@ -27,6 +27,7 @@
 }
 
 .chr-scope__default-layout {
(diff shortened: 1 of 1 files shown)


Hyperkid123 (commented):
@karelhala just making sure. Is this a new bug after the recent DOM layout changes?

karelhala:
@Hyperkid123 somewhat related, yeah.

Hyperkid123 (approved):

PR #2002, opened 2022-08-10
Reproduce page issue

Hyperkid123:
This is here to remove all possible code that could cause [this issue](https://github.com/patternfly/patternfly-react/issues/7810) from the chrome side, while being easily able to reproduce it.


Hyperkid123:
Issue was resolved

PR #2007, opened 2022-08-15
Visual regression testing

Hyperkid123:
Based on the recent layout issue it would be a good thing to implement some simple visual regression testing: https://issues.redhat.com/browse/RHCLOUD-20740

This PR adds a setup and a sample test for visual regression testing that will help us eliminate potential platform-wide breaking changes dues to layout/pf/styling changes

Jira task: https://issues.redhat.com/browse/RHCLOUD-20815

### Failed diff example

Banner -- mounts diff


cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
...
+// declare global {
+//   namespace Cypress {
+//     interface Chainable {
+//       login(email: string, password: string): Chainable<void>
+//       drag(subject: string, options?: Partial<TypeOptions>): Chainable<Element>
+//       dismiss(subject: string, options?: Partial<TypeOptions>): Chainable<Element>
+//       visit(originalFn: CommandOriginalFn, url: string, options: Partial<VisitOptions>): Chainable<Element>
+//     }
+//   }
+// }
```
aneelac22, line 37: Is this file needed? Did you add this for future reference?

aneelac22 (approved):
LGTM!

cypress/support/commands.ts
Hyperkid123: Its a file that is added by the cypress setup. I suspect we will eventually require it to add a new API to the cypress.

PR #2011, opened 2022-08-17
Enable Translations

aneelac22:
[Jira Ref RHCLOUD-20762]


Hyperkid123 (commented):
@aneelac22 can you provide a read me on how the translation files are consumed?

aneelac22:
> @aneelac22 can you provide a read me on how the translation files are consumed?

Sure!

scripts/createDataJson.js
```diff
@@ -0,0 +1,53 @@
+const program = require('commander');
```
Hyperkid123, line 1: I don't see this new module in package.json dependecies

scripts/createDataJson.js
```diff
@@ -0,0 +1,53 @@
...
+const fs = require('fs');
+const { sync: globSync } = require('glob');
+const last = require('lodash/last');
+
+let LANG_DIR = 'src/locales/';
```
Hyperkid123, line 6: So for directories, I would prefer if we use the `path.resolve` or other `path` functions. It will generate absolute paths so we can be sure we get the right FS paths.

scripts/createDataJson.js
```diff
@@ -0,0 +1,53 @@
...
+      return { [locale]: JSON.parse(fs.readFileSync(filename, 'utf8')) };
+    }
+  })
+  .reduce((acc, localeObj) => {
+    return { ...acc, ...localeObj };
+  }, {});
+
+// Merge aggregated default messages with the translated json files and
+// write the messages to this directory
+fs.writeFileSync(`${rootFolder}${LANG_DIR}data.json`, JSON.stringify({ ...mergedTranslations }, null, 2));
```
Hyperkid123, line 53: No need for the spread operator.

Hyperkid123 (commented):
I am not sure why but when I ran the `npm run translations` the command just froze. This is the whole output I got:
```
yarn run v1.22.19
$ npm-run-all translations:*
$ npx formatjs extract --ignore ./types.d.ts ./src/js/*.{js,jsx,ts,tsx} --out-file ./locales/Messages.json
[@formatjs/cli] [WARN] Reading source file from TTY.


```
Nothing seems to be happening.

scripts/createDataJson.js
aneelac22: It is in node modules, no need to add as a new package

aneelac22:
Fixed the extract freezing. It should work fine now

scripts/createDataJson.js
Hyperkid123: Well, it's a good practice to do so if we are using it directly. Its likely a peerDependency of another module. NPM installs it but we should have it listed in dev dependencies as well.

aneelac22: Cool! will do it, it is a dependency of formatjs so, I thought if it removed from there there probably will be a different approach

Hyperkid123 (approved):

PR #2021, opened 2022-09-02
fix ctrl-click causing route change or refresh

InsaneZein:
[RHCLOUD-20357](https://issues.redhat.com/browse/RHCLOUD-20357)

Ctrl-click doesn't cause an in-app routing event after opening new tab.


src/js/App/Sidenav/Navigation/ChromeLink.tsx
```diff
@@ -84,6 +84,9 @@ const LinkWrapper: React.FC<LinkWrapperProps> = ({ href, isBeta, onLinkClick, cl
   };
   const dispatch = useDispatch();
   const onClick = (event: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
+    if (event.ctrlKey) {
```
Hyperkid123, line 87: I think we also want to handle a `shift` key click. That opens a new browser window.

Can you also a a test for the `ChromeLink` that the URL wont change after firing an event with ctrt/shift keys pressed?

Hyperkid123 (approved):

PR #2041, opened 2022-09-15
Remove sso URL file, it was causing errors while trying to load it

karelhala:
### Description

Remove SSO url file loader. We had a mechanism, that populated SSO url trough this file, however recentrly the UI started to fail while trying to load said file. This PR removes such mechanism, because we now have a much easier and robust mechanism (trough consuming chrome config from `fed-modules.json`).


psav (approved):

Hyperkid123:
And we can all rejoice :100: :tada: !

PR #2065, opened 2022-10-02
Refactor global filter

karelhala:
## Description

We need to adjust the global filter to new customer needs, before doing so we should refactor it so it's easier to use.

#### Code cleanup

While cleaning the code, new hooks were introduced to better isolate certain logic, these hooks are used just within specific component so we don't have to directly test them.

#### Removed features

This PR removes one feature - storing global filter in localstorage, it proved to be "magic" for customers and didn't brought any real value to any of the applications.

#### Follow-up

`GlobalFilter.js` is still written in plain JS instead of typescript, however it's blocked by usage of `useTagsFilter` which has incomplete types - https://github.com/RedHatInsights/frontend-components/pull/1615


src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -1,211 +1,68 @@
...
+    ({ globalFilter: { tags, sid, workloads } }) => ({
       count: tags.count + sid.count + workloads.count,
       total: tags.total + sid.total + workloads.total,
-      sapCount: workloads.hasSap,
-      aapCount: workloads.hasAap,
-      mssqlCount: workloads.hasMssql,
-      isDisabled: globalFilterHidden || !appId,
+      tags: tags.items || [],
+      sid: sid.items || [],
+      workloads,
```
Hyperkid123, line 57: Why do we not return the workload items? Does use a different data structure?

src/js/App/GlobalFilter/GlobalFilter.js
```diff
@@ -1,211 +1,68 @@
...
-          search: filterTagsBy,
-        })
-      );
-    });
-  };
-  const debouncedLoadTags = useCallback(debounce(loadTags, 800), []);
   const { filter, chips, selectedTags, setValue, filterTagsBy } = useTagsFilter(
-    [...workloads, ...sid, ...tags],
-    isLoaded && Boolean(token),
+    [workloads, ...sid, ...tags],
```
Hyperkid123, line 64: related to the above

src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -7,126 +7,132 @@ import debounce from 'lodash/debounce';
...
+const usePagination = (loaded: boolean | unknown, perPage?: number, page?: number, count?: number) => {
+  return useMemo(() => {
+    if (loaded) {
+      return {
+        perPage: perPage,
+        page: page,
+        count: count,
+      };
+    }
+    return {};
```
Hyperkid123, line 42: I am wondering if we maybe want to return attributes with the value 0  instead of undefined.

src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -7,126 +7,132 @@ import debounce from 'lodash/debounce';
...
+          cells: cellsMapper(tag),
+        }))
       );
-    }, 800),
-    []
-  );
-  const debounceGetSIDs = useCallback(
+    }
+    return [];
+  }, [resource, selected]);
```
Hyperkid123, line 67: the `loaded` variable is missing in the dependencies

src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -7,126 +7,132 @@ import debounce from 'lodash/debounce';
...
+        count: count,
+      };
+    }
+    return {};
+  }, [loaded, perPage, page, count]);
+};
+
+const useRow = (
+  resource: SID[] | GlobalFilterTag[],
+  loaded: boolean | unknown,
```
Hyperkid123, line 48: Why `unknown` type? Do we ever send something else than a boolean? And if so why?

src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -7,126 +7,132 @@ import debounce from 'lodash/debounce';
...
+        callback(
           {
-            registeredWith: filterScope,
-            activeTags: selectedTags,
+            registeredWith,
+            activeTags,
             search,
           },
-          { page: sidPage, perPage: sidPerPage }
+          { page: 1, perPage }
```
Hyperkid123, line 82: Is the page filter always 1?

src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -7,126 +7,132 @@ import debounce from 'lodash/debounce';
...
+};
+
+const TagsModal = ({ isOpen, filterTagsBy, onApplyTags, toggleModal, selectedTags }: TagsModalProps) => {
+  const intl = useIntl();
+  const [tagsSelected, setTagsSelected] = useState<CommonSelectedTag[]>([]);
+  const [sidsSelected, setSidsSelected] = useState<CommonSelectedTag[]>([]);
+  const [filterBy, setFilterBy] = useState('');
+  const [filterSIDsBy, setFilterSIDsBy] = useState('');
+  const dispatch = useDispatch();
+  const [tagsLoaded, tagsCount, tagsPage, tagsPerPage] = useMetaSelector('tags');
```
Hyperkid123, line 97: Maybe we can have an `enum` for the hook arguments? To avoid typos in the future?

src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -7,126 +7,132 @@ import debounce from 'lodash/debounce';
...
+  );
+  const sidRows = useRow(
+    sids,
+    sidLoaded,
+    ({ key }) => key as string,
+    ({ key }) => [key],
+    sidsSelected
+  );
+
+  console.log(tagsSelected, sidsSelected, 'mmmmm');
```
Hyperkid123, line 126: doktor oetker? :wink:

Hyperkid123 (commented):
Got some initial review round. Looking good.

src/js/App/GlobalFilter/GlobalFilter.js
karelhala: Yup, workloads have slightly different data structure.

src/js/App/GlobalFilter/TagsModal.tsx
karelhala: Good point!

src/js/App/GlobalFilter/TagsModal.tsx
karelhala: I don't think so, if the pagination sees empty config it doesn't render. However if we were to sent `count: 0` it would render it and then once data comes in it would flip it to the real data counter.

src/js/App/GlobalFilter/TagsModal.tsx
karelhala: It can be something else, but given the problem with inventory throwing 404 if no data are found for said page I think it's safer to always return to first page.

src/js/App/GlobalFilter/TagsModal.tsx
karelhala: Mhh, instead of returning an array, to return object with `{ tagsLoaded: boolean, tagsCount: number... }`?

src/js/App/GlobalFilter/TagsModal.tsx
Hyperkid123: Ah yeah that bit.

src/js/App/GlobalFilter/TagsModal.tsx
karelhala: Yup, the `loaded` from metaSelector can be `unknown`

Hyperkid123 (approved):

PR #2068, opened 2022-10-03
new chrome API to add a quickstart to a namespace

radekkaluzik:
https://issues.redhat.com/browse/RHCLOUD-20677


src/js/redux/index.ts
```diff
@@ -66,6 +68,7 @@ import {
 } from './action-types';
 import { ChromeState, GlobalFilterState, ReduxState } from './store';
 import { AnyAction } from 'redux';
+//import { addQuickstart } from './actions';
```
Hyperkid123, line 71: extra comment

src/js/redux/reducers.ts
```diff
@@ -250,6 +250,19 @@ export function populateQuickstartsReducer(
   };
 }
 
+export function addQuickstartstoApp(state: ChromeState, { app, quickstart }: { app: string; quickstart: QuickStart }) {
```
Hyperkid123, line 253: Can you write some tests for the function?

src/js/redux/index.ts
radekkaluzik: removed

src/js/redux/reducers.ts
radekkaluzik: done

Hyperkid123 (approved):

PR #2076, opened 2022-10-11
Consume isDisabled in global filter

karelhala:
It won't work until https://github.com/RedHatInsights/frontend-components/pull/1626 is merged and fec version updated. 

EDIT: This required a bit more work due to type missmatch. Updated the fec package to newest version which required a couple of fixes as well.


src/js/App/GlobalFilter/TagsModal.tsx
```diff
@@ -129,27 +131,28 @@ const TagsModal = ({ isOpen, filterTagsBy, onApplyTags, toggleModal, selectedTag
...
-      isOpen={isOpen}
-      toggleModal={(_e: unknown, isSubmit: boolean) => {
+      isOpen={isOpen as boolean}
+      toggleModal={(_e?: any, open?: boolean) => {
         setSidsSelected([]);
         setTagsSelected([]);
         setFilterBy('');
         setFilterSIDsBy('');
-        toggleModal(isSubmit);
+        toggleModal(open as boolean);
```
Hyperkid123, line 144: Instead of the TS override can we do a `!!open` or in the argument make it optional with default value `_e?: any, open = false`?

Better than typecasting it.

Hyperkid123:
@karelhala can you fix the conflict?

PR #2106, opened 2022-11-04
Purge quickstarts storage when switching between applications.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-22833


src/components/RootApp/RootApp.tsx
```diff
@@ -77,7 +81,7 @@ const RootApp = (props: RootAppProps) => {
   };
   return (
     <Router history={chromeHistory}>
-      <SegmentProvider activeModule={activeModule}>
+      <SegmentProvider activeModule={activeModule!}>
```
karelhala, line 84: Interesting usage of non-null operator. Do I interpret this correctly as throw error is `activeModule` is null at build time?

karelhala (approved):

src/components/RootApp/RootApp.tsx
Hyperkid123: The module is just always string at this point. But it can be undefined in the redux store.

PR #2108, opened 2022-11-08
Update text based on UX review and fix translations flow

aneelac22:
[RHCLOUD-20863]


ryelo (approved):

aneelac22:
> @aneelac22 I see a completely new file with translations. Is that expected?
Yeah, looks like the original files were empty when npm run translations was run before because it was trying to reference non-existent files

Issue #2036, opened 2022-09-11
build fails on Fedora 36 with ERR_OSSL_EVP_UNSUPPORTED


cben:
Fedora 36 blocked some insecure hash functions in openssl.  Webpack defaults to md4 in file names, resulting in `npm run build` giving a long stacktrace:
```
...
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:67:19)
    at Object.createHash (node:crypto:130:10)
    at BulkUpdateDecorator.hashFactory (/home/beni/ocm/insights-chrome/node_modules/webpack/lib/util/createHash.js:145:18)
    at BulkUpdateDecorator.update (/home/beni/ocm/insights-chrome/node_modules/webpack/lib/util/createHash.js:46:50)
    at RawSource.updateHash (/home/beni/ocm/insights-chrome/node_modules/webpack-sources/lib/RawSource.js:70:8)
    at NormalModule._initBuildHash (/home/beni/ocm/insights-chrome/node_modules/webpack/lib/NormalModule.js:888:17)
    at handleParseResult (/home/beni/ocm/insights-chrome/node_modules/webpack/lib/NormalModule.js:954:10)
    at /home/beni/ocm/insights-chrome/node_modules/webpack/lib/NormalModule.js:1048:4
    at processResult (/home/beni/ocm/insights-chrome/node_modules/webpack/lib/NormalModule.js:763:11)
    at /home/beni/ocm/insights-chrome/node_modules/webpack/lib/NormalModule.js:827:5 {
  opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
  library: 'digital envelope routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_EVP_UNSUPPORTED'
}
```

https://stackoverflow.com/questions/69394632/webpack-build-failing-with-err-ossl-evp-unsupported/ has some suggestions...

Hyperkid123:
Hmm. Not sure if we can do much about it. Did you try any of the solutions posted in the linked discussion? It seems more like OS/Webpack issue rather than chrome.

Do you run the LTS node?

cben:
I think I got it working with slightly newer webpack, will send PR.

PR #2112, opened 2022-11-10
webpack: Use xxhash64 in emitted filenames

cben:
Fixes https://github.com/RedHatInsights/insights-chrome/issues/2036.

Webpack defaults to md4 in file names, using implementation from OpenSSL library.  Some builds e.g. Fedora 36 disable it completely, resulting in ERR_OSSL_EVP_UNSUPPORTED error.
This switches to another hash function (also insecure but we don't need crypto strength in file names, just low probability of collisions), which is implemented in pure JS, so should work on all platforms.

Slightly newer webpack version was required.


Hyperkid123 (approved):
Thanks for the update.

I do agree that we do not need strong hashes for filenames. I am just sad that is something we have to be considering.

PR #2120, opened 2022-11-18
Fix left nav flickering

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-23070

### Changes
- reduce the number of overall rendering cycles
- move the global filter visibility flag to the global filter component
- remove the `dynamic` flag from chrome link (was required for chrome one apps)

### Before
flicker-old

### After
flicker-new


fhlavac (approved):
Looks good!

PR #2165, opened 2022-12-14
Print sentry debug info, adjust tracesSampleRate

gkarat:
This increases the value for `tracesSampleRate` to accept more transactions. Also, I propose to enable debug information in console. 

This is due to some unexpected behavior we experience in production: less we set the value, more transactions start to come. Documentation states the opposite behavior to be expected, and we were asked by Sentry support to set the higher value with the 1/10 precision. The debug info must not cause any safety concerns: https://docs.sentry.io/platforms/javascript/guides/react/configuration/options/#debug.

### DIFF
--- a/src/utils/sentry.ts
+++ b/src/utils/sentry.ts
@@ -73,7 +73,8 @@ function initSentry() {
     maxBreadcrumbs: 50,
     attachStacktrace: true,
     integrations: [new BrowserTracing()],
-    tracesSampleRate: 0.007,
+    tracesSampleRate: 0.2,
+    debug: !!window.localStorage.getItem('chrome:sentry:debug'),
   });
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/sentry.ts
```diff
@@ -73,7 +73,8 @@ function initSentry() {
     maxBreadcrumbs: 50,
     attachStacktrace: true,
     integrations: [new BrowserTracing()],
-    tracesSampleRate: 0.007,
+    tracesSampleRate: 0.2,
+    debug: true,
```
Hyperkid123, line 77: We usually use local storage flags to enable dbeugging modes.

We have these functions: https://github.com/RedHatInsights/insights-chrome/blob/dd4b078586c5981781eca0636b4759050f2b7492/src/utils/debugFunctions.ts

They are available on `window.insights.chrome.enable`, so you can turn on/off the debugging when required.

Hyperkid123 (changes requested):

src/utils/sentry.ts
gkarat: Oh, cool. PTAL at the updated commit.

Hyperkid123 (approved):

PR #2173, opened 2022-12-19
Update patternfly react core dependency

suyash-naithani:
We need to update @patternfly/react-core package to use the updated select component which fixes a bug that affects our current UI.
Following the update here, we plan to bump the PF version on the ui-components repo as well

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "insights-chrome",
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@suyash-naithani I've fixed the lock file issue

Hyperkid123 (approved):

suyash-naithani:
Thank you @Hyperkid123  :)

PR #2178, opened 2023-01-05
Silently fail if feature flags API throws server error.

Hyperkid123:
Fixes infinite nav loading if feature flags API returns 500 status response.

The build-in error handler of the unleash client is not triggered on 500 for some reason. This will act as if the API returned no toggles but will not block the UI features. The user just won't have any feature flags.


src/components/FeatureFlags/FeatureFlagsProvider.tsx
```diff
@@ -52,6 +71,7 @@ const FeatureFlagsProvider: React.FC = ({ children }) => {
     return <FlagProvider config={config}>{children}</FlagProvider>;
   }
 
+  // unleashClientInternal.current
```
fhlavac, line 74: probably not needed

fhlavac (approved):

PR #2181, opened 2023-01-05
AllServices round 2

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-23500

- Update links
- Remove dummy search
- Replace "All Apps and Services" dropdown in masthead with "All Services" button linking to page (will move to nav later)
- Goes to prod-beta

Screen Shot 2023-01-05 at 10 42 17 AM

### DIFF
--- a/src/components/Header/Header.scss
+++ b/src/components/Header/Header.scss
@@ -13,6 +13,23 @@
   .chr-c-brand {
     width: 193px;
   }
+
+  a.chr-c-button-allservices  {
+    display: block;
+    height: 75px;
+    background: transparent;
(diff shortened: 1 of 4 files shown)


aneelac22 (approved):

Hyperkid123 (approved):
Good for now.

We have to render the page from a JSON array in the future to reduce the number of same JSX.

epwinchell:
> Good for now.
> 
> We have to render the page from a JSON array in the future to reduce the number of same JSX.

I think it will use cloud-services-config like the app filter did.

PR #2197, opened 2023-01-11
Masthead updates: Services icon & Favorites placeholder button

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-23607

- Add icons
- Add Favorites teaser (disabled) button

Screen Shot 2023-01-11 at 11 45 44 AM


src/components/Header/Header.tsx
```diff
@@ -52,8 +55,13 @@ export const Header = () => {
             <ToolbarGroup variant="filter-group">
               {user && (
                 <ToolbarItem>
-                  <Button className="chr-c-button-allservices" component={(props) => <ChromeLink {...props} href="/AllServices" />}>
-                    All Services
+                  <Button className="chr-c-button-masthead" component={(props) => <ChromeLink {...props} href="/AllServices" />}>
+                    <CloudIcon />
+                    Services
+                  </Button>
+                  <Button className="chr-c-button-masthead disabled" component={(props) => <ChromeLink {...props} href="#" />}>
```
Hyperkid123, line 62: This should be hidden behind a feature flag. We don't want to show this page/button before its ready.

epwinchell: @Hyperkid123 Kat would like to show it as a teaser. That's why I styled it as a disabled link.

Hyperkid123: Ah, I was not aware of that. OK then

Hyperkid123 (approved):

PR #2211, opened 2023-01-13
Integrate chrome service client

Hyperkid123:
### Changes
- track the user's last visited pages
- enable favoriting pages in all services page
- reflect the current environment in all services pages (remove links not available in the current env based on chrome navigation files)

### DIFF
--- a/src/components/AllServices/allServicesLinks.ts
+++ b/src/components/AllServices/allServicesLinks.ts
@@ -1,0 +1,377 @@
+import AllServicesIcons from './AllServicesIcons';
+
+export type AllServicesLink = {
+  href: string;
+  title: string;
+  isExternal?: boolean;
+};
+export type AllServicesGroup = {
+  isGroup: true;
+  title: string;
+  links: AllServicesLink[];
+};
+export type AllServicesSection = {
+  icon: keyof typeof AllServicesIcons;
+  title: string;
+  description: string;
+  links: (AllServicesLink | AllServicesGroup)[];
+};
+
+const allServicesLinks: AllServicesSection[] = [
+  {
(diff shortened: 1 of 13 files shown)


aneelac22 (approved):

src/components/AllServices/allServicesLinks.ts
```diff
@@ -0,0 +1,377 @@
...
+  links: AllServicesLink[];
+};
+export type AllServicesSection = {
+  icon: keyof typeof AllServicesIcons;
+  title: string;
+  description: string;
+  links: (AllServicesLink | AllServicesGroup)[];
+};
+
+const allServicesLinks: AllServicesSection[] = [
```
BlakeHolifield, line 20: Any thoughts on loading this from an external source? It might be easier to version control separately rather than binding it to a chrome release.

BlakeHolifield (approved):
Just one question. Otherwise looks good

src/components/AllServices/allServicesLinks.ts
Hyperkid123: Yeah, thats a good point. Currently, we reference it with the navigation files, but I agree the structure should be extracted into another file, similar to what we do with the navigation files.

BlakeHolifield: Sure we don't have to address this right now, just a thought for the future.

epwinchell:
@Hyperkid123 I don't we want to enable favoriting for external links at this point. The UX design alots the same real estate for the ext link icon and the star in the nav

Screen Shot 2023-01-18 at 10 39 24 AM
Screen Shot 2023-01-18 at 10 46 25 AM

PR #2227, opened 2023-01-18
Add the config to the container via the Dockerfile

adamrdrew:
This should copy the config-files during build into the container. Please sanity check this syntax - I do not write a lot of Dockerfiles! This is where the files should be populated: https://github.com/RedHatInsights/insights-frontend-builder-common/blob/master/src/frontend-build.sh#L47


adamrdrew:
This will be blocked on https://github.com/RedHatInsights/insights-chrome/pull/2228 - my bad

PR #2245, opened 2023-01-24
Spreach chromeConfig.config and chromeConfig to jwtSetup

karelhala:


Hyperkid123 (commented):
can the chromeConfig be undefined? If that is true we might need `?.` to prevent runtime crashes in non-clowder environments.

karelhala:
@Hyperkid123 updated

Hyperkid123 (approved):

PR #2248, opened 2023-01-25
Send cookie to server if path starts with /wss or /ws

karelhala:
### Description

Send cookie with requests when they start either with `/ws`, `/wss` or `/api`.


Hyperkid123 (approved):

config/webpack.config.js
```diff
@@ -141,7 +141,7 @@ const commonConfig = ({ dev }) => {
       // HMR flag
       hot: true,
       ...proxy({
-        env: 'stage-beta',
+        env: 'prod-beta',
```
fhlavac, line 144: do we want to push this?

Hyperkid123: No

karelhala: Good point!

fhlavac (approved):

PR #2249, opened 2023-01-26
Add Azure redirect banner

aneelac22:
[RHCLOUD-23504]


src/components/Stratosphere/RedirectBanner.test.tsx
```diff
@@ -49,9 +49,9 @@ describe('<RedirectBanner>', () => {
     expect(screen.queryByText('Congratulations, your Red Hat and AWS accounts are linked')).not.toBeInTheDocument();
   });
 
-  test('should return inline alert if required query param does not exist', () => {
+  test('should return inline alert if required query param exists', () => {
     render(
-      <Wrapper store={store} initialEntries={[`/foo/bar?${STRATOSPHERE_BANNER_NAME}`]}>
+      <Wrapper store={store} initialEntries={[`/foo/bar?${AWS_BANNER_NAME}`]}>
```
Hyperkid123, line 54: Can we add new test instead of replacing the current one?

src/components/Stratosphere/RedirectBanner.tsx
```diff
@@ -5,17 +5,26 @@ import { useLocation, useNavigate } from 'react-router-dom';
...
 // TODO: Figure out what param chrome should expect
-export const STRATOSPHERE_BANNER_NAME = 'from-aws';
+export const AWS_BANNER_NAME = 'from-aws';
+export const AZURE_BANNER_NAME = 'from-azure';
 
 const RedirectBanner = () => {
   const { pathname, search, hash, state } = useLocation();
   const navigate = useNavigate();
   const params = new URLSearchParams(search);
+  const partner = () => {
```
Hyperkid123, line 15: Can we make `partner` directly into a variable instead of a function? Its just an extra step that is not really necessary is the function is called in each render anyways.

You can also wrap it in `useMemo` and use the `search` as its trigger.

Hyperkid123 (approved):

PR #2255, opened 2023-01-30
Integrate with openshift dynamic sdk.

Hyperkid123:
### TODO
- [x] make sure HCC modules are working as expected
- [x] migrate HAC modules to the SDK v2
- [ ] merge required changes to hac-core
- [x] use stable release of scalprum


config/webpack.plugins.js
```diff
@@ -38,6 +38,7 @@ const plugins = (dev = false, beta = false) => {
         { 'react-dom': { singleton: true, eager: true, requiredVersion: deps['react-dom'] } },
         { 'react-router-dom': { singleton: true, requiredVersion: deps['react-router-dom'] } },
         { 'react-redux': { requiredVersion: deps['react-redux'] } },
+        { '@openshift/dynamic-plugin-sdk': { singleton: true, requiredVersion: deps['@openshift/dynamic-plugin-sdk'] } },
```
vojtechszocs, line 41: These `requiredVersion: package.dependencies['module-name']` bits (here and above/below) might not be necessary.

AFAIK webpack module federation plugin code already attempts to auto-detect this information from dependencies in the project's `package.json` file.

src/chrome/update-shared-scope.ts
```diff
@@ -0,0 +1,72 @@
...
+    return to;
+  };
+
+  const modules = {
+    'react-router-dom': async () => () => {
+      // We have to hack our way around react-router-dom
+      // Since we are no longer using basename we have to include `/hac` prefix
+      const reactRouter = require('react-router-dom');
+      return {
+        ...reactRouter,
```
vojtechszocs, line 28: This is an interesting way to hack into / extend the existing shared modules 😃

src/components/RootApp/ScalprumRoot.tsx
```diff
@@ -78,6 +79,8 @@ const ScalprumRoot = memo(
     }
 
     useEffect(() => {
+      // prepapre webpack module sharing scope overrides
```
vojtechszocs, line 82:
```suggestion
      // prepare webpack module sharing scope overrides
```

vojtechszocs (changes requested):

config/webpack.plugins.js
Hyperkid123: I think you might be right. We should take a look at it during our next tech debt sweep.

src/chrome/update-shared-scope.ts
Hyperkid123: I am not a huge fan but for now it's the only way if we don't want to enforce code changes in existing hac plugins :upside_down_face: . We are hoping to come up with some generic plugin(extenstion) that would simulate the basename.

florkbr: I think this will replicate the same behavior/limitation we have today with HAC running on Chrome. We should have a larger discussion around our stance going forward on this. I don't think we're opposed to making code changes in HAC plugins - more. so do we want another explicit relationship between plugins and Chrome versus abstracting this functionality (whether that abstraction is an override, extension, utility, etc)

karelhala (approved):

PR #2258, opened 2023-01-30
Add sentry replay to config

adonispuente:
In an effort to make sentry errors more understandable (per my squad assignment), I want to add the replay functionality to our config. This allows us to watch the replays similar to hotjar, except with sentry now we'll know what error is associated to what bug. 

I've added the sentry-mask class to block certain aspects that hotjar was also blocking. I also added sentrys capture message function to report errors above 400 as a warning as opposed to an error

I followed the instructions found here 
https://docs.sentry.io/platforms/javascript/guides/react/session-replay/?_ga=2.107964160.348271133.1675046398-87740463.1664195656


Hyperkid123 (approved):

Hyperkid123:
Had to re-open the PR to wake up the bot

src/components/FeatureFlags/FeatureFlagsProvider.tsx
```diff
@@ -22,8 +23,9 @@ const config: IFlagProvider['config'] = {
       .fetch(url, headers)
       .then((resp) => {
         // prevent the request from falling back to default error behavior
```
jiridostal, line 25: Correct me if I'm wrong, but doesn't this change "re-enable" bubbling this error up to the default error handler?

Hyperkid123: No, there is a catch branch below, that is the actual error handler.

jiridostal: I see, thanks!

PR #2270, opened 2023-02-08
Caddy for frontend

adamrdrew:
This PR moves from the nginx config to caddy for the webserver for the frontend container. I tested this to the best of my ability locally by building the image and running it - it seemed to work though nothing rendered. That may be expected behavior if there's no dashboard or anything? At very least I know the Dockerfile and Caddyfile syntax are valid, that the container builds, caddy serves, and I can get to index.html and all the various js files. That's about as much as I know to do locally, so we'll need to test on the clusters from here.


aneelac22 (approved):
The code looks good to me

PR #2278, opened 2023-02-13
Load navigation chunks on console.dev

karelhala:
### Description

console.dev.redhat.com recently stopped working because navigation chunks were not loaded. This PR disables unleash check for console.dev origin.

### DIFF
--- a/cypress/component/helptopics/HelpTopicManager.cy.tsx
+++ b/cypress/component/helptopics/HelpTopicManager.cy.tsx
@@ -2,7 +2,7 @@
 
 import React, { useEffect } from 'react';
 import { Provider } from 'react-redux';
-import { Store } from 'redux';
+import { AnyAction, Store } from 'redux';
 import ReducerRegistry from '@redhat-cloud-services/frontend-components-utilities/ReducerRegistry';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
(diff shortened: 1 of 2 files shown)


src/utils/useNavigation.ts
```diff
@@ -103,7 +103,7 @@ const useNavigation = () => {
     let observer: MutationObserver | undefined;
     // reset no nav flag
     setNoNav(false);
-    if (currentNamespace && (flagsReady || flagsError)) {
+    if (currentNamespace && (flagsReady || flagsError || location.origin.includes('console.dev.rehdat.com'))) {
```
flacatus, line 106: console.dev.rehdat.com => console.dev.redhat.com ?

Hyperkid123:
@karelhala the unleash check should be disabled dif it fails to retrieve the data and acts as if all flags were false. I think you might have encountered a scenario that the current check did not cover.

karelhala:
@Hyperkid123 the problem here is that the unleash is not deployed on console.dev.redhat.com and fails to load the navigation. I was trying to make this in a hotfix manner, but I'll probably launch chrome locally, change the url for unleash and do a proper fix.

Hyperkid123:
The fetch handler should be able to resolve 404s with no issues. Maybe we are getting different error?

Hyperkid123 (approved):

PR #2288, opened 2023-02-20
Only fetch one host for AAS and MSSQL host counts

PaulWay:
At the moment these requests fetch up to 50 hosts, which can use several tens of kilobytes just to get the total number of hosts matching this query.  This (hopefully) reduces the number of hosts actually returned to one.


Hyperkid123:
Hi @PaulWay can you link/create a Jira ticket for tracking purposes thanks. @karelhala is this going to be OK with the global filter?

karelhala (approved):

karelhala:
@Hyperkid123 yeah, this should be just fine. We are getting the count from meta object which is sent in every request, so fetching just one item should be just fine (it'll at least slim down the response)

PR #2300, opened 2023-02-21
add patternfly/react-table and react-icons as shared packages

mkholjuraev:
This PR adds `@patternfly/react-icons`,` @patternfly/react-table` packages to the list of shared libraries in the default scope. With this in place, dependant apps, will not load their local copy of the package, but use the shared one as both the fallback module and the main instance in the shared scope. This PR should be looked at together with this demo [Patch PR](https://github.com/RedHatInsights/patchman-ui/pull/966). 

**The bundle size decreased from 3.56 mb to 1.28 mb.**

Before the change: 
before the standalone packages

After the change:
After standalone packages

The actual bundle size loaded into browser has decreased from 23 mb to 12 mb

Before the change: 
browser before standalone packages

After the change:
brwoser afer standalone packages

### DIFF
--- a/config/webpack.plugins.js
+++ b/config/webpack.plugins.js
@@ -40,6 +40,8 @@ const plugins = (dev = false, beta = false) => {
         { 'react-redux': { requiredVersion: deps['react-redux'] } },
         { '@openshift/dynamic-plugin-sdk': { singleton: true, requiredVersion: deps['@openshift/dynamic-plugin-sdk'] } },
         { '@patternfly/react-core': { requiredVersion: deps['@patternfly/react-core'] } },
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@mkholjuraev these package will be added to the shared scope as chrome is not using these exact import paths anywhere. can you test your patch demo without these changes?

mkholjuraev:
@Hyperkid123 I have already tested as you described. but I was not able to declare those packages as a singleton on Patch. Because they do not exist in the shared scope and the application breaks.

Hyperkid123:
If a module is not in a shared scope and the container requires the, it should add it. Can you try again?

mkholjuraev:
Closing this for the sake of having consistent versions by periodically monitoring the new package version conflict on the console,  updating deps to match the version with the chrome.

PR #2306, opened 2023-02-28
Update modules not in scope, auth and Satellite registration for frh

aneelac22:
Fixing some styles on Satellite Page but, other than that it is ready

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -10,7 +10,7 @@ import { ACTIVE_REMOTE_REQUEST, CROSS_ACCESS_ACCOUNT_NUMBER } from './utils/cons
 import auth, { LibJWT, crossAccountBouncer } from './auth';
 import sentry from './utils/sentry';
 import registerAnalyticsObserver from './analytics/analyticsObserver';
-import { getEnv, isFedRamp, loadFedModules, noop, trustarcScriptSetup } from './utils/common';
+import { ITLess, getEnv, loadFedModules, noop, trustarcScriptSetup } from './utils/common';
 import messages from './locales/data.json';
 import ErrorBoundary from './components/ErrorComponents/ErrorBoundary';
 import LibtJWTContext from './components/LibJWTContext';
@@ -56,10 +56,11 @@ const libjwtSetup = (chromeConfig: { ssoUrl?: string }) => {
   return libjwt;
 };
 
+const isITLessEnv = ITLess();
+
 const useInitialize = () => {
   const [{ isReady, libJwt }, setState] = useState<{ isReady: boolean; libJwt?: LibJWT }>({ isReady: false, libJwt: undefined });
   const store = useStore();
-  const fedRampEnv = isFedRamp();
   const chromeInstance = useRef({ cache: undefined });
   useEffect(() => {
     // We have to use `let` because we want to access it once jwt is initialized
@@ -84,7 +85,7 @@ const useInitialize = () => {
     // setup trust arc
     trustarcScriptSetup();
     // setup adobe analytics
-    if (!fedRampEnv && typeof window._satellite !== 'undefined' && typeof window._satellite.pageBottom === 'function') {
+    if (!isITLessEnv && typeof window._satellite !== 'undefined' && typeof window._satellite.pageBottom === 'function') {
       window._satellite.pageBottom();
       registerAnalyticsObserver();
     }
@@ -106,6 +107,10 @@ const App = () => {
     const title = typeof documentTitle === 'string' ? `${documentTitle} | ` : '';
     document.title = `${title}console.redhat.com`;
   }, [documentTitle]);
+
+  if (isITLessEnv) {
+    return isReady && modules && scalprumConfig ? <RootApp config={scalprumConfig} /> : <AppPlaceholder />;
+  }
 
   return isReady && modules && scalprumConfig && libJwt ? (
     <LibtJWTContext.Provider value={libJwt}>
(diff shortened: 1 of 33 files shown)


src/bootstrap.tsx
```diff
@@ -56,10 +56,17 @@ const libjwtSetup = (chromeConfig: { ssoUrl?: string }) => {
   return libjwt;
 };
 
+const fedRampEnv = isFedRamp();
+
+// const useFrhInitialize = () => {
```
Hyperkid123, line 61: I see some old comment

src/bootstrap.tsx
```diff
@@ -79,6 +86,7 @@ const useInitialize = () => {
           libJwt,
           isReady: true,
         });
+        setIsFedRampReady(true);
```
Hyperkid123, line 89: Should we trigger this even in non-fedramp env?

src/bootstrap.tsx
```diff
@@ -93,20 +101,25 @@ const useInitialize = () => {
...
   const documentTitle = useSelector(({ chrome }: ReduxState) => chrome?.documentTitle);
-  const { isReady, libJwt } = useInitialize();
+  const { isReady, libJwt, isFedRampReady } = useInitialize();
 
   useEffect(() => {
     const title = typeof documentTitle === 'string' ? `${documentTitle} | ` : '';
     document.title = `${title}console.redhat.com`;
   }, [documentTitle]);
 
+  if (isFedRampReady && modules && scalprumConfig && fedRampEnv) {
```
Hyperkid123, line 119: I think we want to slightly enhance the condition so we can also get the placeholder component in.

```
if(fedRampEnv) {
  return isFedRampReady && modules && scalprumConfig ? <RootApp config={scalprumConfig} /> : <AppPlaceholder />
}
```

src/bootstrap.tsx
```diff
@@ -93,20 +101,25 @@ const useInitialize = () => {
...
+    return <RootApp config={scalprumConfig} />;
+  }
+
   return isReady && modules && scalprumConfig && libJwt ? (
     <LibtJWTContext.Provider value={libJwt}>
```
Hyperkid123, line 124: We will have to be very careful about the potentially missing Provider.

src/cognito/auth.ts
```diff
@@ -0,0 +1,185 @@
...
+  const searchParamsObject: { [key: string]: string } = {};
+  for (const [key, value] of searchParams) {
+    searchParamsObject[key] = value;
+  }
+  return searchParamsObject;
+}
+
+export async function getTokenWithAuthorizationCode() {
+  const code = getSearchParams(window.location.href).code;
+  const refreshToken = localStorage.getItem('REFRESH_TOKEN');
```
Hyperkid123, line 27: Is the `REFRESH_TOKEN` set by the cognito or did we pick that specific key name? I would to have a slightly more unique local storage key if possible.

src/cognito/auth.ts
```diff
@@ -0,0 +1,185 @@
...
+      body: `grant_type=authorization_code&client_id=${dataConfig?.ClientId}&redirect_uri=${redirectUri}&code=${code}`,
+    });
+
+    if (!response.ok) {
+      throw new Error(`Request failed with status code ${response.status}`);
+    }
+
+    const tokens = await response.json();
+    localStorage.setItem('REFRESH_TOKEN', tokens.refresh_token);
+    localStorage.setItem('ACCESS_TOKEN', tokens.access_token);
```
Hyperkid123, line 74: We should probably store the storage keys in some constants so we have them in one place if we have to change them later on.

src/components/Header/Header.tsx
```diff
@@ -69,14 +71,14 @@ export const Header = () => {
                       ) : (
                         <>
                           <ServicesLink />
-                          <FavoritesLink />
+                          {isFedRamp() ? user?.identity?.user?.is_org_admin && <SatelliteLink /> : <FavoritesLink />}
```
Hyperkid123, line 74: Is the favorite link supposed to be hidden on fedramp?

src/components/RootApp/RootApp.tsx
```diff
@@ -110,19 +108,17 @@ const RootApp = memo((props: RootAppProps) => {
     <HistoryRouter history={chromeHistory as unknown as HistoryRouterProps['history']} basename={isBeta() ? '/beta' : '/'}>
       <SegmentProvider activeModule={activeModule}>
         <FeatureFlagsProvider>
-          <IDPChecker>
```
Hyperkid123, line 113: The IDP checker is required for normal environments. If it's problematic in fedramp it can be disabled, but it has to stay in the rest of the environments.

src/jwt/jwt.ts
```diff
@@ -235,9 +240,14 @@ export function isExistingValid(token?: string) {
 }
 
 // keycloak init successful
-export function initSuccess() {
+export async function initSuccess() {
```
Hyperkid123, line 243: I assume these are supposed to be promises?

src/components/RootApp/RootApp.tsx
aneelac22: I did local build and this seems to be not needed in stage either, keith told me it was added for fr in the first place

Hyperkid123: OK. If Keight says that then its probably true.

Hyperkid123:
@jjaquish This one is quite big and has the potential to break the whole console. Do you think we can run your IQE on this PR before we merge it?

Hyperkid123:
@aneelac22 I tested the PR in UI and it seems to run OK. But I would like to wait for some feedback from Jack. Is it ok of we wait for a bit on this one?

aneelac22:
> @aneelac22 I tested the PR in UI and it seems to run OK. But I would like to wait for some feedback from Jack. Is it ok of we wait for a bit on this one?

Ideally I would like for it to be merged today so, I can reference actual build reference and lock the build to that commit and there are no surprise changes that will effect fr env

Hyperkid123 (approved):

PR #2307, opened 2023-03-01
Dependency clean up

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24470

### Changes
- use swc-loader insetad of babel and ts loaders
- update majority of outdated dependencies
- update some TS errors
- update PF to latest stable version


cypress.config.ts
```diff
@@ -1,39 +1,38 @@
-import { defineConfig } from "cypress";
-const {
-  addMatchImageSnapshotPlugin,
-} = require('@simonsmith/cypress-image-snapshot/plugin');
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { defineConfig } from 'cypress';
+const { addMatchImageSnapshotPlugin } = require('@hyperkid/cypress-image-snapshot/plugin');
```
karelhala, line 3: Any reason for custom cypress image? I'd be more comfortable with forking it to RHInsights repository and deploying it under our org.

karelhala (commented):
Looking good! Just one question about custom cypress image.

cypress.config.ts
Hyperkid123: Yeah, it has locked peer dependency of cypress... I will open a PR in the original package but don't want to wait. I also don't want to take ownership of the package. I see this only as a temporary swap.

Hyperkid123: @karelhala fix was published in https://github.com/simonsmith/cypress-image-snapshot/pull/4. I've used the original dependency.

karelhala (approved):

PR #2322, opened 2023-03-08
Setup e2e testing environment

Hyperkid123:
Runs a simple cypress e2e test on every PR check. In the future, these tests should be expanded and serve as release gates.


fhlavac (approved):

fhlavac (approved):

scripts/e2e.js
```diff
@@ -0,0 +1,39 @@
...
+};
+let child;
+async function runTests() {
+  process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+  child = spawn('npm', ['run', 'dev:beta'], {
+    stdio: [process.stdout, process.stdout, process.stdout],
+    detached: false,
+  });
+  await waitOn(options);
+  execSync(`NO_COLOR=1 E2E_USER=${process.env.CHROME_ACCOUNT} E2E_PASSWORD=${process.env.CHROME_PASSWORD} npm run cypress run`, {
```
karelhala, line 24: I'd be cautious with using the password like this. Can we fill these values in .env file from vault?

karelhala (commented):
Looking good! I am a bit of unsure about passing the password trough env variables. They might leak because of shell printing.

scripts/e2e.js
Hyperkid123: @karelhala AFAIK the credentials are not printed anywhere. This command is not printed into the shell as.

PR #2327, opened 2023-03-13
make some shared package versions acept version range

mkholjuraev:
This is intended to have a wider package version choice while deciding the package version in the shared scope. If the chrome app accepts a wider range of package versions, consumer apps will be freer to choose/update their version. This helps to avoid version mismatches and duplicate packages in the shared scope.

We need to continue with this approach in the future package updates in chrome.


mkholjuraev:
@Hyperkid123 @karelhala I am not able to ask for your review on GH for some reason that I do now know.

Fewwy:
I have no issues with this PR, but I have a question about 1 thing. Have you forgotten to create a branch for this PR or you created this PR on the master branch for some purpose? :)

mkholjuraev:
@Fewwy yup, this is on purpose. I felt lazy to create a separate branch and delete afterwards...

package.json
```diff
@@ -174,7 +174,7 @@
     "@redhat-cloud-services/frontend-components-utilities": "^3.3.11",
     "@redhat-cloud-services/host-inventory-client": "1.2.0",
     "@redhat-cloud-services/rbac-client": "1.2.0",
-    "@scalprum/core": "0.4.0",
+    "@scalprum/core": "^0.4.0",
```
Hyperkid123, line 177: This has to be locked as the package is still not stable.

Hyperkid123 (commented):
Looks good. Scalprum rules can't be relaxed yet.

package.json
mkholjuraev: @Hyperkid123  done!

Hyperkid123 (approved):

PR #2332, opened 2023-03-14
Refresh page once on chunk load error

BlakeHolifield:
* Reloads page when a chunk-load-error is seen
* Adds guards to keep the page from refreshing in a loop


src/utils/common.ts
```diff
@@ -164,6 +165,20 @@ export function lastActive(searchString: string, fallback: string) {
 
 export const isAnsible = (sections: string[]) => (sections.includes('ansible') && sections.includes('insights') ? 1 : 0);
 
+// converts from whatever-this-case-is given by path to camelCase moduleName
+export const getModuleNameFromURL = (pathname: string) => {
```
Hyperkid123, line 169: You can actually get the currently active module from the redux store.
```TS
const activeModule = useSelector(({ chrome: { activeModule } }: ReduxState) => activeModule);
```

I would probably use this instead of parse it from the URL as it does not have to match the actual module.

Hyperkid123 (approved):

PR #2340, opened 2023-03-17
Use full profile from module's config

karelhala:
### Description

Since we don't want to duplicate code and tickets let's use full profile config from module's config. This value will be populated from frontend operator as well.


src/@types/types.d.ts
```diff
@@ -140,9 +140,7 @@ export type RemoteModule = {
 export type ChromeModule = {
   manifestLocation: string;
   ssoUrl?: string;
-  config?: {
-    ssoUrl: string;
-  };
+  config?: Record<string, unknown>;
```
Hyperkid123, line 143: As this is not a public type I think we might want to be as specific as possible. Just so we know what we might expect from apps to come. So far we only have ssoUrl and the profile requirement coming.

karelhala: Ah, good point!

Hyperkid123 (approved):

PR #2341, opened 2023-03-17
Add handlers for beta and preview to Caddyfile

adamrdrew:
This is to enable the beta environment and prepare for the move to preview


Caddyfile
```diff
@@ -14,7 +14,30 @@
...
+            index opt/app-root/src/build/index.html
+            browse
+        }
+    }
+
+    @preview_match {
+        path_regexp /preview/chrome(.*)
+    }
+    handle @preview_match {
+        uri strip_prefix /beta/chrome
```
karelhala, line 40: This should remove the preview prefix

```suggestion
        uri strip_prefix /preview/chrome
```

adamrdrew:
Sorry so many plates in the air I never got back to this lol - we're gonna go into stage-beta tomorrow afternoon so if we could get this in tomorrow that would be awesome!!!

Caddyfile
```diff
@@ -37,7 +37,7 @@
         path_regexp /preview/chrome(.*)
     }
     handle @preview_match {
-        uri strip_prefix /beta/chrome
+        uri strip_prefix /preview/chrome
```
Hyperkid123, line 40: Is it possible to keep both? There will be a period where both /beta and /preview should be active.

Hyperkid123 (approved):

PR #2348, opened 2023-03-21
Remove staleness filter from global filter

karelhala:
### Description

Inventory now returns by default all staleness type systems. This PR removes `stale` and `fresh` values from default filter to follow same approach as what is used in inventory.

### JIRA

RHCLOUD-24776


karelhala:
ping @computercamplove

PR #2350, opened 2023-03-21
Add chrome debugger

apinkert:
For [RHCLOUD-23603](https://issues.redhat.com/browse/RHCLOUD-23603). I'm using a lot of structure from the `Feedback` component to model the skeleton of this component. As it stands, this will only show a user's roles and entitlements in a modal as this is just to lay down the foundation for future additions. Future changes could include changing from a modal to a new sidebar (as the pendo debugger does). ~~Tests are currently failing until I get `ChromeAPI` updated with `toggleDebuggerModal` and `enabledDebugging` in a separate PR, so marking as draft at the moment.~~

debugger


src/components/Header/Header.tsx
```diff
@@ -53,6 +56,7 @@ export const Header = () => {
         </MastheadBrand>
       </MastheadMain>
       <MastheadContent>
+        {user?.identity?.account_number && !isFedRamp() && isDebuggerEnabled && ReactDOM.createPortal(<Debugger user={user} />, document.body)}
```
Hyperkid123, line 59: Just a nitpick:  I don't think we want the debugger to be part of a Header component. I would use the `RootApp` or its direct child as a host of the debugger component.

And because this component will be rarely used by actual users, I would lazily load it and wrap it in Suspens to make sure the code is only loaded when requested.

src/components/Debugger/DebuggerModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }, []);
+  const transformPermissions = Object.values(permissions).map((val, index) => {
+    return (
+      <tr key={index}>
+        <td>{val}</td>
+      </tr>
+    );
+  });
+
+  function renderSelection() {
```
Hyperkid123, line 87: Can you make this a component instead of scoped function? it's generally not a good practice to create these render functions instead of components.

src/components/Debugger/DebuggerModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+          ...acc,
+          [`userPermissionsList_${key}`]: userPermission.permission,
+        }),
+        {}
+      );
+      setPermissions(userPermissionsList);
+    }
+    getUserPermissions();
+  }, []);
+  const transformPermissions = Object.values(permissions).map((val, index) => {
```
Hyperkid123, line 79: This can be directly part of JSX or a component.

src/components/Debugger/DebuggerModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+      );
+    } else {
+      return null;
+    }
+  });
+
+  const [permissions, setPermissions] = React.useState({});
+  React.useEffect(() => {
+    async function getUserPermissions() {
+      const userPermissions = await window.insights.chrome.getUserPermissions();
```
Hyperkid123, line 67: You can import this function directly. No need for using the window object.

src/components/Debugger/DebuggerModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+export type DebuggerModalProps = {
+  user: DeepRequired<ChromeUser>;
+};
+
+const DebuggerModal = ({ user }: DebuggerModalProps) => {
+  const isOpen = useSelector<ReduxState, boolean | undefined>(({ chrome: { isDebuggerModalOpen } }) => isDebuggerModalOpen);
+  const dispatch = useDispatch();
+  const setIsModalOpen = (isOpen: boolean) => dispatch(toggleDebuggerModal(isOpen));
+  const handleCloseModal = () => {
```
Hyperkid123, line 24: This function wrapper is not necessary. It's not doing anything extra. Setter can be called directly with the `false` argument.

src/components/Debugger/Debugger.scss
```diff
@@ -0,0 +1,53 @@
+button.pf-c-button.chr-c-button-debugger {
+  svg {
+    transform: rotate(90deg);
+  }
+  position: fixed;
+  z-index: 100;
+  padding: 5px;
```
Hyperkid123, line 7: I would try using the PF variables as much as possible. You can find them here: https://www.patternfly.org/v4/developer-resources/global-css-variables#global-css-variables

src/components/Debugger/DebuggerModal.tsx
apinkert: I'm not sure where this is being exported if you don't mind pointing me in the right direction

apinkert:
@Hyperkid123 I decided to uncouple everything table-related from the modal itself and put it into a separate component to make it a little cleaner

Hyperkid123 (approved):

PR #2351, opened 2023-03-21
chore(sentry): Update error title messages and sentry package

adonispuente:
This just enables sentry errors to have clearer descriptions. They would look like this
Screen Shot 2023-03-21 at 12 32 49 PM

As opposed to this 
Screen Shot 2023-03-21 at 12 33 42 PM

At the moment the Error is always 'Unhandled UI runtime error', this changes that to the information disclosed when first expanding the dropdown on 'something when wrong' error pages.


src/components/ErrorComponents/DefaultErrorComponent.tsx
```diff
@@ -53,6 +53,7 @@ const DefaultErrorComponent = (props: DefaultErrorComponentProps) => {
       });
     }
   }, []);
+  const exceptionMessage = (props.error as Error)?.message ? (props.error as Error).message : 'Unhandled UI runtime error';
```
Hyperkid123, line 56: Nit: the message should be defined before the `useEffect`.

adonispuente: Originally I had it above, but for some reason, when testing it with the message being defined above the useEffect, the error results in the default 'Unhandled' default message 100% of the time. 
I just ran this test again to be sure : https://red-hat-it.sentry.io/issues/4027309030/?project=1484024&query=is%3Aunresolved+%21app_name%3Asubscriptions+%21app_name%3Aimage-builder&referrer=issue-stream

Hyperkid123: Because the error prop can be populated on a second render and the effect is triggered only on a mount. That changes with this PR: https://github.com/RedHatInsights/insights-chrome/pull/2332

Hyperkid123: Yes, I am aware of that `undefined` issue. Can we skip the sentry call if the error is empty? Or do you want to merge the PR as it is?

adonispuente: Sorry for the late update. I was on PTO. I made the change so it doesnt sent the empty error message. Good catch. @Hyperkid123

Hyperkid123 (approved):

PR #2362, opened 2023-03-24
All serviced dropdown integration.

Hyperkid123:
### Changes
- use a composable menu instead of dropdown
- add favorite tab
- enable favoriting interactions
- allow link groups to be rendered
- split the code into multiple components
- use feature flags to enable/disable new dropdown

### TO DO
- [x] check content and update if required
- [x] deal with mixed groups and plain links in a tab


epwinchell (approved):
Styling tweaks for a follow-up, but looks great!

PR #2365, opened 2023-03-27
Reset segment integration on KEY change.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24806


src/analytics/resetIntegrations.ts
```diff
@@ -0,0 +1,20 @@
+// reset segment integrations (pendo, intercom, ...) when API key has changed
+const resetIntegrations = () => {
+  console.log('Calling reset');
```
karelhala, line 3: Can we log this with `logger` instead of plain' old `console.log`?

karelhala (approved):
Looking good, just one minor question.

src/analytics/resetIntegrations.ts
Hyperkid123: I tell you what. It should not even be there :smile:

karelhala (approved):

PR #2367, opened 2023-03-28
Allow /preview pages to be displayed.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24082

### Changes
- allow `/preview` routes to have behavior as `/beta` routes
- `/beta` still works and its the primary preview route!

### DIFF
--- a/src/locales/Messages.ts
+++ b/src/locales/Messages.ts
@@ -429,9 +429,9 @@ export default defineMessages({
   tryThisFeatureInBeta: {
     id: 'tryThisFeatureInBeta',
     description:
-      'Try this feature in our Beta Environment on cloud.redhat.com/beta. The Beta Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
+      'Try this feature in our Preview Environment on console.redhat.com/preview. The Preview Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
     defaultMessage:
-      'Try this feature in our Beta Environment on cloud.redhat.com/beta. The Beta Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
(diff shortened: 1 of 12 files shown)


src/locales/Messages.ts
```diff
@@ -429,9 +429,9 @@ export default defineMessages({
   tryThisFeatureInBeta: {
     id: 'tryThisFeatureInBeta',
     description:
-      'Try this feature in our Beta Environment on cloud.redhat.com/beta. The Beta Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
+      'Try this feature in our Preview Environment on cloud.redhat.com/preview. The Preview Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
```
ryelo, line 432: We should probably go ahead and update the `cloud.redhat.com` to `console.redhat.com` while we're here

src/locales/Messages.ts
```diff
@@ -429,9 +429,9 @@ export default defineMessages({
...
     defaultMessage:
-      'Try this feature in our Beta Environment on cloud.redhat.com/beta. The Beta Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
+      'Try this feature in our Preview Environment on cloud.redhat.com/preview. The Preview Environment allows you to interact with new features in an active development space. Because beta pre-release software is still being developed, you may encounter bugs or flaws in availability, stability, data, or performance.',
```
ryelo, line 434: same here ^

src/locales/Messages.ts
Hyperkid123: I see. Do you want me to go through all the text and change the beta to preview in all places? This was kind of random as I was searching for /beta string.

Hyperkid123:
@karelhala @ryelo the Kamai rule was added and `/preview` no loads the beta chrome assets. I think we can merge this early next week to test if normal `/beta` was affected in any way. It should not be.

PR #2368, opened 2023-03-28
Lift all services dropdown data to parent component.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24933

### Changes
- prevent state updates after component unmounts for `useAllServices` hook
- lift data loading from `AllServicesMenu` to `AllServicesDropdown` to preserve the data when the menu is closed


Hyperkid123:
I found and fixed another issue with the CSS position of the dropdown that can occur:
Peek 2023-03-29 11-09

PR #2384, opened 2023-03-30
Add layout breadcrumbs.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24713

### Changes
- add breadcrumbs component to the default layout

Peek 2023-03-31 10-31


src/hooks/useBreadcrumbsLinks.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (activeNavSegment && isNavItems(activeNavSegment)) {
+      const activeNavigation = extractNavItemGroups(activeNavSegment);
+      const activeFragment = activeNavigation.find((item) => item?.active);
+      if (activeFragment && isExpandableNav(activeFragment)) {
+        const leafFragment = activeFragment.routes.find((item) => item.active);
+        const groupLink = (() => {
+          const appFragments = leafFragment?.href?.split('/');
+          appFragments?.pop();
+          return appFragments ? `${appFragments.join('/')}` : `/${bundleId}`;
+        })();
```
karelhala, line 51: Any reason to wrap this in IIFE?

Hyperkid123: I just felt like it tbh. I can replace it if you'd like.

karelhala: Yeah, would be good to change it, no need for extra complexity.

karelhala, line 51:
```suggestion
        const appFragments = leafFragment?.href?.split('/');
        appFragments?.pop();
        const groupLink = appFragments ? `${appFragments.join('/')}` : `/${bundleId}`;
```

karelhala (approved):

PR #2398, opened 2023-03-31
Use cloud services config to populate all services

karelhala:
### Description

Let's pull the services config from one place and sew the navigation chunks and information to it. This way we don't have to keeep the ordered list in chrome and can be kept somewhere else with references to the navigation.

* [x] must go after https://github.com/RedHatInsights/chrome-service-backend/pull/35
* [ ] optional https://github.com/RedHatInsights/chrome-service-backend/pull/38 for other than stage environment.

### JIRA

RHCLOUD-24829


src/components/AllServices/AllServicesGroup.tsx
```diff
@@ -4,17 +4,18 @@ import { ITLess } from '../../utils/common';
 
 import AllServicesLink from './AllServicesLink';
 import { AllServicesGroup } from './allServicesLinks';
+import type { AllServicesLink as AllServicesLinkType } from './allServicesLinks';
 
 export type AllServicesGroupProps = AllServicesGroup;
 const AllServicesGroup = ({ title, links }: AllServicesGroupProps) => {
-  const filteredLinks = ITLess() ? links.filter((link) => link.ITLess) : links;
+  const filteredLinks = ITLess() ? links.filter((link) => (link as AllServicesLinkType).ITLess) : links;
```
Hyperkid123, line 11: I think we probably want a proper type guard.

src/components/AllServices/AllServicesSection.tsx
```diff
@@ -12,24 +12,32 @@ export const isAllServicesGroup = (item: AllServicesGroupType | AllServicesLinkT
...
           <Text component={TextVariants.p} className="pf-u-mb-md">
-            {description}
+            {description || null}
           </Text>
           {filteredLinks.map((link, index) =>
-            isAllServicesGroup(link) ? <AllServicesGroup key={index} {...link} /> : <AllServicesLink key={index} {...link} />
+            isAllServicesGroup(link as AllServicesGroupType) ? (
+              <AllServicesGroup key={index} {...(link as AllServicesGroupType)} />
+            ) : (
+              <AllServicesLink key={index} {...(link as AllServicesLinkType)} />
```
Hyperkid123, line 39: Again I think we want a type guard here. I think there is already one that checks for the groups / links.

src/components/AllServices/allServicesLinks.ts
```diff
@@ -12,15 +12,16 @@ export type AllServicesGroup = {
...
 export type AllServicesSection = {
-  icon: keyof typeof AllServicesIcons;
+  id?: string;
+  icon?: keyof typeof AllServicesIcons;
   ITLess?: boolean;
   title: string;
-  description: string;
-  links: (AllServicesLink | AllServicesGroup)[];
+  description?: string;
+  links: (string | AllServicesLink | AllServicesGroup)[];
```
Hyperkid123, line 24: When can we get links as string? Each link requires both pathname and title

src/components/AllServicesDropdown/AllServicesGallery.tsx
```diff
@@ -13,10 +13,10 @@ const AllServicesGallery = ({ selectedService }: AllServicesGalleryProps) => {
   const sections: AllServicesGroup[] = [];
   const links: AllServicesLink[] = [];
   selectedService.links.forEach((link) => {
-    if (isAllServicesGroup(link)) {
-      sections.push(link);
+    if (isAllServicesGroup(link as AllServicesGroup)) {
```
Hyperkid123, line 16: We can change the `isAllServicesGroup` to accept `any` or `unknow` rather than typecasting the arguments. it kind of defeats the purpose of the typeguard.

src/components/Header/Header.tsx
```diff
@@ -43,7 +42,9 @@ const FeedbackRoute = ({ user }: { user: DeepRequired<ChromeUser> }) => {
 export const Header = ({ breadcrumbsProps }: { breadcrumbsProps?: Breadcrumbsprops }) => {
   const searchEnabled = useFlag('platform.chrome.search.enabled');
   const user = useSelector(({ chrome }: DeepRequired<ReduxState>) => chrome.user);
-  const navDropdownEnabled = useFlag('platform.chrome.navigation-dropdown');
+  // TODO: remove me!
+  const navDropdownEnabled = true;
+  // TODO: remove me!
```
Hyperkid123, line 47: You forgot

src/hooks/useAllServices.ts
```diff
@@ -24,7 +29,7 @@ const handleBundleResponse = (bundle: { data: Omit<BundleNavigation, 'id' | 'tit
             ...rest,
             navItems,
           },
-        }),
+        }).links,
```
Hyperkid123, line 32: Can links be `undefined`?

src/hooks/useAllServices.ts
```diff
@@ -36,7 +41,7 @@ const handleBundleResponse = (bundle: { data: Omit<BundleNavigation, 'id' | 'tit
     // regular NavItem
     return [...acc, rest];
   }, []);
-  return flatLinks.flat();
+  return { id: bundle.data.id as string, title: bundle.data.title as string, links: (flatLinks || []).flat() };
```
Hyperkid123, line 44: If the id and title cannot be `undefined` we should make sure to properly define the input type so we don't need to typecast as much.

src/hooks/useAllServices.ts
```diff
@@ -95,103 +100,114 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
   }, []);
 };
+
+const findNavItems = (
+  items: (string | AllServicesLink | AllServicesGroup)[],
+  availableLinks: { id?: string; title?: string; items: AvailableLinks }[]
+): (AllServicesLink | AllServicesGroup)[] =>
+  items
+    .map((item) => {
+      if (isAllServicesGroup(item as AllServicesGroup)) {
```
Hyperkid123, line 110: Same type guard casting. You won't need the additional typecast below if the type guard does not use it.

src/hooks/useAllServices.ts
```diff
@@ -95,103 +100,114 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
+            .catch((err) => {
+              console.error('Unable to load appfilter bundle', err, fragment);
+              return [];
+            })
+        )
+      ),
+    []
+  );
+  const fetchSections = useCallback(
+    async (): Promise<AllServicesSection[]> => (await axios.get(`${getChromeStaticPathname('services')}/services.json?ts=${Date.now()}`)).data,
```
Hyperkid123, line 156: We shoud not need the `ts=...` query param. The response should not be cached because of the API.

src/hooks/useAllServices.ts
```diff
@@ -95,103 +100,114 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
+            items: parseBundlesToObject(bundle.links.flat()),
+          };
+        });
+        const availableSections = sections
+          .reduce<AllServicesSection[]>((acc, { links, ...rest }) => {
+            return [
+              ...acc,
+              {
+                ...rest,
+                links: findNavItems((links || []) as string[], availableLinks).filter(Boolean),
```
Hyperkid123, line 177: You can initialize the links as empty array in the argument and you won't require the type cast.

src/hooks/useAllServices.ts
```diff
@@ -95,103 +100,114 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
+              },
+            ];
+          }, [])
+          .filter(({ links }: AllServicesSection) => {
+            if (links?.length === 0) {
+              return false;
+            }
+
+            return (links as AllServicesGroup[])?.filter(({ isGroup, links }) => !isGroup || links.length !== 0).flat().length !== 0;
```
Hyperkid123, line 186: The top-level links array is not a group all the time.

src/hooks/useAllServices.ts
```diff
@@ -95,103 +100,114 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
-    // run hook after data are loaded or filter value changed
-  }, [ready, filterValue]);
+  const linkSections = useMemo(() => filterAllServicesSections(availableSections, filterValue), [ready, filterValue]);
 
   // Provide a flat list of all avaiable links
   const servicesLinks = useMemo(
     () =>
       linkSections
-        .flatMap(({ links }) => links)
+        .flatMap(({ links }) => links as (AllServicesGroup | AllServicesLink)[])
```
Hyperkid123, line 210: The option of the link is a `string` is messing this all up. Maybe, it should have been added to the type and if necessary type cast the scenario that can expect it to be a `string`. That way we could remove a lot of casting.

src/components/AllServices/allServicesLinks.ts
karelhala: The services config from chrome service holds the link information as strings to reference the real thing `rhel.inventory` will be translated to 
```
{
            "id": "inventory",
            "appId": "inventory",
            "title": "Inventory",
            "href": "/insights/inventory",
            "product": "Red Hat Insights",
            "description": "View details about your Red Hat Enterprise Linux systems."
          },
```

src/components/AllServicesDropdown/AllServicesGallery.tsx
karelhala: Good point!

src/hooks/useAllServices.ts
karelhala: Ah, good point!

src/hooks/useAllServices.ts
karelhala: I don't think so, I can make it a safe accessor though.

src/hooks/useAllServices.ts
karelhala: The typecast is not even required here.

src/hooks/useAllServices.ts
karelhala: Aaah, good point!

src/hooks/useAllServices.ts
karelhala: Updated all `AllServicesGroup` not to include `string` as option and same goes for the sections.

src/components/Header/Header.tsx
Hyperkid123: @karelhala this has to be reverted. We want to feature flag to be in place for now.

src/@types/types.d.ts
```diff
@@ -53,12 +54,13 @@ export type NavItem = {
...
   permissions?: NavItemPermission[] | NavItemPermission;
   dynamicNav?: string;
+  description?: string;
 };
 
 export type BundleNavigation = {
   id: string;
   title: string;
-  navItems: NavItem[];
+  navItems?: NavItem[];
```
Hyperkid123, line 63: This does not have to be optional. Instead, we should use the `BundleNavigation` type for the first axios call in `useAllServices` `fetchNavitation` callback. We currently use `BundleNav`. But the fallback uses `BundleNavigation` anyways,

src/components/AllServices/AllServicesGroup.tsx
```diff
@@ -3,18 +3,19 @@ import React, { Fragment } from 'react';
...
-  const filteredLinks = ITLess() ? links.filter((link) => link.ITLess) : links;
+  const filteredLinks = ITLess() ? links.filter((link) => isAllServicesLink(link) && link.ITLess) : links;
   return (
     <Fragment>
       <Text component={TextVariants.p} className="pf-u-pt-xs pf-u-font-weight-bold">
         {title}
       </Text>
       {filteredLinks.map((link, index) => (
-        <AllServicesLink key={index} {...link} />
+        <AllServicesLink key={index} {...(link as AllServicesLinkType)} />
```
Hyperkid123, line 18: I've tried to compile the code without the type cast and it works just fine.

src/components/AllServices/AllServicesSection.tsx
```diff
@@ -4,32 +4,32 @@ import { ITLess } from '../../utils/common';
...
 export type AllServicesSectionProps = AllServicesSection;
 
-export const isAllServicesGroup = (item: AllServicesGroupType | AllServicesLinkType): item is AllServicesGroup => {
-  return (item as AllServicesGroupType).isGroup === true;
-};
-
 const AllServicesSection = ({ icon, title, description, links }: AllServicesSectionProps) => {
-  const TitleIcon = AllServicesIcons[icon];
-  const filteredLinks = ITLess() ? links.filter((link) => link.ITLess) : links;
+  const TitleIcon = AllServicesIcons[icon as keyof typeof AllServicesIcons];
```
Hyperkid123, line 12: You can do this to get around the type cast
```TS
const TitleIcon = icon ? AllServicesIcons[icon] : null;
```

src/components/AllServices/AllServicesSection.tsx
```diff
@@ -4,32 +4,32 @@ import { ITLess } from '../../utils/common';
...
+  const filteredLinks = ITLess() ? links.filter((link) => (link as AllServicesLinkType).ITLess) : links;
```
Hyperkid123, line 13: This type cast is not required anymore.

src/components/AllServices/AllServicesSection.tsx
```diff
@@ -4,32 +4,32 @@ import { ITLess } from '../../utils/common';
...
           <Text component={TextVariants.p} className="pf-u-mb-md">
-            {description}
+            {description || null}
           </Text>
           {filteredLinks.map((link, index) =>
-            isAllServicesGroup(link) ? <AllServicesGroup key={index} {...link} /> : <AllServicesLink key={index} {...link} />
+            isAllServicesGroup(link) ? (
+              <AllServicesGroup key={index} {...link} />
+            ) : (
+              isAllServicesLink(link) && <AllServicesLink key={index} {...link} />
```
Hyperkid123, line 31: You should not require the else branch type guard. Now that we don't typecast the first guard argument, TS will correctly evaluate the else branch type.

src/components/AllServicesDropdown/AllServicesGallery.tsx
```diff
@@ -14,9 +13,9 @@ const AllServicesGallery = ({ selectedService }: AllServicesGalleryProps) => {
   const links: AllServicesLink[] = [];
   selectedService.links.forEach((link) => {
     if (isAllServicesGroup(link)) {
-      sections.push(link);
+      sections.push(link as AllServicesGroup);
     } else {
-      links.push(link);
+      links.push(link as AllServicesLink);
```
Hyperkid123, line 18: No need for type casting in these two arrays pushes.

src/components/AllServicesDropdown/AllServicesGallerySection.tsx
```diff
@@ -14,7 +15,7 @@ const AllServicesGallerySection = ({ title, links }: AllServicesGallerySectionPr
       <div>
         <Gallery hasGutter>
           {links.map((link, index) => (
-            <AllServicesGalleryLink {...link} key={index} />
+            <AllServicesGalleryLink {...(link as AllServicesLinkType)} key={index} />
```
Hyperkid123, line 18: No typecast here

src/hooks/useAllServices.ts
```diff
@@ -61,7 +71,7 @@ const filterAllServicesLinks = (links: (AllServicesLink | AllServicesGroup)[], f
   return links.reduce<(AllServicesLink | AllServicesGroup)[]>((acc, link) => {
     // groups have links nested, we have to filter them as well
     if (isAllServicesGroup(link)) {
-      const groupLinks = filterAllServicesLinks(link.links, filterValue);
+      const groupLinks = filterAllServicesLinks(link.links as AllServicesLink[], filterValue);
```
Hyperkid123, line 74: This type cast does not seem to be required.

src/hooks/useAllServices.ts
```diff
@@ -95,105 +105,119 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
   const isMounted = useRef(false);
   const [filterValue, setFilterValue] = useState('');
   // TODO: move constant once the AppFilter is fully replaced
   const bundles = requiredBundles;
+  const fetchNavitation = useCallback(
+    () =>
+      Promise.all(
+        bundles.map((fragment) =>
+          axios
+            .get<BundleNav>(`${getChromeStaticPathname('navigation')}/${fragment}-navigation.json?ts=${Date.now()}`)
```
Hyperkid123, line 150: The bundle nav can be swapped for `BundleNavigation`. It also removes the need for `navItems` to be now optional in the `BundleNavigation` type.

src/hooks/useAllServices.ts
```diff
@@ -95,105 +105,119 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
+        )
+      ),
+    []
+  );
+  const fetchSections = useCallback(
+    async (): Promise<
+      (Omit<AllServicesSection, 'links'> & {
+        links: (string | AllServicesLink | AllServicesGroup)[];
+      })[]
+    > => (await axios.get(`${getChromeStaticPathname('services')}/services.json`)).data,
```
Hyperkid123, line 166: The type should be part of the `axios.get` generic. It will be safer for the future as we get the type directly as the axios response payload instead of forcing the callback return type.

```ts
  const fetchSections = useCallback(
    async () =>
      (
        await axios.get<
          (Omit<AllServicesSection, 'links'> & {
            links: (string | AllServicesLink | AllServicesGroup)[];
          })[]
        >(`${getChromeStaticPathname('services')}/services.json`)
      ).data,
    []
  );
```

src/hooks/useAllServices.ts
```diff
@@ -95,105 +105,119 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
-          // fallback static CSC for EE env
-          .catch(() => axios.get<BundleNavigation>(`${isBeta() ? '/beta' : ''}/config/chrome/${fragment}-navigation.json?ts=${Date.now()}`))
-          .then(handleBundleResponse)
-          .catch((err) => {
-            console.error('Unable to load appfilter bundle', err, fragment);
-            return [];
-          })
-      )
-    ).then((bundleItems) => {
+    const setNavigation = async () => {
```
Hyperkid123, line 171: Can we declare the function outside of the effect? The ref should be accessible within the whole component scope.

src/hooks/useAllServices.ts
```diff
@@ -95,105 +105,119 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
+    () =>
+      Promise.all(
+        bundles.map((fragment) =>
+          axios
+            .get<BundleNav>(`${getChromeStaticPathname('navigation')}/${fragment}-navigation.json?ts=${Date.now()}`)
+            .catch(() => axios.get<BundleNavigation>(`${isBeta() ? '/beta' : ''}/config/chrome/${fragment}-navigation.json?ts=${Date.now()}`))
+            .then(handleBundleResponse)
+            .catch((err) => {
+              console.error('Unable to load appfilter bundle', err, fragment);
+              return [];
```
Hyperkid123, line 155: This will mess up the function return type.

You want to probably add another then branch with a filter and a type guard
```ts
function isBundleNav(item: unknown): item is BundleNav {
  return typeof item !== 'undefined';
}

  const fetchNavitation = useCallback(
    () =>
      Promise.all(
        bundles.map((fragment) =>
          axios
            .get<BundleNavigation>(`${getChromeStaticPathname('navigation')}/${fragment}-navigation.json?ts=${Date.now()}`)
            .catch(() => axios.get<BundleNavigation>(`${isBeta() ? '/beta' : ''}/config/chrome/${fragment}-navigation.json?ts=${Date.now()}`))
            .then(handleBundleResponse)
            .catch((err) => {
              console.error('Unable to load appfilter bundle', err, fragment);
            })
        )
      ).then((data) => data.filter(isBundleNav)),
    []
  );
```

src/hooks/useAllServices.ts
```diff
@@ -95,105 +105,119 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
-            return [];
-          })
-      )
-    ).then((bundleItems) => {
+    const setNavigation = async () => {
+      const bundleItems = await fetchNavitation();
+      const sections = await fetchSections();
       if (isMounted.current) {
-        const availableLinks = parseBundlesToObject(bundleItems.flat());
+        const availableLinks = (bundleItems as BundleNav[]).map((bundle) => {
```
Hyperkid123, line 175: If you update the `fetchNavitation` you won't need the type cast here as the return type will always be `Promise<BundleNav[]>` instead of `Promise<never[] | BundleNav[]>`.

src/hooks/useAllServices.ts
```diff
@@ -95,105 +105,119 @@ const filterAllServicesSections = (allServicesLinks: AllServicesSection[], filte
...
+            if (links?.length === 0) {
+              return false;
+            }
+
+            return links.filter((item) => isAllServicesLink(item) || (isAllServicesGroup(item) && item.links.length !== 0)).flat().length !== 0;
+          });
         setState((prev) => ({
           ...prev,
-          availableLinks: bundleItems.flat(),
+          availableLinks: bundleItems as BundleNav[],
```
Hyperkid123, line 200: This typecast won't be required after `fetchNavitation` updates.

src/hooks/useFavoritedServices.ts
```diff
@@ -8,7 +9,9 @@ const useFavoritedServices = () => {
 
   // extract human friendly data from the all services data set
   const favoritedServices = favoritePages.reduce<ServiceTileProps[]>((acc, curr) => {
-    const service = servicesLinks.find(({ isExternal, href }) => !isExternal && href.includes(curr.pathname));
+    const service = servicesLinks.find(
+      (service) => !(service as AllServicesLink).isExternal && (service as AllServicesLink).href.includes(curr.pathname)
```
Hyperkid123, line 13: None of the new typecasts, including the find result are no longer required.

src/components/AllServices/AllServicesSection.tsx
karelhala: Ah, good point! I had to use it because we had string in the services link, but that's no longer the case. No extra guard is required

src/hooks/useAllServices.ts
karelhala: Ah, sweet! Updated the code based on your suggestion.

PR #2401, opened 2023-04-03
Consume CSC from chrome-service

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-25055

requires: https://github.com/RedHatInsights/chrome-service-backend/pull/33

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -314,14 +314,48 @@ export const trustarcScriptSetup = () => {
   document.body.appendChild(trustarcScript);
 };
 
-export const loadFedModules = () =>
-  axios.get(`${window.location.origin}${isBeta() ? '/beta' : ''}/config/chrome/fed-modules.json?ts=${Date.now()}`, {
-    headers: {
-      'Cache-Control': 'no-cache',
-      Pragma: 'no-cache',
-      Expires: '0',
+const CHROME_SERVICE_BASE = '/api/chrome-service/v1';
+export const chromeServiceStaticPathname = {
+  beta: {
+    stage: '/static/beta/stage',
+    prod: '/static/beta/prod',
+  },
+  stable: {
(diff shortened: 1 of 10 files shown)


src/utils/common.ts
```diff
@@ -314,14 +314,48 @@ export const trustarcScriptSetup = () => {
...
+
+export const loadFedModules = async () => {
+  const commnHeaders = {
+    'Cache-Control': 'no-cache',
+    Pragma: 'no-cache',
+    Expires: '0',
+  };
+  // we need to fetch both old and new to get the chrome config from FEO
+  const [FEOModules, staticModules] = await Promise.all([
+    axios.get(`${window.location.origin}${isBeta() ? '/beta' : ''}/config/chrome/fed-modules.json?ts=${Date.now()}`, {
```
karelhala, line 343: We are not getting the fed-modules config from FEO. I don't think we need this thing.

karelhala (commented):
Looking good! Just a minor question wrt getting fed-mods from FEO.

src/utils/common.ts
Hyperkid123: I think we still need this for EE right? For the SSO URL and stuff. And also console.dev

karelhala: Aaah, yes! For the ephemeral environment we'll need this. For the console.dev not anymore.

karelhala (approved):

Hyperkid123:
Turns out we can't use this yet because of 3scale auth gating. We need and exception.

Hyperkid123:
OK I have sequenced the data loading to ensure the ssoURL gets loaded from legacy config in necessary environments and after auth was sucesfull we load the modules config from chrome-service. This should also add a for qaprodauth.

karelhala (approved):

PR #2403, opened 2023-04-03
[RHCLOUD-24573] add support for extra props

BlakeHolifield:
Adds the ability for an app to pass in various `props` under `routes` in cloud-services-confg.

``` json
"learningResources": {
        "manifestLocation": "/apps/learningResources/fed-mods.json",
        "modules": [
            {
                "id": "learningResources",
                "module": "./RootApp",
                "routes": [
                    {
                        "pathname": "/settings/learning-resources",
                        "props": {
                            "bundle": "settings"
                        }
                    }
                ]
            }
        ]
    },
```

Can be loaded into learning resources via `props.bundle` in the AppEntry component.

### DIFF
--- a/src/chrome/create-chrome.ts
+++ b/src/chrome/create-chrome.ts
@@ -31,7 +31,7 @@ import chromeHistory from '../utils/chromeHistory';
 import { ReduxState } from '../redux/store';
 import { STORE_INITIAL_HASH } from '../redux/action-types';
 import { ChromeModule, FlagTagsFilter } from '../@types/types';
-import { createFedrampAuthObject } from '../cognito';
+import { createCognitoAuthObject } from '../cognito';
 import { getTokenWithAuthorizationCode } from '../cognito/auth';
 import useBundle from '../hooks/useBundle';
 
@@ -99,7 +99,7 @@ export const createChromeContext = ({
 
   const api: ChromeAPI = {
     ...actions,
-    auth: isITLessEnv ? createFedrampAuthObject() : createAuthObject(libJwt, getUser, store, modulesConfig),
+    auth: isITLessEnv ? createCognitoAuthObject(store) : createAuthObject(libJwt, getUser, store, modulesConfig),
     initialized: true,
     isProd,
     forceDemo: () => Cookies.set('cs_demo', 'true'),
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@BlakeHolifield nope we are good. Unless you are changing URLs. This can go in. but I'll try to get the CSC thing in early today.

src/@types/types.d.ts
```diff
@@ -127,6 +127,7 @@ export type RouteDefinition = {
   manifestLocation: string;
   dynamic?: boolean;
   exact?: boolean;
+  props?: any;
```
karelhala, line 130: We want to consume these props on two places? In the route definition[1] and in fed-modules route config[2]? It would be better to choose one so app developers are not confused which props they should use.

[1] https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/rhel-navigation.json#L6
[2] https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/fed-modules.json#L10

src/@types/types.d.ts
```diff
@@ -158,6 +160,12 @@ export type ChromeModule = {
   modules?: RemoteModule[];
   defaultDocumentTitle?: string;
   fullProfile?: boolean;
+  props?: ExtraProps | any;
```
karelhala, line 163: You should utilize the config field. We agreed with frontend operator on sending any additional data trough that field, so we don't have to add arbitrary fields.

src/@types/types.d.ts
```diff
@@ -135,6 +136,7 @@ export type ModuleRoute =
       pathname: string;
       exact?: boolean;
       dynamic?: boolean;
+      props?: any;
```
karelhala, line 139: Would be better to have Record<string, unknown>, unless we want to allow props: 'foobar' or props: 42.

karelhala (commented):
Looking good! Let's just choose the right place to store these extra props.

src/@types/types.d.ts
Hyperkid123: We should use the one just for routes. It's similar to how you would a component in a source code.

PR #2405, opened 2023-04-04
[RH-21442] - Help symbol button in the user dropdown closes the dropdown on click

InsaneZein:
https://issues.redhat.com/browse/RHCLOUD-21442


src/components/Header/UserToggle.tsx
```diff
@@ -122,8 +122,10 @@ const UserToggle = ({ isSmall = false, extraItems = [] }: UserToggleProps) => {
     };
   });
 
-  const onSelect = () => {
-    setIsOpen(!isOpen);
+  const onSelect = (event: any) => {
+    if (event.target.tagName === 'A') {
```
karelhala, line 126: We're also using `button` on the logout. Might be better to have `['BUTTON', 'A'].includes(event.target.tagName)` to cover both usecases.

karelhala (approved):

PR #2416, opened 2023-04-11
Use Jenkins env for current branch variable.

Hyperkid123:
@karelhala this will replace the static env variables we included for the hotfix. We can then proceed with moving to use chrome service as a source of CSC for the rest of the environments.

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -5,7 +5,6 @@ export IMAGE="quay.io/cloudservices/$COMPONENT"
 export WORKSPACE=${WORKSPACE:-$APP_ROOT}  # if running in jenkins, use the build's workspace
 export APP_ROOT=$(pwd)
 export INCLUDE_CHROME_CONFIG="true"
-export CHROME_CONFIG_BRANCH="prod-beta"
 cat /etc/redhat-release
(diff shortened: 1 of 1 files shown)


build_deploy.sh
```diff
@@ -14,6 +13,19 @@ COMMON_BUILDER=https://raw.githubusercontent.com/RedHatInsights/insights-fronten
...
+# get correct config build per build
+ENV_BRANCH=$(sed "s/origin\///" <<< "$GIT_BRANCH")
+CHROME_CONFIG_BRANCH=""
+if [[ $ENV_BRANCH == "master" ]]
+then
+  CHROME_CONFIG_BRANCH="prod-beta"
+else
+  CHROME_CONFIG_BRANCH=$(sed "s/master/prod/" <<< "$ENV_BRANCH")
+fi
+export CHROME_CONFIG_BRANCH=$CHROME_CONFIG_BRANCH
```
karelhala, line 27: I don't think you have to write it like this, it should be enough to export the branch
```suggestion
if [[ $ENV_BRANCH == "master" ]]
then
  export CHROME_CONFIG_BRANCH="prod-beta"
else
  export CHROME_CONFIG_BRANCH=$(sed "s/master/prod/" <<< "$ENV_BRANCH")
fi
```

karelhala (approved):

PR #2423, opened 2023-04-12
Enable multiple custom sso scopes per module.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-24910

depends on: https://github.com/RedHatInsights/chrome-service-backend/pull/43 to be in prod


karelhala (approved):
Does the config PR has to be merged before this one? It feels like both PRs can go in and it would be active once the config hits prod.

Hyperkid123:
@karelhala The config has to go in first. The chrome code no longer looks at the `fullProfile` flag but at the scopes from config.

PR #2445, opened 2023-04-20
Use visibility function in All services

karelhala:
### Description

Some apps don't want to surface in the all services if user doesn't have correct rights. This PR evaluates each bundle links and filters out those not available.

While testing API request visibility function I've found we are not sending the token to non HCC urls, so I've included it in all api visibility requests.

### DIFF
--- a/src/hooks/useAllServices.ts
+++ b/src/hooks/useAllServices.ts
@@ -10,6 +10,7 @@ import {
 } from '../components/AllServices/allServicesLinks';
 import { requiredBundles } from '../components/AppFilter/useAppFilter';
 import { getChromeStaticPathname, isBeta } from '../utils/common';
+import { evaluateVisibility } from '../utils/isNavItemVisible';
 
 export type AvailableLinks = {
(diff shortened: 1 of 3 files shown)


src/hooks/useAllServices.ts
```diff
@@ -160,6 +161,10 @@ const useAllServices = () => {
             .get<BundleNavigation>(`${getChromeStaticPathname('navigation')}/${fragment}-navigation.json?ts=${Date.now()}`)
             .catch(() => axios.get<BundleNavigation>(`${isBeta() ? '/beta' : ''}/config/chrome/${fragment}-navigation.json?ts=${Date.now()}`))
             .then(handleBundleResponse)
+            .then(async (bundleNav) => ({
```
Hyperkid123, line 164: A todo here to rewrite the function into async await. We are now mixing the two and we will have the possible edge case of eating error message within this branch.

Hyperkid123 (approved):

PR #2447, opened 2023-04-20
Supported nested navigation structures

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-25233

### TODO
- [x] highlight nested items
- [x] enable nesting in all services dropdown


src/hooks/useAllServices.ts
```diff
@@ -183,6 +189,7 @@ const useAllServices = () => {
           items: parseBundlesToObject(bundle.links?.flat()),
         };
       });
+      console.log({ availableLinks });
```
karelhala, line 192: No need for console.log
```suggestion
```

karelhala (approved):
Looking good! Just one extra console.log

PR #2456, opened 2023-04-25
Gate custom react router dom behind a check

karelhala:
### Description

When user is on hac application and clicks on anything in services drawer, and then navigates back with browser, they land on broken URL, because we are pushing in history stack non existent URL prepended with /hac.

This PR fixes such issue, by adding if statement to check if the requested URL contains hac application.

### JIRA

RHCLOUD-25258


src/chrome/update-shared-scope.ts
```diff
@@ -29,6 +36,9 @@ const updateSharedScope = () => {
         useNavigate: () => {
           const oldNavigate = reactRouter.useNavigate();
           return (to: To, options?: NavigateOptions) => {
+            console.log(options, 'this is options!');
+            console.log(to, 'this is to');
+            console.log(this, 'this is this');
```
Hyperkid123, line 41: Console logs.

Hyperkid123 (approved):

Issue #2461, opened 2023-04-26
Kiota headers get garbled


edewit:
When using Kiota in a console dot app the `iqeEnablement.ts` fetch replacement change the headers to something invalid. I can't 100% say what is the cause

This is what happens an array 
```
Access-Control-Request-Headers: 0,user-agent
```

Hyperkid123:
@edewit We automatically add auth header to any fetch/xhr. Can you give an example of where in the UI we can reproduce the issues and what would be the expected header vs the mangled one?

edewit:
@Hyperkid123 see the fix, the error is when the headers are in the form of `Record<string, sting[]>`

PR #2465, opened 2023-04-26
Fix fetch headers spread syntax in case of arrays

andreaTP:
Headers can be passed to fetch in various ways, but the current spread mechanism works only for `Object`s and fails with the [`Array` of `Array`s](https://developer.mozilla.org/en-US/docs/Web/API/Headers#examples) usage.

This is an attempt to fix the issue, but I'm struggling in testing this out, anyone can help?
cc. @riccardo-forina @edewit

### DIFF
--- a/src/utils/iqeEnablement.test.js
+++ b/src/utils/iqeEnablement.test.js
@@ -1,0 +1,20 @@
+import iqeEnablement from './iqeEnablement';
+
+describe('iqeEnablement', () => {
+  test('should correctly spread headers object', async () => {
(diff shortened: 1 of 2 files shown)


edewit:
fixes: #2461

andreaTP:
rebased on latest `master` and fixed the lining issues.

Hyperkid123 (approved):

Hyperkid123:
The changes should be in stage beta momentarily. We plan to deploy to other environments on Tuesday.

andreaTP:
Sounds perfect, we can check and follow up if something is still off 👍

PR #2468, opened 2023-04-27
Add akamaiCacheBustPaths for frontend resource

adamrdrew:
This patch adds paths for assets we want to cache bust on deploy.

I'm not 100% what branch y'all are using for contributions we want to get to stage-beta, so if this is the wrong branch let me know :)

### DIFF
--- a/frontend.yml
+++ b/frontend.yml
@@ -23,6 +23,10 @@ objects:
         paths:
           - /
           - /config/chrome
+      akamaiCacheBustPaths:
+        - /config/chrome/fed-modules.json
+        - /apps/chrome/index.html
+        - /apps/chrome/js/fed-mods.json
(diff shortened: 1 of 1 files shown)


frontend.yml
```diff
@@ -23,6 +23,10 @@ objects:
         paths:
           - /
           - /config/chrome
+      akamaiCacheBustPaths:
+        - /config/chrome/fed-modules.json
```
Hyperkid123, line 27: Just curious. How is `/beta` path handled? Is it dynamically prefixed somewhere or do we have to define it?

adamrdrew: Yup, beta and preview are added based on the frontened environment

Hyperkid123 (approved):

PR #2479, opened 2023-05-02
create landing page cards with pagination

InsaneZein:
favorites_landing_page


Hyperkid123 (commented):
The exposed module seems to be working and is consumable on the landing page. I am not sure about the empty state though:
screenshot-stage foo redhat com_1337-2023 05 02-16_12_21

InsaneZein:
I was just following the Sketch https://www.sketch.com/s/be45a74e-0565-4307-a926-de392d71d4d8/a/EeLa7yv#Inspect 
Is there a change you would suggest?

Hyperkid123:
Oh, it's supposed to look like that. OK then discard my comment. It just felt very tall when I saw it.

Hyperkid123:
Can you fix the eslint issues then? And the commented JSX? Other than that this looks very good.

src/components/RootApp/ScalprumRoot.tsx
```diff
@@ -201,6 +202,7 @@ const ScalprumRoot = memo(
             element={
               <Suspense fallback={LoadingFallback}>
                 <AllServices Footer={<Footer setCookieElement={setCookieElement} cookieElement={cookieElement} />} />
+                {/* <LandingNavFavorites /> */}
```
Hyperkid123, line 205: This is probably extra.

src/components/FavoriteServices/LandingNavFavorites.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <Gallery hasGutter className="hr-c-tile-landing">
+            {buildFavorites()}
+          </Gallery>
+        </React.Fragment>
+      )}
+    </React.Fragment>
+  );
+};
+
+export default LandingNavFavorites;
```
Hyperkid123, line 80: Missing EOL

src/components/FavoriteServices/LandingNavFavorites.scss
```diff
@@ -0,0 +1,25 @@
...
+.hr-c-tile-landing {
+  --pf-l-gallery--m-gutter--GridGap: 10px !important;
+}
+
+.chr-c-icon-favorites {
+  padding-right: 18px !important;
+}
+
+.pf-c-options-menu__toggle {
+  display: none !important;
```
karelhala, line 19: This will remove every toggle, which is not good. Any reason for it?

src/components/FavoriteServices/LandingNavFavorites.scss
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.landing-all-services-link {
+  padding-left: 18px !important;
```
karelhala, line 23: Same as with other spacing, let's use it in landing page and discuss it with UXD.

src/components/FavoriteServices/LandingNavFavorites.scss
```diff
@@ -0,0 +1,25 @@
...
+  font-size: var(--pf-global--FontSize--md);
```
karelhala, line 24: No need to set custom font size, it should be handled by the component itself.

src/components/FavoriteServices/LandingNavFavorites.scss
karelhala, line 1: I wouldn't use any of these stylings as it adds extra spacing etc. Moreover the `display: none` will break any dropdown.

karelhala:
Looking good! Except for the custom styling, I wouldn't use it at all.

src/components/FavoriteServices/LandingNavFavorites.scss
InsaneZein: This is to remove the dropdown for the pagination since its not needed. I saw no other way to do this. Would this be ok if I put the pagination in its own class?

InsaneZein: @karelhala does the class I added in pagination fix that problem?

Hyperkid123 (approved):

PR #2490, opened 2023-05-04
Use longer fragment matching to get better breadcrumbs links results.

Hyperkid123:
Improves breadcrumb fragment HREFs when URL fragments do not match the nav levels (there is more fragments in URL than there is levels in left nav)


karelhala (approved):

gkarat:
Also fixes https://issues.redhat.com/browse/OCPADVISOR-90.

PR #2521, opened 2023-05-19
Make left and right sections of new nav scrollable

epwinchell:
JIRA: https://issues.redhat.com/browse/RHCLOUD-25357

- two scrollbar regions
- Panel height set to 80vh
- "Browse all services" fixed to bottom

Screenshot 2023-05-22 at 11 44 45 AM

### DIFF
--- a/src/components/AllServicesDropdown/AllServicesMenu.tsx
+++ b/src/components/AllServicesDropdown/AllServicesMenu.tsx
@@ -68,8 +68,8 @@ const AllServicesMenu = ({ setIsOpen, isOpen, menuRef, linkSections, favoritedSe
       <Backdrop>
         <Panel variant="raised" className="pf-u-p-0 chr-c-panel-services-nav">
           <PanelMain>
-            <Sidebar>
-              <SidebarPanel>
+            <Sidebar style={{ height: '698px', overflow: 'auto' }}>
+              <SidebarPanel variant="sticky">
                 <Flex className="pf-u-flex-direction-column pf-u-flex-grow-1">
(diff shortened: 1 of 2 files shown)


src/components/AllServicesDropdown/AllServicesMenu.tsx
```diff
@@ -68,8 +68,8 @@ const AllServicesMenu = ({ setIsOpen, isOpen, menuRef, linkSections, favoritedSe
       <Backdrop>
         <Panel variant="raised" className="pf-u-p-0 chr-c-panel-services-nav">
           <PanelMain>
-            <Sidebar>
-              <SidebarPanel>
+            <Sidebar style={{ height: '698px', overflow: 'auto' }}>
```
Hyperkid123, line 71: I think we might have to compute the height value based on the height of the screen. Otherwise, we will get into some double scrollbar weirdness.

Screenshot from 2023-05-23 09-09-04

epwinchell: @Hyperkid123 I'm only seeing the double scrollbar when the viewport height is less than 850px and it's a non-issue in mobile. Is this an edge case?

Hyperkid123:
Not sure of this is just mu OS but I see the scrollbars all the time now:
screenshot-stage foo redhat com_1337-2023 06 01-09_24_45

epwinchell:
@Hyperkid123 What OS & browser?

epwinchell:
@Hyperkid123 Updated

Hyperkid123 (approved):

PR #2523, opened 2023-05-22
Added Patternfly user feedback extension

dlabaj:
Added patternfly user feedback extension.  This was done to standardize on the user feedback modal.  Code originated from the feedback modal that was originally created in this repo.


aneelac22 (approved):
/retest

Hyperkid123:
@dlabaj Is it possible to compose the modal from the composite component? I don't really like the way the text is propagated to the component or the fact that I can't really customize the layout, actions and other aspects of the individual forms. Is there way of some "composable" approach to use the feedback component? I think this might become an issue if the future if we have to make some quick modifications to the content.

This is a very rough and quick draft of an API I'd like to be able to use:

```jsx
<FeedbackModal feedbackImg={feedbackImage} defaultActions={[<FeedbackCloseButton>Close</FeedbackCloseButton>, <Button>Some custom action</Button>]} {...restOfModalProps}>
  {/* the clickable tile */}
  <FeedbackSection id="custom-feedback" title="Share feedback" description="Share What has your console......">
   {/* child is displayed if the section with the id is active */}
    <FeedbackForm
      type="custom"
      onSubmit={submitFeedback}
      actionButtons={[
        // calls the on submit
        <FeedbackSubmitButton>Foo</FeedbackSubmitButton>,
        // goes back to the initial modal view
        <FeedbackBackButton>Bar</FeedbackBackButton>,
        // closes the modal
        <FeedbackCancelButton>Baz</FeedbackCancelButton>,
      ]}
    >
      {/* state in these would be handled externally by the developer */}
      <TextArea label="Enter your feedback" />
      <Checkbox label="Yes, I would like to...." />
    </FeedbackForm>
    {/* have some pre-defined forms with internal state management*/}
    {/* either have type */}
    <FeedbackForm type="some-pre-defined-type" formLabels={{...}} onSubmit={...} />
    {/* or an actual component */}
    <FeedBackFormPredefined onSubmit={...} title={intl.formatMessage(....)} otherText={intl.formatMessage(....)}/>
    ...
  </FeedbackSection>
</FeedbackModal>
```

I think we can also have the "god" component that does everything. But with this approach, we will be covered for 100% of the use cases. And the app will not bundle any extra code that they do not need in their bundles.

dlabaj:
@Hyperkid123 Do you mind waiting for v5 for the api updates? Would you take this for v4 with the current api?

Hyperkid123:
@dlabaj I am OK waiting for v5

PR #2527, opened 2023-05-24
added missing space

petracihalova:
image

### DIFF
--- a/src/components/Feedback/FeedbackModal.tsx
+++ b/src/components/Feedback/FeedbackModal.tsx
@@ -153,7 +153,7 @@ const FeedbackModal = memo(({ user }: FeedbackModalProps) => {
             modalTitle={intl.formatMessage(messages.informRedhatDirection)}
             modalDescription={
               <Text>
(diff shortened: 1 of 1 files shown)


src/components/Feedback/FeedbackModal.tsx
```diff
@@ -153,7 +153,7 @@ const FeedbackModal = memo(({ user }: FeedbackModalProps) => {
             modalTitle={intl.formatMessage(messages.informRedhatDirection)}
             modalDescription={
               <Text>
-                {intl.formatMessage(messages.informDirectionDescription)}
+                {intl.formatMessage(messages.informDirectionDescription)}{' '}
```
Hyperkid123, line 156: Although the empty string will work, it is very vulnerable to being deleted as this exact code is sometimes incorrectly injected by code formatters. It will be safer to use the HTML character code for a space: `&nbsp;`. We will never remove it.

Hyperkid123 (approved):

PR #2534, opened 2023-06-01
introduces sentry webpack plugin for sourcemap uploading to sentry

mkholjuraev:
This enables uploading sourmaps generated by the `hidden-source-map` type of devtool to Sentry. `hidden-source-map` helps not to upload sourcemaps to browser according to the [doc](https://webpack.js.org/configuration/devtool/#production) . I used [@sentry/webpack](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/webpack/) plugin to use a new method introduced by the sentry to upload sources as artifacts instead of releases. 

The changes only prod-beta env for testing purposes. We need to create a new travis secret `SENTRY_AUTH_KEY`  before merging. I believe I do not have rights for this.


.travis.yml
```diff
@@ -35,6 +35,8 @@ env:
   global:
     - REPO="git@github.com:RedHatInsights/insights-chrome-build"
     - REPO_DIR="insights-chrome-build"
+    - BRANCH=${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}
+    - SENTRY_AUTH_KEY=${SENTRY_AUTH_KEY}
```
Hyperkid123, line 39: We do not use Travis anymore to publish our assets. This has to be set up in the vault.

config/webpack.config.js
```diff
@@ -51,7 +51,7 @@ const commonConfig = ({ dev }) => {
           },
         }
       : {}),
-    devtool: false,
+    devtool: process.env.BRANCH === 'prod-beta' ? 'hidden-source-map' : false,
```
Hyperkid123, line 54: @karelhala do you remember if we have this env variable in the containerized build?

.travis.yml
mkholjuraev: Thank you. I do not know what is vault. Can you some guidance on what is it and how I can provide the secret to node env so that I can fix this?

Hyperkid123: I think you won't be able to do it because it is the platex store. If you PM me the secret, I will ensure it will be added to the vault store.

PR #2539, opened 2023-06-05
[RHCLOUD-26032] Help drop-down has duplicated data-ouia-component-id

rsun19:
- Used hook and event listener to dynamically show `<HeaderTools />` during window resizes. The ouia chrome-help does not duplicate anymore.

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -44,7 +44,7 @@ export const Header = ({ breadcrumbsProps }: { breadcrumbsProps?: Breadcrumbspro
   const isITLessEnv = ITLess();
   const { pathname } = useLocation();
   const noBreadcrumb = !['/', '/allservices', '/favoritedservices'].includes(pathname);
-  const { lg } = useWindowWidth();
+  const lg = useWindowWidth();
 
   return (
     <Fragment>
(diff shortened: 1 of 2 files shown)


src/components/Header/Header.tsx
```diff
@@ -44,6 +44,21 @@ export const Header = ({ breadcrumbsProps }: { breadcrumbsProps?: Breadcrumbspro
   const { pathname } = useLocation();
   const noBreadcrumb = !['/', '/allservices', '/favoritedservices'].includes(pathname);
 
+  const useWindowWidth = () => {
```
Hyperkid123, line 47: Can we have a separate file for this hook? It might be useful in other places later on. There is a hook directory with some files already.

Hyperkid123 (commented):
Looks good, just one small request.

src/hooks/useWindowWidth.tsx
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react';
+
+const useWindowWidth = () => {
+  const [lg, setLg] = useState(window.innerWidth >= 1450);
+
+  useEffect(() => {
+    const handleResize = () => {
+      setLg(window.innerWidth >= 1450);
+    };
+    window.addEventListener('resize', handleResize);
```
karelhala, line 10: You should also be nice and add `window.removeEventListener`
```suggestion
    window.addEventListener('resize', handleResize);
    () => window.removeEventListener('resize', handleResize);
```

src/hooks/useWindowWidth.tsx
```diff
@@ -0,0 +1,16 @@
...
+  }, []);
+
+  return { lg };
```
karelhala, line 13: Do we want to return the boolean value in object? Wouldn't it be better to return it directly?

rsun19:
Good catch --- thanks!

src/hooks/useWindowWidth.tsx
Hyperkid123: We might want other breakpoint values in the future not just lg.

karelhala (approved):

PR #2540, opened 2023-06-05
Use session to store login creds

karelhala:
### Description

We don't have to wait each time we want to test something for login, let's store the actual login creds in session and share between all tests.

### JIRA

RHCLOUD-26213


fhlavac (approved):

Hyperkid123 (approved):
Just out of curiosity. I assume there is a way to bypass this? If we for some reason need a fresh session or a different account.

PR #2546, opened 2023-06-07
RHCLOUD-26169	[CYPRESS] Add test to add things to favorites

rsun19:
- As requested, intercepted `/user` and `/favorited-service` requests.
- Tested using the API Management block.
- Used control flow. If POST request isn't intercepted (meaning favorites is broken), the script throws an exception.
- Let me know if there's something that needs to be in there that I didn't add.
- Update: added assertions for starring and seeing if it has been added to favorites
--> also switched to use All Services Page to make testing simpler and more reliable!


karelhala:
@rsun19 looking good! Can you fix the failing lint errors? Once they are green we can merge this!

rsun19:
Fixed lint issues!

karelhala (commented):
Looking good! I just have a question about the static responses. We want to test E2E with these tests, it's expected to take a while and I think that keeping these static responses would effectively beat the purpose of these tests.

rsun19:
> Looking good! I just have a question about the static responses. We want to test E2E with these tests, it's expected to take a while and I think that keeping these static responses would effectively beat the purpose of these tests.

Unless I misunderstood Martin, I think he said to intercept the requests. Maybe we can ask him when he gets back

Hyperkid123:
@karelhala yeah, I am worried about us interfering with the other e2e tests. We will need a unique account before we can start making this persistent.

karelhala (approved):

PR #2549, opened 2023-06-08
RHCLOUD-26168 [CYPRESS] Add test to load landing page

rsun19:
Intensive testing that takes ~22 seconds to complete. I tried to make it as maintainable as possible by including variables in case the test needs to change in the future.

OVERVIEW:
1. Intercepts last visited links & favorited services
2. Cross-references last visited service names, favorited services, and clicks on "next button" for favorited services.
3. Counts number of slides in DOM, tests the "next" button next to the slides
4. Cross references & counts the service names under `Get started with Hybrid Cloud Console capabilities`.

Let me know if there are any other testing I need to do.


karelhala:
Looking hella good! Can you fix the failing lint test?

rsun19:
Fixed lint issues!

Also added `cy.visit('/');` because in some instances, the window hangs after reload. This mitigates it.

cypress/e2e/release-gate/landing-page.cy.ts
```diff
@@ -1,17 +1,250 @@
...
+          'application-services': true,
+        },
+      },
+    });
+    cy.wait(3000);
+    cy.contains('API Management').should('exist');
+    cy.get('.pf-c-button.pf-m-plain').eq(1).click();
+    for (let i = 0; i < FAVORITED_OVERFILLS.length; i++) {
+      cy.get('.chr-c-favorite-service__tile').eq(i).find('.pf-u-mb-sm').should('contain', FAVORITED_OVERFILLS[i]);
+    }
```
karelhala, line 224: No need to use `for`, let's use `forEach` for better readability
```suggestion
    FAVORITED_OVERFILLS.forEach((item, idx) => {
      cy.get('.chr-c-favorite-service__tile').eq(idx).find('.pf-u-mb-sm').should('contain', item);
    })
```

cypress/e2e/release-gate/landing-page.cy.ts
```diff
@@ -1,17 +1,250 @@
...
     cy.contains('View my favorite services').should('exist');
+    cy.contains('Recently visited').should('exist');
+    for (let i = 0; i < lastVisitedArray.length; i++) {
```
karelhala, line 227: Same as above, `for` is less readable then `forEach`.

karelhala (commented):
Looking good, just changing the `for` loop over to `forEach` might be a better fit. I am trying to think if this doesn't kill the e2e testing if we'd keep the JSONs static for each test. We want to test the full stack with these tests and it beats the purpose if we're going to keep the JSONs static.

rsun19:
Let me know what you decide. The test will probably increase in time dramatically if JSON is not static

Hyperkid123:
@karelhala 
>  I am trying to think if this doesn't kill the e2e testing if we'd keep the JSONs static for each test. We want to test the full stack with these tests and it beats the purpose if we're going to keep the JSONs static.

We will need a custom account for this. We are currently using the common qa account. Meaning that any changes we make, will have a persistent impact which might mess with other e2e testing that is happening. I'd say, let's switch but only once we have a unique account for these tests.

rsun19:
Fixed merge conflicts

cypress/e2e/release-gate/landing-page.cy.ts
```diff
@@ -1,8 +1,258 @@
...
+    }
+    for (let i = 0; i < NUM_SLIDE_ELEMENTS - 1; i++) {
+      cy.get('.slick-arrow.slick-next').should('exist');
+      cy.wait(1000);
+      cy.get('.slick-arrow.slick-next').click();
+    }
+    cy.get('.slick-arrow.slick-next.slick-disabled').should('exist');
+    for (let i = 0; i < FRONT_PAGE_SERVICES.length; i++) {
+      cy.get('.pf-l-gallery.pf-m-gutter')
+        .eq(1)
```
catastrophe-brandon, line 242: Could you elaborate on the meaning of `.eq(1)` here?

cypress/e2e/release-gate/landing-page.cy.ts
```diff
@@ -16,9 +266,10 @@ describe('Landing page', () => {
   });
 
   it('tooltip is shown when hovering over the gear/question icon', () => {
+    cy.visit('/');
     cy.login();
-
     cy.visit('/');
+    cy.reload();
```
catastrophe-brandon, line 272: Why is a reload necessary here? A one-line comment might help explain.

cypress/e2e/release-gate/landing-page.cy.ts
```diff
@@ -1,8 +1,258 @@
+// Declared variables to increase reusability if making breaking change!
+const lastVisitedJSON =
```
catastrophe-brandon, line 2: Is this an e2e test or a component test? If it's a component-level test then hard-coded JSON is probably fine. For an e2e test, I would expect the data to be produced through user interactions with the page.

cypress/e2e/release-gate/landing-page.cy.ts
rsun19: For reference for anyone who reads this, my internship ended so I can't change any of this if I wanted to... That being said, the `.reload()` is there because sometimes, when not reloaded, the DOM tree fails to render correctly, which fails some of the tests

cypress/e2e/release-gate/landing-page.cy.ts
rsun19: If I remember correctly, there are multiple elements named `.pf-l-gallery.pf-m-gutter`, so the `.eq(1)` is there to tell Cypress that we are looking for the second element of `.pf-l-gallery.pf-m-gutter` that appears on the page

cypress/e2e/release-gate/landing-page.cy.ts
rsun19: I had this question before. If nothing changed from then, read Martin's comment above and ping him if it should be done differently --- "We will need a custom account for this. We are currently using the common qa account. Meaning that any changes we make, will have a persistent impact which might mess with other e2e testing that is happening. I'd say, let's switch but only once we have a unique account for these tests."

cypress/e2e/release-gate/landing-page.cy.ts
rsun19: Maybe there's a better way to do this, but I am not sure

cypress/e2e/release-gate/landing-page.cy.ts
Hyperkid123: @catastrophe-brandon we need a new testing account. We have run into issues that were false negative results because during the test run someone else was using the testing account in the same area. Once we have one, we can remove the intercept definitions.

karelhala:
Closing in favor of https://github.com/RedHatInsights/insights-chrome/pull/2668

PR #2559, opened 2023-07-03
Updated the file to add Attributes functionality to App Studio in-app help file

gtrivedi88:
**JIRA:** [HACDOCS-461](https://issues.redhat.com/browse/HACDOCS-461)

### DIFF
--- a/src/components/RootApp/RootApp.tsx
+++ b/src/components/RootApp/RootApp.tsx
@@ -124,6 +124,21 @@ const RootApp = memo((props: RootAppProps) => {
           </Suspense>
           <ChromeProvider bundle={bundleTitle}>
             <QuickStartContainer {...quickStartProps}>
+              const drawerProps: QuickStartContainerProps = {
+                markdown: {
+                    extensions: [
+                      // Variable substitution example
+                      // This replaces the strings [PROJECT_NAME]
+                      {
+                        type: 'output',
+                        filter: function(html: string) {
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@gtrivedi88 I like the idea. I think we also want some way to narrow down if we want to use these. And possibly have the parser defined in a more distributed way. So modules/topics can define them on their own.

I am good with this as a first pass, but I'd like to explore these two points:

1. Add the replace regular expressions to the help topic YAML file.
2. Have the original topic source as a second argument within the markdown extension filter function and let the runtime code pull out the regular expressions from the helptopic source.

I think with these, we can provide solid formatting options and avoid potential conflicts between different help topic definitions and UI modules.

src/components/RootApp/RootApp.tsx
```diff
@@ -124,6 +124,22 @@ const RootApp = memo((props: RootAppProps) => {
...
             <QuickStartContainer {...quickStartProps}>
+              const drawerProps: QuickStartContainerProps = {
+                markdown: {
+                    extensions: [
+                      // Variable substitution example
+                      // This replaces the strings [PROJECT_NAME]
+                      {
+                        type: 'output',
+                        filter: function(html: string) {
+                          html = html.replace(/\[PRODUCT_NAME\]/g, 'App Studio');
```
dlabaj, line 135: html.replace(/\[(.+)]{{(product_name) ([\w-]+)}}/g, ...) over here.  Let's see what we need to get the product_name from the backend in your other PR.  https://github.com/gtrivedi88/quickstarts/pull/1

PR #2576, opened 2023-07-18
[RHINENG-663] - Add subtitle property in AllServices

AsToNlele:
[RHINENG-663](https://issues.redhat.com/browse/RHINENG-663)

Is it possible to use the `product` field for the subtitle here?
image
At first I tried changing the bundle names, but that would not work for some ansible items

https://github.com/RedHatInsights/chrome-service-backend/pull/228 depends on this PR

### DIFF
--- a/src/components/AllServicesDropdown/AllServicesGalleryLink.tsx
+++ b/src/components/AllServicesDropdown/AllServicesGalleryLink.tsx
@@ -12,7 +12,7 @@ import { AllServicesDropdownContext } from './common';
 
 export type AllServicesGalleryLinkProps = AllServicesLinkProps;
 
-const AllServicesGalleryLink = ({ href, title, description, isExternal }: AllServicesGalleryLinkProps) => {
+const AllServicesGalleryLink = ({ href, title, description, isExternal, subtitle }: AllServicesGalleryLinkProps) => {
   const bundle = bundleMapping[href.split('/')[1]];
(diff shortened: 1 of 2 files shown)


Hyperkid123 (changes requested):
> Is it possible to use the product field for the subtitle here?

@AsToNlele unfortunately we can't as it would impact other tiles that do want to use the product as a subtitle and have the product defined in their links. I would suggest creating a new attribute inside the Chrome service as that instead.

AsToNlele:
I've added `subtitle` property

src/components/AllServicesDropdown/AllServicesGalleryLink.tsx
```diff
@@ -68,8 +68,12 @@ const AllServicesGalleryLink = ({ href, title, description, isExternal }: AllSer
...
           </Split>
           <TextContent>
-            {/* do not show bundle if the card title matches bundle title */}
-            <Text component="small">{bundle !== title ? bundle : null}</Text>
+            {/* 
+              if subtitle is not set use bundle
+
+              do not show bundle if the card title matches bundle title
+            */}
+            <Text component="small">{subtitle ? subtitle : bundle !== title ? bundle : null}</Text>
```
Hyperkid123, line 76: Can you change this to not use multiple ternary statements? To something like
```JS
subtitle || (bundle !== title ? bundle : null)
```

Hyperkid123 (approved):

PR #2588, opened 2023-07-21
jenkinsfile frontend pipeline poc

maknop:


Hyperkid123:
@maknop I Need to double-check. Is this going to replace the current Jenkins jobs?

maknop:
@Hyperkid123 At the moment this will just add a second one that will run. The goal here is not to have it actually deploy anything yet but simply get all the other stages in place.

PR #2592, opened 2023-07-26
Add explicit search query option.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-27216

In addition to a wildcard search option add some variants and also an explicit term query to bypass the expected character. Gives higher priority to the explicit search matches.

Primarily deals with issues when an exact query does not match the search term.

### DIFF
--- a/src/components/Search/SearchInput.tsx
+++ b/src/components/Search/SearchInput.tsx
@@ -36,7 +36,7 @@ const FUZZY_RANGE_TAG = 'FUZZY_RANGE_TAG';
  */
 
 const BASE_SEARCH = new URLSearchParams();
-BASE_SEARCH.append('q', `${REPLACE_TAG}*~${FUZZY_RANGE_TAG}`); // add query replacement tag and enable fuzzy search with ~1
+BASE_SEARCH.append('q', `*${REPLACE_TAG}~${FUZZY_RANGE_TAG} OR ${REPLACE_TAG}*~${FUZZY_RANGE_TAG} OR ${REPLACE_TAG}~${FUZZY_RANGE_TAG}`); // add query replacement tag and enable fuzzy search with ~ and wildcards
 BASE_SEARCH.append('fq', 'documentKind:ModuleDefinition'); // search for ModuleDefinition documents
 BASE_SEARCH.append('rows', '10'); // request 10 results
 BASE_SEARCH.append('hl', 'true'); // enable highlight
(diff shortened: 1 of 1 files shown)


src/components/Search/SearchInput.tsx
```diff
@@ -196,7 +196,7 @@ const SearchInput = () => {
   }, [isOpen, menuRef]);
 
   const handleFetch = (value = '') => {
-    return fetch(SEARCH_QUERY.replace(REPLACE_TAG, value).replace(FUZZY_RANGE_TAG, value.length > 3 ? '2' : '1'))
+    return fetch(SEARCH_QUERY.replaceAll(REPLACE_TAG, value).replaceAll(FUZZY_RANGE_TAG, value.length > 3 ? '2' : '1'))
```
vkrizan, line 199: Could the search query build be moved to a separate function?

Can the URL prams be build safely using a URL builder instead of search and replace?

Hyperkid123: The query is built with `URLSearchParams` and then the URL is built with `URL` classes on the top of the file. The search endpoint currently **does not allow encoded query string**. It had to be disabled a few months back. Therefore there is no point in rebuilding the query on each request and the creating was lifted out of the callback to be static and save some resources.

Currently, the replace method is as good as using the search params constructor.

vkrizan: Okay, I've missed the use of `URLSearchParams`.

Do we escape [special characters](https://solr.apache.org/guide/7_3/the-standard-query-parser.html#escaping-special-characters) within the search query?

vkrizan: Which API does not support encoded URI (query string)? Is that SOLR? Query string is part of the standard of RFC3986.

vkrizan: Found the references:
* https://github.com/RedHatInsights/insights-chrome/pull/2413
* https://issues.redhat.com/browse/RHCLOUD-25116

@Hyperkid123 do you have more info on the reason or can you ask the search team please? Was it reported to them?

vkrizan: I'm trying to send encoded ampersand `%26` to stage API, it does not fail. However, not encoding these characters leads to HTTP 400.

Hyperkid123: I do not have any formal records. Discussion with the search team happened on a chat. I understand the encoded strings are part of the spec. The hydra API was returning 400 responses when we tried to send encoded queries.

florkbr: Discussed with Martin offline. IMO we will need to follow up on the URL encoding in a separate pull request - that way we can revert either (or both) PRs if there is an issue. It seems the search team fixed an issue with the Hydra REST endpoint that was originally causing the 400's ([see slack thread](https://redhat-internal.slack.com/archives/C04KBMF9Z5W/p1681187623521589?thread_ts=1680765196.170799&cid=C04KBMF9Z5W)). We should now be able to send the encoded special characters in the SOLR queries.

florkbr (approved):

PR #2593, opened 2023-07-26
Build Deploy Frontends Jenkinsfile Effort

maknop:


ci/unit_tests.sh
```diff
@@ -1,18 +1,13 @@
...
-source ./ci/helpers.sh
-
 PROJECT_NAME="insights-chrome"
 TEST_CONT="${PROJECT_NAME}-unit-tests"
 IMG_TAG=$(git rev-parse --short=8 HEAD)
 
-docker build -t "${TEST_CONT}:${IMG_TAG}" -f Dockerfile.test .
+docker run --name "${TEST_CONT}-${IMG_TAG}" -d -i "${NODE_BASE_IMAGE}" /bin/sh
+
+docker cp . "${TEST_CONT}-${IMG_TAG}":.
```
Victoremepunto, line 9: when copying files with `cp` you have to specify an absolute path, hence now they are copying into the container's root, where the user doesn't have permissions. Adding the prefix `/opt/app-root/src/` should do the trick.

the same goes for the `exec` command, you can specify where the command will be run with the `-w` flag.

ci/cypress.sh
```diff
@@ -0,0 +1,15 @@
...
+
+PROJECT_NAME="insights-chrome"
+TEST_CONT="${PROJECT_NAME}-cypress"
+IMG_TAG=$(git rev-parse --short=8 HEAD)
+
+docker login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
+
+docker run --name "${TEST_CONT}-${IMG_TAG}" -d -i "${CYPRESS_TEST_IMAGE}" /bin/bash
+
+docker cp . "${TEST_CONT}-${IMG_TAG}:/e2e/"
```
Victoremepunto, line 11: let's add

```
docker exec -i "${TEST_CONT}-${IMG_TAG}" sh -c "ls -lrd ."
docker exec -i "${TEST_CONT}-${IMG_TAG}" sh -c "ls -l *"
```

maknop:
retest

PR #2608, opened 2023-08-08
Remove cookie from ws:/ and wss:/ paths

karelhala:
### Description

When cookie is set for `ws:/` or `wss:/` it's automatically enabled for any request. This breaks some APIs as they are confused which token parse (either cooki or auth token). This PR removes these 2 endpoints since we are using `/ws` and `/wss` paths anyway for sending tokens to websocket connections.

### JIRA

RHCLOUD-27537


apinkert (approved):
👍

Hyperkid123:
@karelhala I'll have to verify first. I remember there being some issues with the protocols not being covered.

Hyperkid123 (approved):

PR #2609, opened 2023-08-11
Integrate Keycloak in int and scr env

aneelac22:


src/utils/common.ts
```diff
@@ -38,10 +38,25 @@ export const DEFAULT_SSO_ROUTES = {
...
+    portal: 'https://console.stage.openshiftusgov.com',
+  },
+  ephem: {
     url: ['ephem.outsrights.cc'],
-    sso: 'https://stage-gov-console.auth.us-east-1.amazoncognito.com/login',
+    sso: 'https://keycloak-fips-test.apps.fips-key.2vn8.p1.openshiftapps.com',
     portal: 'https://ephem.outsrights.cc/',
   },
+  int: {
+    url: ['console.int.openshiftusgov.com'],
```
bastilian, line 51: Would it not be better to allow these URLs to be configured when building the assets via env variables and avoid exposing URLs like these on GitHub?

Hyperkid123: @bastilian we configure the SSOUrl for our environments in the frontend operator and the values are not public. However, the specific environment of this PR is not yet using the frontend operator. It's in the process of migration. The same applies to the qaproduaht. Until qaprodauth is discontinued and all envs are using the FEO, we need these values still.

src/jwt/jwt.ts
```diff
@@ -134,7 +136,7 @@ export const doOffline = (key: string, val: string, configSsoUrl?: string) => {
       scopes.push(partnerScope);
     }
 
-    if (ssoScopes) {
+    if (ssoScopes && !itLessKeycloakEnv) {
```
Hyperkid123, line 139: Just a double-check. The offline token is not supposed to have all SSO scopes of the current session?

Hyperkid123 (approved):
LGTM, there is just a small conflict due to the pf migrations.

PR #2615, opened 2023-08-18
Update fetch wrapper in iqeEnablement

lucasgarfield:
The wrapper around fetch in iqeEnablement.ts was dropping headers if a Request object was passed in as the input argument to fetch. This commit ensures that if a Request object is passed that the headers are preserved.

Note that this commit does NOT necessarily ensure full compatibility with the fetch API... it simply takes one step in that direction by making sure the headers are preserved.

### DIFF
--- a/src/utils/iqeEnablement.ts
+++ b/src/utils/iqeEnablement.ts
@@ -103,9 +103,25 @@ function init(store: Store, libJwt?: () => LibJWT | undefined) {
    * Check response errors for cross_account requests.
    * If we get error response with specific cross account error message, we kick the user out of the corss account session.
    */
-  window.fetch = function fetchReplacement(path = '', options, ...rest) {
+  window.fetch = function fetchReplacement(path: URL | RequestInfo = '', options, ...rest) {
     const tid = Math.random().toString(36);
     const additionalHeaders: any = spreadAdditionalHeaders(options);
+
+    /**
+     * If path is a Request object, it may contain headers. Those headers will be overwritten, not merged,
+     * by the headers property containing the authorization headers and additionalHeaders in options. Therefore,
+     * we need to merge any headers present in the Request object into additionalHeaders.
+     */
+
+    function isRequest(p: URL | RequestInfo): p is Request {
+      return !(p instanceof URL) && typeof p !== 'string';
+    }
+
+    if (isRequest(path)) {
+      for (const pair of path.headers.entries()) {
+        additionalHeaders[pair[0]] = pair[1];
+      }
+    }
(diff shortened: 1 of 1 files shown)


src/utils/iqeEnablement.ts
```diff
@@ -103,10 +103,26 @@ function init(store: Store, libJwt?: () => LibJWT | undefined) {
...
     const additionalHeaders: any = spreadAdditionalHeaders(options);
 
+    /**
+     * If path is a Request object, it may contain headers. Those headers will be overwritten, not merged,
+     * by the headers property containing the authorization headers and additionalHeaders in options. Therefore,
+     * we need to merge any headers present in the Request object into additionalHeaders.
+     */
+
+    function isRequest(path: URL | RequestInfo): path is Request {
+      return (path as Request).headers !== undefined;
```
lucasgarfield, line 117: I'm still a bit new to Typescript and I'm not sure if this is the best way to implement the type guard but it seems to work.

Hyperkid123: The `RequestInfo` type itself is defined as `type RequestInfo = Request | string`. We can use that in combination with the `instanceof` operator to get the correct type without any typecasting.

```ts
function isRequest(p: URL | RequestInfo): p is Request {
  return !(p instanceof URL) && typeof p !== 'string';
}
```

Your code will work, but this one is a bit safer in the event that for whatever reason someone assigns the `headers` property to the URL instance which is possible in JS. (I am sure there is a nicer way to write the condition)

lucasgarfield:
I didn’t put very much thought into the fetch wrapper API, this PR was a proof of concept and feels just a bit hacky to me.

Would it make more sense to try to keep things closer to the fetch API? For instance, if a RequestInfo object is passed in, we could add the authorization and any additional headers to the RequestInfo object and then call oldFetch using the RequestInfo object. That way any other properties on the RequestInfo object are also preserved. 

What do you think, @Hyperkid123? Would you be willing to merge this as is or would you prefer to try to more closely emulate the fetch API in case a RequestInfo object is passed in?

Hyperkid123:
Ideally, we would "do it the other way around". Meaning that we always expect the argument to be the `RequestInfo` object. And if the argument is `string` or `URL` we would create a new one and use the argument as a parameter in the `RequestInfo` object.

That way setting the headers would not feel so hacky. Although it will always feel like that at least a little bit. The issue with the current implementation is that we expect the argument to be string, which is a wrong assumption.

Hyperkid123 (approved):

PR #2620, opened 2023-08-22
Fix inventory API issues

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-27815

- Runtime error caused by an older version of the migrated package.
- API error caused by improper axios query parameters parsing?? We will have to re-visit the generated clients.


karelhala (approved):

Hyperkid123:
SSo is down

PR #2629, opened 2023-08-30
Hook notifications drawer to chrome-services websockets

aneelac22:
[RHCLOUD-25965]


src/components/NotificationsDrawer/DrawerPanelContent.tsx
```diff
@@ -65,7 +66,17 @@ const DrawerPanelBase = ({ innerRef }: DrawerPanelProps) => {
...
+    try {
+      const notifications = await axios.get('/api/notifications/v1/notifications/drawer');
+      dispatch(populateNotifications(notifications.data.data));
+    } catch (error) {
+      console.error('Unable to get Notifications ', error);
+    }
+  }
+
   useEffect(() => {
+    getNotifications();
```
Hyperkid123, line 79: The drawer should be populated on the Chrome start-up. Even if the drawer is closed. We need the data to change the toggle state if needed.

Getting the data after the drawer is opened is too late.

florkbr (approved):
Overall looks good to me for now. I have a few concerns around adding some tests and documentation for the newly added TS types (NotificationsPayload etc.), but that's also an issue for the entire repo (not just this PR). Confirmed with @Hyperkid123 offline that this PR works against the chrome-service WS.

florkbr:
Jenkins is confused by itself again 😮‍💨 (PR check container name conflict)

PR #2637, opened 2023-09-07
Notification Drawer Test Suite

arivepr:
Covering the [following ticket](https://issues.redhat.com/browse/RHCLOUD-27784)

### DIFF
--- a/src/components/NotificationsDrawer/DrawerPanelContent.test.tsx
+++ b/src/components/NotificationsDrawer/DrawerPanelContent.test.tsx
@@ -80,7 +80,7 @@ describe('Drawer panel functionality', () => {
     const actions = store.getActions();
 
     await waitFor(() => {
-      expect(actions).toContainEqual(markNotificationAsRead(1));
+      expect(actions).toContainEqual(markNotificationAsRead('1'));
     });
   });
 
@@ -98,7 +98,7 @@ describe('Drawer panel functionality', () => {
     const actions = store.getActions();
 
     await waitFor(() => {
-      expect(actions).toContainEqual(markNotificationAsUnread(1));
+      expect(actions).toContainEqual(markNotificationAsUnread('1'));
     });
   });
 
--- a/src/components/NotificationsDrawer/DrawerPanelContent.test.tsx
+++ b/src/components/NotificationsDrawer/DrawerPanelContent.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fireEvent, render } from '@testing-library/react';
+import { act, fireEvent, render, waitFor } from '@testing-library/react';
 import { Provider } from 'react-redux';
 import configureMockStore, { MockStore } from 'redux-mock-store';
 import DrawerPanel from './DrawerPanelContent';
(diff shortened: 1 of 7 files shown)


src/components/NotificationsDrawer/DrawerPanelContent.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    expect(renderedResult.getByText('Test Notification 1')).toBeInTheDocument();
+  });
+
+  test('Marking notification as read successfully', () => {
+    const store = mockStore(stateWithNotifications);
+
+    const renderedResult = renderComponent(store);
+
+    const checkbox = renderedResult.getAllByRole('checkbox');
+    fireEvent.click(checkbox[0]);
```
Hyperkid123, line 75: Events typically have to be wrapped in `act` or `waitFor` or their async variants. Otherwise it will output a lot of errors in console and state transitions might not happen.

```
  console.error
    Warning: An update to Popper inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
    
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at Popper (/home/martin/insights/insights-chrome/node_modules/@patternfly/react-core/src/helpers/Popper/Popper.tsx:152:3)
        at DropdownBase (/home/martin/insights/insights-chrome/node_modules/@patternfly/react-core/src/components/Dropdown/Dropdown.tsx:67:3)
        at Dropdown
        at div
        at div
        at NotificationDrawerHeader (/home/martin/insights/insights-chrome/node_modules/@patternfly/react-core/src/components/NotificationDrawer/NotificationDrawerHeader.tsx:30:3)
        at div
        at NotificationDrawerBase (/home/martin/insights/insights-chrome/node_modules/@patternfly/react-core/src/components/NotificationDrawer/NotificationDrawer.tsx:16:3)
        at NotificationDrawer
        at innerRef (/home/martin/insights/insights-chrome/src/components/NotificationsDrawer/DrawerPanelContent.tsx:58:28)
        at basenameProp (/home/martin/insights/insights-chrome/node_modules/react-router/lib/components.tsx:378:13)
        at basename (/home/martin/insights/insights-chrome/node_modules/react-router-dom/index.tsx:354:3)
        at Provider (/home/martin/insights/insights-chrome/node_modules/react-redux/lib/components/Provider.js:19:3)

```

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "noImplicitAny": true,
     "module": "esnext",
     "target": "esnext",
-    "jsx": "react",
+    "jsx": "react-jsx",
```
Hyperkid123, line 8: Why this change?

src/components/NotificationsDrawer/DrawerPanelContent.test.tsx
arivepr: Oh interesting, that's true for every "fireEvent" and things of that nature  i use?

I'll go over them

Hyperkid123: It's not about the `fireEvent` specifically. It's about whatever triggers a state transition in the component. Usually, these are events, but they also can be promises or even when component mounts. But in this case, it is likely its caused by the fireEvent calls.

arivepr:
I tried wrapping everything, but the warning continues to pop up. I'm assuming something withing the patternfly dropdown / popper integrations is causing the message, but I'm unsure how to clear it.

lv-settings.lua
Hyperkid123, line 1: This file should be pushed.

package-lock.json
Hyperkid123, line 1: Why are there changes to a lock file? This file should not change at all.

src/components/NotificationsDrawer/NotificationItem.tsx
```diff
@@ -38,7 +38,7 @@ const NotificationItem: React.FC<NotificationItemProps> = ({ notification, onNav
   return (
     <React.Fragment>
       <NotificationDrawerList>
-        <NotificationDrawerListItem variant="info" isRead={notification.read}>
+        <NotificationDrawerListItem aria-label={`Notification item ${notification.id}`} variant="info" isRead={notification.read}>
```
Hyperkid123, line 41: aria-label should be an accurate human readable description of an element. Using `id` is not great in this case. I would probably use the title if we want to use this attribute.

lv-settings.lua
arivepr: I have 0 clue how this got here

Hyperkid123:
> I tried wrapping everything, but the warning continues to pop up. After researching my guess is something withing the patternfly dropdown / popper integrations is causing the message, but I'm unsure how to clear it.

This might be because of async updates. Sometimes the act has to be used as an async wrapper:
```jsx
await act(async () => {
   // event that triggers state mutation
})
```

Even the initial render can cause these errors to pop up if there are any state mutations happening after mount.

Hyperkid123 (approved):

PR #2642, opened 2023-09-22
Fix notifications data undefined in ephem

aneelac22:


src/components/RootApp/ScalprumRoot.tsx
```diff
@@ -62,7 +62,7 @@ const ScalprumRoot = memo(
     async function getNotifications() {
       try {
         const notifications = await axios.get('/api/notifications/v1/notifications/drawer');
-        dispatch(populateNotifications(notifications.data.data));
+        dispatch(populateNotifications(notifications.data?.data || []));
```
Hyperkid123, line 65: I am not sure if this is the problem. It's already in a try-catch block. I think we have to verify the payload itself and its shape before we even attempt dispatching the action.

aneelac22: I think this should fix it, data basically returns undefined

fhlavac (approved):

PR #2658, opened 2023-10-09
RHCLOUD-28295: update chrome to use autosuggest search API

florkbr:
CC @Hyperkid123 

I'm still testing things out - namely trying to figure out if autosuggest supports typos. For example - `opensift` returns no results with the new API.


Hyperkid123:
> CC @Hyperkid123
> 
> I'm still testing things out - namely trying to figure out if autosuggest supports typos. For example - `opensift` returns no results with the new API.

We might need the second query after all. With the fuzzy range. Say if we get less than 10 results.

florkbr:
Update: the search team is taking a look at adding the ability to fuzzy search for us. I'm reworking this PR to fall back to our old logic - however we should give that new fuzzy match a shot too.

Hyperkid123:
Yup, let's just wait for what we get once they are finished.

Hyperkid123:
Added a commit to address the larger issues. We are forced to swap away to help the search team debug their perf issues.

PR #2663, opened 2023-10-10
Add hac SDKv4 compatibility layer.

Hyperkid123:
in part with: https://github.com/openshift/hac-core/pull/149

### DIFF
--- a/src/components/RootApp/ScalprumRoot.tsx
+++ b/src/components/RootApp/ScalprumRoot.tsx
@@ -27,7 +27,7 @@ import Footer, { FooterProps } from '../Footer/Footer';
 import updateSharedScope from '../../chrome/update-shared-scope';
 import useBundleVisitDetection from '../../hooks/useBundleVisitDetection';
 import chromeApiWrapper from './chromeApiWrapper';
-import { ITLess } from '../../utils/common';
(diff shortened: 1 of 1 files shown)


src/components/RootApp/ScalprumRoot.tsx
```diff
@@ -170,15 +170,16 @@ const ScalprumRoot = memo(
                   loadScripts: [],
                 };
               }
-              return {
+              const newManifest = {
                 ...manifest,
                 // Compatibility required for bot pure SDK plugins, HCC plugins and sdk v1/v2 plugins until all are on the same system.
-                baseURL: '/',
+                baseURL: manifest.name.includes('hac-') && !manifest.baseURL ? `${isBeta() ? '/beta' : ''}/api/plugins/${manifest.name}/` : '/',
```
florkbr, line 176: Obviously not an ideal solution here... Hopefully we can remove this in the near future when we migrate hac-dev

florkbr (approved):

src/components/RootApp/ScalprumRoot.tsx
Hyperkid123: They will have to update the build SDK to at least v3 for use to be able to remove it.

PR #2681, opened 2023-11-06
Combine suggester restuls.

Hyperkid123:
Combine results of the default and the improved suggester results.

### DIFF
--- a/src/components/Search/SearchInput.tsx
+++ b/src/components/Search/SearchInput.tsx
@@ -10,7 +10,7 @@ import uniq from 'lodash/uniq';
 import uniqWith from 'lodash/uniqWith';
 
 import './SearchInput.scss';
-import { AUTOSUGGEST_TERM_DELIMITER, SearchAutoSuggestionResponseType, SearchResponseType } from './SearchTypes';
+import { AUTOSUGGEST_TERM_DELIMITER, SearchAutoSuggestionResponseType, SearchAutoSuggestionResultItem, SearchResponseType } from './SearchTypes';
 import EmptySearchState from './EmptySearchState';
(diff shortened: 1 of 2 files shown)


src/components/Search/SearchInput.tsx
```diff
@@ -82,6 +82,22 @@ type SearchInputListener = {
   onStateChange: (isOpen: boolean) => void;
 };
 
+function parseSuggestions(suggestions: SearchAutoSuggestionResultItem[] = []) {
+  return suggestions.map((suggestion) => {
+    const [allTitle, bundleTitle, abstract] = suggestion.term.split(AUTOSUGGEST_TERM_DELIMITER);
+    const url = new URL(suggestion.payload);
```
karelhala, line 88: Wouldn't it be better to directly destruct the `pathname` here? `const { pathname} = ...`

src/components/Search/SearchInput.tsx
```diff
@@ -82,6 +82,22 @@ type SearchInputListener = {
...
+    const pathname = url.pathname;
+    const item = {
```
karelhala, line 90: Is there specific reason to assign this to a const instead of directly constructing it when returning value?

karelhala (approved):
Looking good, just 2 styling issues.

src/components/Search/SearchInput.tsx
Hyperkid123: It's a class, not an object. Destructing classes might be dangerous.

src/components/Search/SearchInput.tsx
Hyperkid123: None apart from better readability.

PR #2699, opened 2023-11-14
Use jotai for `activeModule` state managment 

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-29369

The Redux (flux) model is no longer the go-to state management model when it comes to React. Jotai is the spiritual successor to Recoil which is a state management model developed by Facebook engineers. Probably a good candidate to follow.

The good thing is, it allows us a gradual migration. We can pick state variables one by one and move them to the new model.

Important files: 
- https://github.com/RedHatInsights/insights-chrome/pull/2699/files#diff-4f391b6614cfc759f52964ba718774cf55654adf2acc6f23f70cf67eb0c0a08f
- https://github.com/RedHatInsights/insights-chrome/pull/2699/files#diff-35d73a5e8d99567a4b2cf5938c57256b822ff611d504775ca1079bd6d1f8689f
- https://github.com/RedHatInsights/insights-chrome/pull/2699/files#diff-3c8623839f8dcc909b7cac049e0daf4a20b5f4091a709c547758ddbd111c604b

EDIT

### Important to realize why single store state management can be problematic
[performance issues if not used correctly](https://stackblitz.com/edit/stackblitz-starters-4qzcyn?file=src%2Findex.tsx).

### DIFF
--- a/src/components/ChromeRoute/ChromeRoute.tsx
+++ b/src/components/ChromeRoute/ChromeRoute.tsx
@@ -2,7 +2,7 @@ import { ScalprumComponent } from '@scalprum/react-core';
 import React, { memo, useContext, useEffect } from 'react';
 import LoadingFallback from '../../utils/loading-fallback';
 import { batch, useDispatch, useSelector } from 'react-redux';
-import { changeActiveModule, toggleGlobalFilter, updateDocumentTitle } from '../../redux/actions';
+import { toggleGlobalFilter, updateDocumentTitle } from '../../redux/actions';
 import ErrorComponent from '../ErrorComponents/DefaultErrorComponent';
 import { getPendoConf } from '../../analytics';
 import classNames from 'classnames';
@@ -12,6 +12,8 @@ import { ReduxState } from '../../redux/store';
 import { DeepRequired } from 'utility-types';
 import { ChromeUser } from '@redhat-cloud-services/types';
 import ChromeAuthContext from '../../auth/ChromeAuthContext';
+import { useAtom } from 'jotai';
+import { activeModuleAtom } from '../../state/atoms';
 
 export type ChromeRouteProps = {
   scope: string;
@@ -29,8 +31,9 @@ const ChromeRoute = memo(
     const { setActiveHelpTopicByName } = useContext(HelpTopicContext);
     const { user } = useContext(ChromeAuthContext);
     const gatewayError = useSelector(({ chrome: { gatewayError } }: ReduxState) => gatewayError);
-    const activeModule = useSelector(({ chrome: { activeModule } }: ReduxState) => activeModule);
     const defaultTitle = useSelector(({ chrome: { modules } }: ReduxState) => modules?.[scope]?.defaultDocumentTitle || scope);
+
+    const [activeModule, setActiveModule] = useAtom(activeModuleAtom);
 
     useEffect(() => {
       batch(() => {
@@ -42,7 +45,7 @@ const ChromeRoute = memo(
            */
           dispatch(updateDocumentTitle(defaultTitle || 'Hybrid Cloud Console'));
         }
-        dispatch(changeActiveModule(scope));
+        setActiveModule(scope);
       });
       /**
        * update pendo metadata on application change
--- a/src/components/Header/HeaderTests/__snapshots__/ToolbarToggle.test.js.snap
+++ b/src/components/Header/HeaderTests/__snapshots__/ToolbarToggle.test.js.snap
@@ -9,7 +9,7 @@ exports[`ToolbarToggle should render correctly 1`] = `
   data-popper-escaped="true"
   data-popper-placement="bottom-end"
   data-popper-reference-hidden="true"
-  style="position: absolute; left: 0px; top: 0px; z-index: 9999; min-width: 0px; transform: translate(0px, 0px);"
+  style="position: absolute; left: 0px; top: 0px; z-index: 9999; opacity: 1; transition: opacity 0ms cubic-bezier(.54, 1.5, .38, 1.11); min-width: 0px; transform: translate(0px, 0px);"
 >
   <div
     class="pf-v5-c-menu__content"
@@ -24,14 +24,14 @@ exports[`ToolbarToggle should render correctly 1`] = `
         data-ouia-component-type="PF5/DropdownItem"
         data-ouia-safe="true"
         href="url1"
-        rel="noopener noreferrer"
         role="none"
-        target="_blank"
       >
         <a
           class="pf-v5-c-menu__item"
+          rel="noopener noreferrer"
           role="menuitem"
           tabindex="0"
+          target="_blank"
         >
           <span
             class="pf-v5-c-menu__item-main"
--- a/src/redux/chromeReducers.ts
+++ b/src/redux/chromeReducers.ts
@@ -154,14 +154,6 @@ export function loadModulesSchemaReducer(
   };
 }
 
-export function changeActiveModuleReducer(state: ChromeState, { payload }: { payload: string }): ChromeState {
-  return {
-    ...state,
-    activeModule: payload,
-    appId: payload,
-  };
-}
-
 export function setPendoFeedbackFlag(
   state: ChromeState,
   {
--- a/src/state/atoms.ts
+++ b/src/state/atoms.ts
@@ -1,0 +1,4 @@
+import { atom } from 'jotai';
+// setup initial chrome atoms
+export const contextSwitcherOpenAtom = atom(false);
+export const activeModuleAtom = atom<string | undefined>(undefined);
--- a/src/state/chromeStore.ts
+++ b/src/state/chromeStore.ts
@@ -1,0 +1,15 @@
+import { createStore } from 'jotai';
+import { activeModuleAtom, contextSwitcherOpenAtom } from './atoms';
+
+const chromeStore = createStore();
+
+// setup initial chrome store state
+chromeStore.set(contextSwitcherOpenAtom, false);
+chromeStore.set(activeModuleAtom, undefined);
+
+// globally handle subscription to activeModuleAtom
+chromeStore.sub(activeModuleAtom, () => {
+  // console.log('activeModule in store', chromeStore.get(activeModuleAtom));
+});
+
+export default chromeStore;
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,7 @@
         "classnames": "^2.3.2",
         "commander": "^10.0.0",
         "history": "^5.3.0",
+        "jotai": "^2.5.1",
         "js-cookie": "^3.0.1",
         "js-yaml": "^4.1.0",
         "localforage": "^1.10.0",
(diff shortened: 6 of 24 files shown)


Hyperkid123:
IQE is timing out on the pr checks

src/state/atoms.ts
```diff
@@ -0,0 +1,4 @@
+import { atom } from 'jotai';
+// setup initial chrome atoms
+export const contextSwitcherOpenAtom = atom(false);
+export const activeModuleAtom = atom<string | undefined>(undefined);
```
florkbr, line 4: @Hyperkid123 curious on your thoughts on utilizing a pattern where we export a named type in support of atoms versus inline primitive types. IE:
```ts
// src/state/atom-types.ts or similar
export type ActiveModule = string | undefined;

// src/state/atoms.ts
export const activeModuleAtom = atom<ActiveModule>(undefined);
```

I get for this particular atom it's not likely it'll change drastically - but for those that might it would be a good habit to get in?

florkbr: And it may be beneficial to reserve `atoms.ts` as the main aggregate file that simply exports from more specific files in the directory - for example `context-switcher-atoms.ts` (to hopefully give us our best chance of avoiding merge conflicts in the main `atoms.ts`).

src/redux/chromeReducers.ts
```diff
@@ -155,14 +155,6 @@ export function loadModulesSchemaReducer(
   };
 }
 
-export function changeActiveModuleReducer(state: ChromeState, { payload }: { payload: string }): ChromeState {
```
florkbr, line 158: It is pretty crazy how much Redux logic can be removed switching to a bottom up state management solution.

src/state/chromeStore.ts
```diff
@@ -0,0 +1,15 @@
+import { createStore } from 'jotai';
+import { activeModuleAtom, contextSwitcherOpenAtom } from './atoms';
+
+const chromeStore = createStore();
+
+// setup initial chrome store state
+chromeStore.set(contextSwitcherOpenAtom, false);
```
florkbr, line 7: @Hyperkid123 a few questions here:
1. In `src/state/atoms.ts` we already specify what appears to be a default value of `false`. What is the point of setting it again here? If this is required by jotai, can we somehow reduce the duplication?
2. Similar to my comments in `src/state/atoms.ts` we may want to consider how to organize these files to avoid merge conflicts in the future.

src/state/chromeStore.ts
```diff
@@ -0,0 +1,15 @@
...
+chromeStore.set(activeModuleAtom, undefined);
+
+// globally handle subscription to activeModuleAtom
+chromeStore.sub(activeModuleAtom, () => {
+  // console.log('activeModule in store', chromeStore.get(activeModuleAtom));
```
florkbr, line 12: I'm assuming we left this in as a POC - but may be cool to be able to log this if we set a debugging flag. Also - this function returns an `unsub` callback - should we have a global handler to invoke all unsub's when the page is being torn down?

src/components/Header/HeaderTests/__snapshots__/ToolbarToggle.test.js.snap
```diff
@@ -9,7 +9,7 @@ exports[`ToolbarToggle should render correctly 1`] = `
   data-popper-escaped="true"
   data-popper-placement="bottom-end"
   data-popper-reference-hidden="true"
-  style="position: absolute; left: 0px; top: 0px; z-index: 9999; min-width: 0px; transform: translate(0px, 0px);"
+  style="position: absolute; left: 0px; top: 0px; z-index: 9999; opacity: 1; transition: opacity 0ms cubic-bezier(.54, 1.5, .38, 1.11); min-width: 0px; transform: translate(0px, 0px);"
```
florkbr, line 12: FWIW these changes are unrelated to the Jotai change - the version of PF updated with the update to package-lock.json

src/components/ChromeRoute/ChromeRoute.tsx
```diff
@@ -41,7 +44,7 @@ const ChromeRoute = memo(
            */
           dispatch(updateDocumentTitle(defaultTitle || 'Hybrid Cloud Console'));
         }
-        dispatch(changeActiveModule(scope));
+        setActiveModule(scope);
```
florkbr, line 47: @Hyperkid123 I think Jotai works great for this use case - simple global state management. However in my research I'm finding that folks with more complex use cases still needed something such as Zustand for more operations/actions/state management. See https://docs.pmnd.rs/zustand/getting-started/comparison#jotai for a good comparison. A good example would be an incrementer - where we want to control how the state is updated. I think it's possible in Jotai - it would just require more custom shared logic.

That said - are we suggesting Jotai replace _all_ of redux? Or just part of it?

src/state/atoms.ts
Hyperkid123: I think splitting atoms into individual files is something we want to do. Don't know about detaching the types to be in separate files. It seems like too much of a separation especially when this chrome is not a library to be consumed by anyone else.

src/redux/chromeReducers.ts
Hyperkid123: Yup. And you can use the atoms outside of the react context out of the box.

src/state/chromeStore.ts
Hyperkid123: > In src/state/atoms.ts we already specify what appears to be a default value of false. What is the point of setting it again here? If this is required by jotai, can we somehow reduce the duplication?

This is actually required. It has to be set.

But I think we might not even need the store TBH, we can work directly with the atoms. I have to rethink this a little bit.

src/state/chromeStore.ts
Hyperkid123: Yes POC, again, we might not even need the store.

src/components/ChromeRoute/ChromeRoute.tsx
Hyperkid123: We want to replace everything redux related.

Looking at the comparison I see outdated bits, like with the external store (which Jotai has and is in this PR).

The overall goal is not to obfuscate the actual state mutation and what is happening for example in a reducer. Or a store. Or a selector. The data should be prepared in the component (or some generic functions) and the atom should be updated with the final value.

The fact that jotai is so "dumb" forces us to be smart.

> I think it's possible in Jotai - it would just require more custom shared logic that would require enforcement on our end.

Exactly the goal. We write our own standarts and enforcements instead of relying on something provided. One nice example of this is the selector magic. It claims to optimize rendering but it never mentions that the second one uses `{...}` and does not include often very expensive compare functions the entire optimization goes out of window. And your component will be re-rendered **on each store update if do not have shallow objects with only primitive types**.

For example
```jsx
const state = {
  chrome: {
    nav: {
      isOpen: false
    },
    schemas: {
      foo: {},
      bar: {}
    },
    activeModule: 'foo'
  }
}

const Component = () => {
  const navSchema = useSelector((state) => {
  // creating new object, and creating shallow copy of a state object totally kills the shallow equal what selector based managers are doing
  // plus it hurts to read this
  // selector will trigger render on any state mutation regardless if the mutated state is actually used
    return {
      ...state.chrome.schemas[state.chrome.activeModule],
      isOpen: state.chrome.nav.isOpen
    }
  }, /** We would to implement some strange comparator to  optimize renders*/)
  return (
   ...
  )
}
```

Using atom based approach does not give us the option to create new state objects. It forces us to write the code how we should with no shortcuts. With the selectors we are just kicking the can down the road hoping that at some point someone will optimize the selectors.

Yes, it will be more work, but I believe ultimately it will pay dividents in both performance and code readability.

Hyperkid123:
I think this speaks for thousand words why single store state managers can introduce [performance issues if not used correctly](https://codesandbox.io/p/sandbox/pensive-stallman-9k6695?file=%2Fsrc%2FApp.tsx%3A4%2C25&layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clpawmgaw00072v6jatpkt62j%2522%252C%2522sizes%2522%253A%255B100%252C0%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clpawmgaw00032v6jjlmiciof%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clpawmgaw00042v6jorhpwpwi%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clpawmgaw00062v6jzsmhbgqq%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B59.84443550395617%252C40.15556449604383%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clpawmgaw00032v6jjlmiciof%2522%253A%257B%2522id%2522%253A%2522clpawmgaw00032v6jjlmiciof%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clpaz1hfq00032v6jerpzpc75%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A4%252C%2522startColumn%2522%253A25%252C%2522endLineNumber%2522%253A4%252C%2522endColumn%2522%253A25%257D%255D%252C%2522filepath%2522%253A%2522%252Fsrc%252FApp.tsx%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%255D%252C%2522activeTabId%2522%253A%2522clpaz1hfq00032v6jerpzpc75%2522%257D%252C%2522clpawmgaw00062v6jzsmhbgqq%2522%253A%257B%2522id%2522%253A%2522clpawmgaw00062v6jzsmhbgqq%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clpawmgaw00052v6jyctiv48m%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522UNASSIGNED_PORT%2522%252C%2522port%2522%253A0%252C%2522path%2522%253A%2522%2522%257D%255D%252C%2522activeTabId%2522%253A%2522clpawmgaw00052v6jyctiv48m%2522%257D%252C%2522clpawmgaw00042v6jorhpwpwi%2522%253A%257B%2522tabs%2522%253A%255B%255D%252C%2522id%2522%253A%2522clpawmgaw00042v6jorhpwpwi%2522%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Afalse%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D).

florkbr:
@Hyperkid123 I agree - the example stackblitz link you shared shows why ***_out of the box_*** some of these store state managers can introduce re-renders and performance issues.

FWIW This is not an argument for nor against redux, however there are well known patterns for avoiding these situations. Below is updated redux code (taken from stackblitz) using `reselect` to create memoized selectors and avoid unnecessary re-renders while focusing on splitting components into container components (that interact with the store and fetch data) and presentational components (which are responsible for on the glass). Splitting into these 2 types of components makes it easier to enforce selector logic etc.

```tsx
import { createStore as createReduxStore, Reducer } from 'redux';
import {
  Provider as ReduxProvider,
  useDispatch,
  useSelector as reduxSelector,
} from 'react-redux';
import { createSelector } from 'reselect';
import { useRef, memo } from 'react';
import { initialUser, User } from './user';

type ReduxState = {
  user: User;
  counter: number;
};

const reduxReducer: Reducer<ReduxState> = (
  state = {
    user: initialUser,
    counter: 0,
  },
  action
) => {
  return {
    ...state,
    counter: state.counter + 1,
  };
};

const reduxStore = createReduxStore(reduxReducer, {
  user: initialUser,
  counter: 0,
});

// Presentational component
const ReduxUserDisplay = (props) => {
  const renders = useRef(0);
  renders.current += 1;
  return (
    <div
      style={{
        border: '1px solid black',
        padding: 4,
        margin: 4,
      }}
    >
      <h3>
        "Isolated" user component that only shows the user value from the redux
        store and its render count
      </h3>
      <h4>Render count: {renders.current}</h4>
      <p>
        Why is the render count so large? Why does this component is rendered
        even though the user data is always the same?
      </p>
      <p>
        The selector always creates a new object reference of a nested
        structure. The base shallow equal function will not help out in this
        case. And the user component will be{' '}
        <b>re-render on any state mutation!</b>
      </p>
      <pre>{JSON.stringify(props?.user)}</pre>
    </div>
  );
};

const ReduxUserDisplayMemo = memo(ReduxUserDisplay);

// container component
const ReduxUserDisplayWrapper = () => {
  const user = createSelector(
    [(state) => state.user],
    user => {
      console.log('memoized selector ran');
      return user;
    }
  );
  return <ReduxUserDisplayMemo user={user} />;
};

// presentational component
const ReduxCounter = (props) => {

  return (
    <div>
      <h3>Counter value: {props?.count}</h3>
      <div>
        <button onClick={() => dispatch({ type: 'foo' })}>Increment</button>
      </div>
    </div>
  );
};

const ReduxCounterMemo = memo(ReduxCounter);

// container component
const ReduxCounterWrapper = () => {
  const count = reduxSelector((state: ReduxState) => state.counter);
  return <ReduxCounterMemo count={count} />;
};

const ReduxComponent = () => {
  return (
    <ReduxProvider store={reduxStore}>
      <div>
        <h2>Redux component</h2>
        <ReduxCounterWrapper />
        <ReduxUserDisplayWrapper />
        <hr />
      </div>
    </ReduxProvider>
  );
};

export default ReduxComponent;

```

Which results in only 1 render as expected:
image

florkbr:
That said - I see your point regarding atomic store. It's impossible to get in this scenario whereas it's very easy to in Redux/Zustand without prior experience.

Hyperkid123:
Yeah, you can avoid the issues with the selectors. But:

1. You will forget
2. You need extra resources to run the compare functions
3. Look at how much extra code you need to make it work, (memo, extra dependency, all the extra TS, etc.)

And in reality, the only major change when it comes to writing the code is basically just moving logic from selectors and actions to components/functions.

PR #2704, opened 2023-11-16
Add aditional logic for pendo guides stopping.

Hyperkid123:
Why?

Even though the `pendo` object might exist on window, it does not mean all the functions do :upside_down_face: . We have to try a few times until the functions are present on the global object.

I've also added a timeout function to push the guides toggle execution on the end of the stack


florkbr (approved):

florkbr:
It would be nice to understand why pendo functions are missing after the JS is loaded... very odd. For now this seems like our best course of action.

PR #2706, opened 2023-11-16
Move console.err after pendo retries are exhausted

fhlavac:
Moved console.err after pendo retries are exhausted to fix errors occurring on stage (Bryan's comment below)

cc @Hyperkid123

### DIFF
--- a/src/hooks/useDisablePendoOnLanding.ts
+++ b/src/hooks/useDisablePendoOnLanding.ts
@@ -12,12 +12,13 @@ function retry(fn: () => void, retriesLeft = 10, interval = 100) {
   try {
     return fn();
   } catch (error) {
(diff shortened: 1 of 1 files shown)


florkbr:
We introduced a fix here: https://github.com/RedHatInsights/insights-chrome/pull/2704 b/c pendo functions are not available even after the pendo global is ready.

We started noticing many errors on stage:
image

There is another commit that removed the safety checks on the functions (as the retry logic works on catching errors): https://github.com/RedHatInsights/insights-chrome/commit/8b07c1f331f4e5ed5f235bb6660951883b0118e8

However this caused our console.error to go crazy. This PR moves the error to the end of the retry attempts.

florkbr (approved):

PR #2714, opened 2023-11-30
[HOLD] Revert "Use oidc instead of keycloack-js."

florkbr:
Reverts RedHatInsights/insights-chrome#2702

See slack thread: https://redhat-internal.slack.com/archives/C023VGW21NU/p1701356452137739 

[subscriptions-inventory-ui](https://github.com/RedHatInsights/subscription-inventory-ui/blob/main/src/utilities/platformServices.ts#L70) is reporting auth failures (empty token from `useChrome().auth.getToken()`). Trying to investigate what is broken before reverting so we know what to fix on our end.


florkbr:
Hold while investigating please.

Hyperkid123:
I think we wont need the revert anymore

PR #2743, opened 2024-01-11
Use mutable reference for fetch/xhr Auth header monkey patch

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-30334


fhlavac (approved):
Tested locally and everything seems to be working 🙂

PR #2762, opened 2024-02-05
[RHCLOUD-30109] Update chrome to use new lastVisited hook

BlakeHolifield:
* use new lastVisited hook to send `last-visited` pages on an interval and when the page is hidden
* store state in local storge, but fill the local storage from the database when empty
* add cypress tests to cover POST and GET to the api


Hyperkid123 (approved):

Hyperkid123:
@BlakeHolifield before we merge this even to stage can you make sure the compatible chrome service API is promoted to production?

Just want to be super safe.

Hyperkid123 (approved):

PR #2779, opened 2024-02-28
Remove bounce in page title

CodyWMitchell:
[RHCLOUD-27874](https://issues.redhat.com/browse/RHCLOUD-27874)

- Prevent the page title changing unexpectedly

### DIFF
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -37,8 +37,8 @@ const App = () => {
   useInitializeAnalytics();
 
   useEffect(() => {
-    const title = typeof documentTitle === 'string' ? `${documentTitle} | ` : '';
-    document.title = `${title}console.redhat.com`;
+    const title = typeof documentTitle === 'string' ? `${documentTitle} | Hybrid Cloud Console` : 'Hybrid Cloud Console';
+    document.title = title;
   }, [documentTitle]);
 
   return <RootApp cookieElement={cookieElement} setCookieElement={setCookieElement} />;
--- a/src/bootstrap.tsx
+++ b/src/bootstrap.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { createRoot } from 'react-dom/client';
 import { Provider, useSelector } from 'react-redux';
 import { IntlProvider, ReactIntlErrorCode } from 'react-intl';
@@ -36,7 +36,7 @@ const App = () => {
 
   useInitializeAnalytics();
 
-  useEffect(() => {
+  useMemo(() => {
     const title = typeof documentTitle === 'string' ? `${documentTitle} | Hybrid Cloud Console` : 'Hybrid Cloud Console';
     document.title = title;
   }, [documentTitle]);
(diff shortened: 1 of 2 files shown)


CodyWMitchell:
@karelhala The bouncing page title mentioned in [RHCLOUD-27874](https://issues.redhat.com/browse/RHCLOUD-27874) is fixed by these changes, and removing those two lines leads to the expected behavior mentioned in the ticket. However, if someone is depending on that update to automatically set the title for them, they will end up defaulting to "Hybrid Cloud Console" as the page title. (This could potentially affect automated tests.)

src/bootstrap.tsx
```diff
@@ -37,8 +37,8 @@ const App = () => {
   useInitializeAnalytics();
 
   useEffect(() => {
-    const title = typeof documentTitle === 'string' ? `${documentTitle} | ` : '';
-    document.title = `${title}console.redhat.com`;
+    const title = typeof documentTitle === 'string' ? `${documentTitle} | Hybrid Cloud Console` : 'Hybrid Cloud Console';
```
Hyperkid123, line 40: We have to ensure we are not trying to assign a different type to the title. We can use `useMemo` instead of `useEffect` to properly set the titles.

The difference between these two is that `useMemo` runs even on a first render in addition to the dependency changes. `useEffect` runs only after the first render (and dependency changes). 

This changes might remvoe some of the issues as well in addtion to the default changes.

src/components/ChromeRoute/ChromeRoute.tsx
```diff
@@ -32,21 +31,13 @@ const ChromeRoute = memo(
     const { setActiveHelpTopicByName } = useContext(HelpTopicContext);
     const { user } = useContext(ChromeAuthContext);
     const gatewayError = useSelector(({ chrome: { gatewayError } }: ReduxState) => gatewayError);
-    const chromeModules = useAtomValue(chromeModulesAtom);
-    const defaultTitle = chromeModules?.[scope]?.defaultDocumentTitle || scope;
 
-    const [activeModule, setActiveModule] = useAtom(activeModuleAtom);
+    const [, setActiveModule] = useAtom(activeModuleAtom);
```
Hyperkid123, line 35: You can use the `useSetAtom` hook if you only want the setter function.

src/components/ChromeRoute/ChromeRoute.tsx
```diff
@@ -32,21 +31,13 @@ const ChromeRoute = memo(
...
 
     useEffect(() => {
       batch(() => {
         // Only trigger update on a first application render before any active module has been selected
         // should be triggered only once per session
-        if (!activeModule) {
-          /**
-           * Default document title update. If application won't update its title chrome sets a title using module config
-           */
-          dispatch(updateDocumentTitle(defaultTitle || 'Hybrid Cloud Console'));
```
Hyperkid123, line 48: Because I don't remember, please make sure the `updateDocumentTitle` action is used in other places. If its not, remove it completely from the codebase.

Hyperkid123:
> @karelhala The bouncing page title mentioned in [RHCLOUD-27874](https://issues.redhat.com/browse/RHCLOUD-27874) is fixed by these changes, and removing those two lines leads to the expected behavior mentioned in the ticket. However, if someone is depending on that update to automatically set the title for them, they will end up defaulting to "Hybrid Cloud Console" as the page title. (This could potentially affect automated tests.)

@CodyWMitchell as long as the title from the state has higher priority, any changes based on UX feedback we make are OK. Services should set their titles. They should not rely on the fallback value we provide.

src/components/ChromeRoute/ChromeRoute.tsx
CodyWMitchell: `updateDocumentTitle` is being used in a few other places (`AllServices.tsx`, `FavoritedServices.tsx`, `create-chrome.ts`), so I left it in place.

CodyWMitchell:
@Hyperkid123 The mentioned issues should be resolved now. :+1:

Hyperkid123 (approved):

PR #2784, opened 2024-03-05
[RHCLOUD-31132] Make sure dev env's use staticConfig instead of operator-generated

aneelac22:


Hyperkid123:
Any particular reason for this? Ideally, we distinguish between envs via configuration, not in code.

aneelac22:
> Any particular reason for this? Ideally, we distinguish between envs via configuration, not in code.

I was told there is no dev deployment and it uses stage deployment for frontend and prod for backend and no way to change config in Operator because of that

Hyperkid123:
Is there a conversation or a ticket or something I can read?

aneelac22:
> Is there a conversation or a ticket or something I can read?

https://redhat-internal.slack.com/archives/C05KTHJU4R4/p1708617749235119?thread_ts=1708540844.466109&cid=C05KTHJU4R4

fhlavac (approved):

PR #2796, opened 2024-03-20
Fix hideGlobalFilter() hook

lucasgarfield:
We would like to remove the global filter from our Insights application because we do not use it and it takes up a large amount of vertical space.

I’ve noticed that there are two hooks related to the global filter, `hideGlobalFilter` and `removeGlobalFilter`. 

`removeGlobalFilter` is said to be deprecated and users are advised to use `hideGlobalFilter` instead. 

However, the behavior of these two hooks is different.

`removeGlobalFilter` is used in the global filter component’s wrapper function to determine whether or not to render the global filter. `hideGlobalFilter` is used in the global filter component itself to determine whether or not to disable the global filter.

Furthermore, the conditional logic in the global filter wrapper related to `removeGlobalFilter` was, I believe, wrong… I’ve fixed it in this commit and the `removeGlobalFilter` hook is working now.

You can see it working here, I have called `removeGlobalFilter` in our application using a local Chrome build with these changes:

removeglobalfilter

One thing I don’t love about this solution is that the filter will still appear while the app is loading, before the removeGlobalFilter hook is called, which you can see in the gif I've uploaded.

If you do not want to update the `removeGlobalFilter` hook, an alternative solution might be to adjust the `isGlobalFilterAllowed()` function https://github.com/RedHatInsights/insights-chrome/blob/5c1b0adc39e0d371d7c9dc024d07f3dc6f09197b/src/utils/common.ts#L420 to return `false` for the our app. This would also have the benefit of hiding the global filter even while our app is loading. I’d be happy to put a pull request together if this sounds reasonable.


Hyperkid123:
@karelhala @rlong keep me honest here. The global filter should be visible in all apps in /the insights bundle regardless if it's used or not within the app. This is based on discussions with UX that happened a while back agon (a few years now).

Maybe we should revisit that discussion again.

ryelo:
Correct. The last guidance we have is that the filter should be available in all applications inside of /insights, whether or not teams are actually using it.

lucasgarfield:
@Hyperkid123 and @ryelo, thanks for pointing that out, I'll get in touch with everyone.

Assuming I can get consensus from UX and HCC, would you prefer we fix the `removeGlobalFilter` hook or hardcode an exception for our app into the `isGlobalFilterAllowed()` function? To be honest, I would prefer the second option so that the filter does **not** appear while our app is loading.

Hyperkid123:
@lucasgarfield the behavior should be consistent across apps/bundles. If UX/PM says its no longer needed to show the everywhere, we will come up with a solution so your does not have to add any workarounds.

lucasgarfield:
We've gotten some traction from UX/PM. I'll follow up once we get the final necessary acks via e-mail. I'll go ahead and close this so it isn't polluting your PR tab.

PR #2807, opened 2024-03-29
sentry hook

adonispuente:
Hook to have individual apps send up their appname to sentry when their initialized (eventually this will dictate the project errors are sent too), but since chrome is initialized first, the sentry innit function always triggered before the app sends up the string. Still working on how to do so


src/redux/actions.ts
```diff
@@ -202,3 +202,9 @@ export const updateNotifications = (payload: NotificationsPayload) => ({
   type: actionTypes.UPDATE_NOTIFICATIONS,
   payload,
 });
+export function updateSentryProject(sentryProject: string) {
```
Hyperkid123, line 205: Just FYI, we are moving away from redux to Jotai. If you want new state variables, use jotai please.

PR #2813, opened 2024-04-10
[RHCLOUD-31672] Add regional ocm api instances

aneelac22:


src/utils/iqeEnablement.ts
```diff
@@ -16,6 +16,7 @@ const AUTH_ALLOWED_ORIGINS = [
   location.origin,
   /https:\/\/api(?:\.[a-z]+)?\.openshift(?:[a-z]+)?\.com/,
   /https:\/\/api?\.demo-experience(?:\.[a-z]+)?\.demo?\.redhat?\.com/,
+  /https:\/\/api?\.aws?\.ap-southeast-1(?:\.[a-z]+)?\.openshift?\.com/,
```
Hyperkid123, line 19: Just a note. We should start thinking about OCP env variables and pass this list down from chrome.

Hyperkid123 (approved):

src/utils/iqeEnablement.ts
KevinFCormier: @aneelac22 There will be many new regional endpoints following the pattern api.<PROVIDER>.<REGION>.<ENV>.openshift.com. Can we make this more generic so that we don't have to request a change for each new region?

PR #2826, opened 2024-04-25
Hide global filter when disabled.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-31582

Uses the `isDisabled` prop to remove the global filter rather than just disabling the dropdown.


Hyperkid123:
Oh yeah.

fhlavac (approved):

PR #2828, opened 2024-04-29
update insights bundle title suffix to RHEL

mkholjuraev:
interact team is working on consolidating the insights page titles [RHINENG-9387](https://issues.redhat.com/browse/RHINENG-9387). It requires those page titles to have 'RHEL' instead of  the 'Red hat insights' suffix at the end. This PR makes this change in chrome repo so that the default suffix is applied to all insights apps in the bundle.

### DIFF
--- a/src/hooks/useBundle.ts
+++ b/src/hooks/useBundle.ts
@@ -19,7 +19,7 @@ export const bundleMapping: {
   'application-services': 'Application and Data Services',
   openshift: 'OpenShift',
   ansible: 'Ansible Automation Platform',
-  insights: 'Red Hat Insights',
+  insights: 'RHEL',
   edge: 'Edge management',
   settings: 'Settings',
   landing: 'Home',


Hyperkid123 (approved):

mkholjuraev:
@Hyperkid123 I am wondering if this commit has already been released. I am trying this out locally with another tenant app, but I still see the old "Red Hat insights" suffix.

Hyperkid123:
Not in prod. Also apps can can set their titles independently if that is where you see it.

Hyperkid123:
@mkholjuraev On this page: https://console.stage.redhat.com/preview/insights/dashboard#SIDs=&tags=

I get this output when I check the method you updated: 

```js
window.insights.chrome.getBundleData()
{bundleId: 'insights', bundleTitle: 'RHEL'}
```

And the tab title is indeed `Dashboard | RHEL`.

Seems ok to me. Is it the Chrome app adding it? Can't it be the app changing the title internally?

PR #2829, opened 2024-05-02
RHCLOUD-32399 update link

aferd:
Jira - https://issues.redhat.com/browse/RHCLOUD-32399
Update link so that on hover affect is only on the link text and not the bundle subtext.


Hyperkid123 (commented):
We have some invalid DOM nesting. Can you remove the double `a` tag as well? And have just one.

image

Hyperkid123 (approved):

PR #2832, opened 2024-05-10
RHCLOUD-32008: create internal jotai state for preview

florkbr:
Adds an internal state variable via jotai to track if preview is enabled or disabled.

Tested locally by adding a sub in the store with a console log (see screenshots). The state is correctly toggled with the preview toggle and initialized when chrome loads.

### Preview disabled
Screenshot 2024-05-10 at 12 50 32 AM
### Preview enabled
Screenshot 2024-05-10 at 12 50 44 AM

### DIFF
--- a/src/state/atoms/utils.ts
+++ b/src/state/atoms/utils.ts
@@ -1,0 +1,11 @@
+import { WritableAtom, atom } from 'jotai';
+
+// recipe from https://jotai.org/docs/recipes/atom-with-toggle
+export function atomWithToggle(initialValue?: boolean): WritableAtom<boolean, [boolean?], void> {
+  const anAtom = atom(initialValue, (get, set, nextValue?: boolean) => {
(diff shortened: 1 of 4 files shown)


src/state/atoms/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { WritableAtom, atom } from 'jotai';
+
+// recipe from https://jotai.org/docs/recipes/atom-with-toggle
```
florkbr, line 3: @Hyperkid123 thoughts on including this? This allows us to eventually toggle the atom value without the need for checking `isBeta()` in the future.

Hyperkid123: Eventually yes, but now with the URL and local storage being in play its a bit too early for it.

Hyperkid123 (approved):

PR #2835, opened 2024-05-14
End preview

Hyperkid123:
tickets:
- https://issues.redhat.com/browse/RHCLOUD-32012
- https://issues.redhat.com/browse/RHCLOUD-32013
- https://issues.redhat.com/browse/RHCLOUD-32011

### Changes
- load the isPreview initial value from Chrome service
- Sync Chrome service data with UI
- update Unleash context preview variable with on preview toggle
- Start using the internal variable instead of `isBeta` **if possible**
- Hide the new preview functions behind the `chrome:local-preview` local storage variable

@catastrophe-brandon This PR should introduce the new preview feature if the local storage is present. The current preview functionality should be unaffected.

To anyone who will be looking at the PR. This must be properly tested/verified. We don't want to block the Chrome UI deployment. The local storage gating must be 100% reliable.


src/utils/useNavigation.ts
```diff
@@ -108,7 +111,11 @@ const useNavigation = () => {
       axios
         .get(`${getChromeStaticPathname('navigation')}/${currentNamespace}-navigation.json`)
         // fallback static CSC for EE env
-        .catch(() => axios.get<BundleNavigation>(`${isBeta() ? '/beta' : ''}/config/chrome/${currentNamespace}-navigation.json?ts=${Date.now()}`))
+        .catch(() => {
+          const LOCAL_PREVIEW = localStorage.getItem('chrome:local-preview') === 'true';
```
florkbr, line 115: @Hyperkid123 more of a nit - but I would suggest we move this to a helper function so we avoid any mis-spellings or incorrect local storage checks. I'd be fine exporting this temporarily from `state/atoms/releaseAtom` as a sort of temporary helper.

src/analytics/index.ts
```diff
@@ -14,23 +14,22 @@ function isInternalFlag(email: string, isInternal = false) {
   return '';
 }
 
-function getUrl(type?: string) {
+function getUrl(type?: string, isPreview = false) {
```
florkbr, line 17: @Hyperkid123 should we place the `isPreview` parameter before `type` as it is not technically optional (even though we have the default value)? 

This would prevent us from having to specify `getUrl(undefined, isPreview)`.

Another alternative would be to convert `getUrl` to accept an object so the key ordering doesn't matter.

src/bootstrap.tsx
```diff
@@ -48,6 +60,23 @@ const App = () => {
   return <RootApp cookieElement={cookieElement} setCookieElement={setCookieElement} />;
 };
 
+const ConfigLoader = () => {
+  const initPreview = useSetAtom(isPreviewAtom);
+  function initSuccess(userConfig: ChromeUserConfig) {
+    initPreview(userConfig.data.uiPreview);
+  }
+  function initFail() {
+    initPreview(false);
```
florkbr, line 69: I wonder if it's worth displaying a notification in this scenario so it's clear we've reverted back to non-preview due to a config init fault.

src/chrome/create-chrome.ts
```diff
@@ -153,7 +155,7 @@ export const createChromeContext = ({
       }
       dispatch(toggleGlobalFilter(isHidden));
     },
-    isBeta,
+    isBeta: () => isPreview,
```
florkbr, line 158: Not sure if possible - but I wonder if we mark `isBeta` deprecated in the type and expose `isPreview` directly? I understand we need to keep this for some time though not to break consumers.

src/auth/OIDCConnector/utils.ts
```diff
@@ -44,6 +44,7 @@ export async function logout(auth: AuthContextProps, bounce?: boolean) {
       key.startsWith(GLOBAL_FILTER_KEY)
   );
   deleteLocalStorageItems([...keys, OFFLINE_REDIRECT_STORAGE_KEY, LOGIN_SCOPES_STORAGE_KEY]);
+  // FIXME: Remove this one local preview is enabled by default
   const pathname = isBeta() ? getRouterBasename() : '';
```
florkbr, line 48: @Hyperkid123 I'm assuming once ready - we can delete the `isBeta` function to flush out any locations like this?

src/utils/common.ts
```diff
@@ -200,11 +202,17 @@ export function isProd() {
   return location.host === 'cloud.redhat.com' || location.host === 'console.redhat.com' || location.host.includes('prod.foo.redhat.com');
 }
 
+/**
+ * @deprecated preview flag is now determined via chrome internal state variable
```
florkbr, line 206: 👍

src/state/atoms/releaseAtom.ts
```diff
@@ -1,5 +1,18 @@
...
+import { getUnleashClient, unleashClientExists } from '../../components/FeatureFlags/unleashClient';
 
-import { isBeta } from '../../utils/common';
-
-export const isPreviewAtom = atomWithToggle(isBeta());
+export const isPreviewAtom = atomWithToggle(undefined, (isPreview) => {
+  // Required to change the `isBeta` function return value in the visibility functions
+  if (visibilityFunctionsExist()) {
+    updateVisibilityFunctionsBeta(isPreview);
+    axios.post('/api/chrome-service/v1/user/update-ui-preview', { uiPreview: isPreview });
```
florkbr, line 10: @Hyperkid123 should we have some sort of error handling here?

src/components/Header/PreviewAlert.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  function handlePreviewToggle() {
+    if (!LOCAL_PREVIEW) {
+      switchRelease(isPreviewEnabled, location.pathname, previewEnabled);
+    }
+    togglePreview();
+  }
+
+  return shouldRenderAlert ? (
+    <HeaderAlert
```
florkbr, line 31: FWIW Github diff doesn't show - but this came from here originally: https://github.com/RedHatInsights/insights-chrome/pull/2835/files#diff-2e39c3da8c6be596802ae76544d42142fa129a8faf2e3406183c38b34b95dc73L383-L408

src/components/Header/HeaderAlert.tsx
```diff
@@ -45,6 +45,9 @@ const HeaderAlert = ({
   const onClose = () => {
     onDismiss && onDismiss();
     setAlertVisible(false);
+    if (timer) {
+      clearTimeout(timer);
```
florkbr, line 49: This doesn't seem related to any of the preview changes, but nonetheless a good change 👍

src/components/FeatureFlags/unleashClient.ts
```diff
@@ -0,0 +1,27 @@
+import { UnleashClient } from '@unleash/proxy-client-react';
+
+let unleashClient: UnleashClient;
```
florkbr, line 3: FWIW some of this came from the provider here: https://github.com/RedHatInsights/insights-chrome/pull/2835/files#diff-4a88bdbf8236baa86cca130fd1db3eecce2e43da7bf613914792361d031b728bL54-L62

👍 for separating it out

src/components/Header/Tools.tsx
```diff
@@ -29,20 +28,26 @@ import { toggleNotificationsDrawer } from '../../redux/actions';
...
 import { isPreviewAtom } from '../../state/atoms/releaseAtom';
-import chromeStore from '../../state/chromeStore';
+import PreviewAlert from './PreviewAlert';
+
+const LOCAL_PREVIEW = localStorage.getItem('chrome:local-preview') === 'true';
 
 const isITLessEnv = ITLess();
 
+/**
+ * @deprecated Switch release will be replaces by the internal chrome state variable
```
florkbr, line 38:
```suggestion
 * @deprecated Switch release will be replaced by the internal chrome state variable
```

src/utils/initUserConfig.ts
```diff
@@ -0,0 +1,39 @@
...
+export type ChromeUserConfig = {
+  data: {
+    uiPreview: boolean;
+  };
+};
+
+export const initChromeUserConfig = async ({ getUser, token }: { getUser: () => Promise<ChromeUser>; token: string }) => {
+  const LOCAL_PREVIEW = localStorage.getItem('chrome:local-preview') === 'true';
+  let config: ChromeUserConfig;
+  if (!LOCAL_PREVIEW) {
```
florkbr, line 16: @Hyperkid123 we should probably also throw a `// FIXME` here once local preview is turned on permanently.

florkbr (commented):
@Hyperkid123 posted some comments. I'll perform some manual testing on this tomorrow and report how it goes.

src/chrome/create-chrome.ts
Hyperkid123: Yep. We have to keep this, probably for ever same as we did with the chrome API on the window. We can update the TS in the types package once we rollout the change and have proper docs.

src/auth/OIDCConnector/utils.ts
Hyperkid123: Exactly. We will no longer need the `/preview` basename.

src/state/atoms/releaseAtom.ts
Hyperkid123: We should. I'll update the PR

florkbr:
@Hyperkid123 upon checking out the latest commit in this PR I'm seeing quite a few various errors trying to run locally:
Screenshot 2024-05-22 at 5 54 39 PM
Screenshot 2024-05-22 at 5 52 16 PM
image

This was in firefox. I'm also seeing errors in chrome:
image

I'm not sure if these are actually related to your PR... I will continue debugging locally.

Hyperkid123:
Hmm, interesting. I have not experienced these. Were you just clicking around the UI? Any steps to reproduce on my side?

src/components/Header/Tools.tsx
```diff
@@ -241,7 +247,12 @@ const Tools = () => {
     },
     {
       title: betaSwitcherTitle,
-      onClick: () => (window.location.href = switchRelease(isBeta(), location.pathname, previewEnabled)),
+      onClick: () => {
+        if (!LOCAL_PREVIEW) {
```
florkbr, line 251: @Hyperkid123 this `onClick` is identical to the `onChange` below. Can we de-dupe them in a helper function?

src/state/atoms/releaseAtom.ts
```diff
@@ -1,5 +1,22 @@
...
+      axios.post('/api/chrome-service/v1/user/update-ui-preview', { uiPreview: isPreview });
+    }
+    if (unleashClientExists()) {
+      // Required to change the `platform.chrome.ui.preview` context in the feature flags, TS is bugged
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      getUnleashClient().updateContext({ 'platform.chrome.ui.preview': isPreview });
+    }
+  } catch (error) {
+    console.error('Failed to update the visibility functions or feature flags context', error);
```
florkbr, line 20: @Hyperkid123 thanks for adding some error handling - however are we sure this actually catches the errors that can occur above? 

In codepen I tried this snippet of code with the axios library - and the catch never executes (even when a 404 occurs to the API being post-ed):

```js
try {
  console.info('Trying the post');
  axios.post('/foo/bar', { fooBar: 'yes' });
} catch(e) {
  console.warn('I was caught');
  console.warn(e);
}
```
Which results in:
image

I think we may need to add the catch to the promise itself?

src/utils/useNavigation.ts
florkbr: @Hyperkid123 thoughts ☝️ ?

florkbr:
@Hyperkid123 with your latest commit I see the preview notification sometimes flash with the old state before the page reloads (without the local storage flag enabled):

preview-dialog-flash

It doesn't happen all the time, but if I stay on the page for a moment and open/close the all-services menu it seems to happen. It can be confusing as when I click to enable preview is briefly shows "preview disabled" before reloading the page, or vice versa.

src/state/atoms/releaseAtom.ts
florkbr: I also confirmed it's not being hit in chrome by blocking the network request for update-preview chrome service API: 
image

florkbr:
Been trying things in the UI with `localStorage.setItem('chrome:local-preview', true);` and so far so good

src/state/atoms/releaseAtom.ts
Hyperkid123: Magnificent. I though we only had to be paranoid about fetch not handling 4xx

Hyperkid123: Oh its because the atom is not async itself. Easy fix.

Hyperkid123:
> @Hyperkid123 with your latest commit I see the preview notification sometimes flash with the old state before the page reloads (without the local storage flag enabled):

This is because of the mix of state and localstorage. I'll be honest. I don't care enough about the flash. It will be fixed by ending the old preview.

src/state/atoms/releaseAtom.ts
florkbr: Verified this now works. Thanks!

florkbr (approved):

Hyperkid123:
What a perfect test run that fails on some OS random error:
```
[2024-06-19T07:53:54.845Z] [303:0619/075223.845745:ERROR:object_proxy.cc(590)] Failed to call method: org.freedesktop.portal.Settings.Read: object_path= /org/freedesktop/portal/desktop: unknown error type: 


```

PR #2836, opened 2024-05-15
Add ip address whitelist UI to manage satellites page for itless

florkbr:
**Depends on:** https://github.com/RedHatInsights/chrome-service-backend/pull/668

Utilizes mbop changes here: https://github.com/RedHatInsights/mbop/pull/105

Adds a simple UI to the manage satellites page to add/remove IP address(es) WRT to our itless environments.

~Marking as draft until I manually try the API requests in this PR via postman on pre-prod env.~
EDIT: verified API requests after deploying


Hyperkid123:
@florkbr do we want this to be baked into chrome? Can this be its own module?

florkbr:
@Hyperkid123 I don't think it necessarily needs to be baked into chrome - I just started from where the UI lived today.

Do you have any suggestions or thoughts on where the new module could/should be placed?

Hyperkid123:
@florkbr a new module :) Like the widget UI.

florkbr:
When deploying this in our itless env I'm seeing the same pod crash loop as noted in: https://github.com/RedHatInsights/insights-chrome/pull/2922

src/layouts/SatelliteToken.tsx
```diff
@@ -38,11 +47,6 @@ const SatelliteToken: React.FC = () => {
       <Page
         className="chr-c-all-services"
         onPageResize={null} // required to disable PF resize observer that causes re-rendring issue
-        header={
-          <Masthead className="chr-c-masthead">
```
florkbr, line 42: @Hyperkid123 since we are now exposing this as a module from chrome at `/insights/satellite` - this is causing the insights masthead and nav to render around the module (hence removing the masthead here otherwise we see 2).

frontend.yml
```diff
@@ -19,6 +19,11 @@ objects:
         manifestLocation: "/apps/chrome/js/fed-mods.json"
         config:
           ssoUrl: ${SSO_URL}
+        modules:
+          - id: 'satellite-token'
```
florkbr, line 23: @Hyperkid123 I noticed this is duplicated with my changes in chrome-service-backend: https://github.com/RedHatInsights/chrome-service-backend/pull/668. I know this is used for the frontend-operator - but can you explain why we need both? Thanks.

src/components/RootApp/ScalprumRoot.tsx
```diff
@@ -259,7 +258,6 @@ const ScalprumRoot = memo(
                   }
                 />
               )}
-              {ITLess() && <Route path="/insights/satellite" element={<SatelliteToken />} />}
```
florkbr, line 262: Removed the react router route in ScalprumRoot in favor of the exposed module approach.

frontend.yml
Hyperkid123: We don't need both. Only the chrome service. This will become relevant and the source of truth once we have the new FEO features.

src/layouts/SatelliteToken.tsx
Hyperkid123: Does this page require some unique layout?

Hyperkid123 (approved):

src/layouts/SatelliteToken.tsx
florkbr: No - the content is just a bit odd/out of the ordinary as it is 3 utilities.

florkbr:
@Hyperkid123 I removed the changes to `frontend.yml` for now. Let me know if you think this needs anything else! I'll likely still need to follow up with some styling tweaks on the IP whitelist UI itself - but the major flows seems to be working for me.

Hyperkid123 (approved):
LGTM

florkbr:
The PR build pipeline is still failing due to the `CrashLoopBackOff` issue mentioned earlier. I've started a thread in our dev-prod channel and will continue investigating on https://github.com/RedHatInsights/insights-chrome/pull/2922

PR #2844, opened 2024-05-27
Identify all outgoing requests with extra frontend header.

Hyperkid123:
PR: https://issues.redhat.com/browse/RHCLOUD-32526


src/utils/iqeEnablement.ts
```diff
@@ -11,6 +11,8 @@ import type { AuthContextProps } from 'react-oidc-context';
 let xhrResults: XMLHttpRequest[] = [];
 let fetchResults: Record<string, unknown> = {};
 
+// this extra header helps with API metrics
+const FE_ORIGIN_HEADER_NAME = 'x-rh-frontend-origin';
```
aneelac22, line 15: This will effect FedRamp

Hyperkid123: What is the effect? Is there only a set of allowed headers?

florkbr: I tested this new header on our pre-prod env and haven't seen any issues. Since I'm aware this change is being made I'll keep an eye out for any issues and can follow up with a gate if required.

florkbr (approved):

aneelac22 (approved):

PR #2856, opened 2024-06-03
Enable marking as read in the notification drawer

CodyWMitchell:
Made the following updates to the notification drawer:
- Selecting and marking notifications as read are now two distinct actions.
- 'Mark visible as read' has been updated to 'Mark selected as read'. Same for 'Mark selected as unread'.  [RHCLOUD-28581](https://issues.redhat.com/browse/RHCLOUD-28581)
- Marking notifications as read/unread now sends the associated request to the API
- Ability to bulk select/deselect notifications [RHCLOUD-32794](https://issues.redhat.com/browse/RHCLOUD-32794)
- Updated tests to match the changes above.


src/components/NotificationsDrawer/DrawerPanelContent.tsx
```diff
@@ -96,8 +105,9 @@ const DrawerPanelBase = ({ innerRef }: DrawerPanelProps) => {
   }, []);
 
   useEffect(() => {
+    console.log('activeFilters', activeFilters);
```
karelhala, line 108: Left over `console.log`
```suggestion
```

src/components/NotificationsDrawer/DrawerPanelContent.tsx
```diff
@@ -109,14 +119,53 @@ const DrawerPanelBase = ({ innerRef }: DrawerPanelProps) => {
...
+        read_status: false,
+      })
+      .then(() => {
+        dispatch(deselectAllNotifications());
+      })
+      .catch((error) => {
+        dispatch(markSelectedNotificationsAsRead());
+        console.error('Failed to update notification read status', error);
+      });
+  };
```
karelhala, line 156: It's not ideal to dispatch action and then run async function. Ideally we can have action in `/src/redux/actions` that dispatches async function. Once it gets back to us we can update `/src/redux/index` to include new reducer and update the selected events in there.

Hyperkid123 (commented):
Here is a late comment. Can you not use redux? We are moving away to `Jotai` for state managment.

PR #2871, opened 2024-06-20
Update documentation on how to run with local services

randomnetcat:
The documented instructions do happen to work for running Advisor (not sure why?) but I could not convince them to work with other applications (I tested learning-resources and registration-assistant). They result in the error:
```
Registration is not running or available via http://stage.foo.redhat.com:8004

Make sure to run `npm run start -- --port=8004` in the registration application directory to start it's webpack dev server and the bundle is built.
```

Based on the `static` script in [learning-resources](https://github.com/RedHatInsights/learning-resources/blob/7c82a8c60cc51e3d09bc5a479d8ef81d1ee1e655/package.json#L17), I found that `npx fec static --port=NNNN` did work for all repositories.

Also, I found that the separator between the port and protocol is documented incorrectly; that is fixed here.


bastilian:
@randomnetcat The `-- --port=0909` behaviour is broke atm. However, it will be fixed as soon as this PR is merged: https://github.com/RedHatInsights/frontend-components/pull/2007

Then the documentation should again be valid. 

What you can do meanwhile is to add a `port` prop with the desired port to the `fec.config.js`. The part with `LOCAL_APPS` is still valid and should work as expected.

randomnetcat:
> The -- --port=0909 behaviour is broke atm. However, it will be fixed as soon as this PR is merged: https://github.com/RedHatInsights/frontend-components/pull/2007

Ah, thank you, sorry.

> The part with LOCAL_APPS is still valid and should work as expected.

The example happens to work because the protocol it attempts to specify is `http`, and that's already the default. See https://github.com/RedHatInsights/frontend-components/blob/fc61adcd088fe25997c36b8ef799f37dd7691739/packages/config-utils/src/proxy.ts#L70-L72

PR #2872, opened 2024-06-24
Add package that displays a message for oudated browser

randomnetcat:
See [RHCLOUD-27572](https://issues.redhat.com/browse/RHCLOUD-27572).

The `outdatedBrowser()` call does accept an argument that tells it what browser versions to consider outdated, but I wasn't sure how to even start approaching determining that. For instance, the Patternfly [documentation](https://www.patternfly.org/get-started/about-patternfly) says the supported version for all browser is "latest", which is... less than helpful.


src/outdated-browser-rework.js
```diff
@@ -0,0 +1,5 @@
+import outdatedBrowser from 'outdated-browser-rework';
+import 'outdated-browser-rework/dist/style.css';
+
+// Use the default versions until we have something meaningful to add?
```
Hyperkid123, line 4: We might want to follow the PF support. https://www.patternfly.org/get-started/about/#supported-browsers If PF is not working, the UI won't work.

src/outdated-browser-rework.js
Hyperkid123, line 1:
```
Processed 318 files (1.5s) (212 warnings)

✖ Found 1 circular dependency!

1) outdated-browser-rework.js

```

This is a little bit dumb, but. Because the the filename is the same as the dependency name, the circular dependency tool thinks its referencing itself. And the CI flips out. Can you rename it please?

config/webpack.config.js
```diff
@@ -56,6 +57,7 @@ const commonConfig = ({ dev }) => {
         }
       : {}),
     devtool: false,
+    target: ['web', 'es4'],
```
Hyperkid123, line 60: Why do we need the `es4`?

randomnetcat: Without it, Webpack will generate the `runtime` chunk using, at the very least, an arrow function, which is an ES6 feature. I figured it was good to set this as low as possible to maximize the chances old browsers at least get to display the outdated browser warning. I'm happy to remove it if you'd prefer.

src/outdated-browser-rework.js
ryelo: if we're moving to just "latest", we should also update https://github.com/RedHatInsights/insights-chrome/blob/d9cba3c94d516a28935796bc79e54a349cb873b8/src/components/Footer/Footer.tsx#L83 so we're not having two different answers

config/webpack.config.js
Hyperkid123: I think we can keep the web, but I'd remove the es4. It's way too old and PF is not supporting these browsers anyway. We won't lose any users over this.

randomnetcat: Ah, sorry. That's the default, so I'll just remove it.

Hyperkid123:
Seems like the extensions is messing with e2e cypress/iqe tests:

image

randomnetcat:
It appears Cypress is configured to use an ancient User-Agent https://github.com/RedHatInsights/insights-chrome/blob/d17fb3de33d95128a92c8b2fd0f20fa743d50375/cypress.config.ts#L46-L47 . Can that be changed at this point? Not sure what the "chromium issue" is.

ryelo:
We should run the content and banner by UX do double check brand standards. I know from looking the banner doesn't follow PatternFly guidelines and would like to see if they have guidance for us.

Hyperkid123:
@randomnetcat 
> Can that be changed at this point? Not sure what the "chromium issue" is.

It might be. It's very likely the lib is looking at this. its worth a shot. If that says its chrome 51 its basically an ancient version. Current is 120 or something.

randomnetcat:
Updated the userAgent. I still have a test failing locally, but it appeared to also be failing on `master`?

Hyperkid123:
/retest

There is always one run failing

PR #2875, opened 2024-06-27
Filter in notifications drawer

CodyWMitchell:
[RHCLOUD-29014](https://issues.redhat.com/browse/RHCLOUD-29014)
- Enable filtering in the notifications drawer
- Fetch correct filters from `/api/notifications/v1/notifications/facets/bundles`


src/components/NotificationsDrawer/DrawerPanelContent.tsx
```diff
@@ -38,6 +37,18 @@ import BulkSelect from '@redhat-cloud-services/frontend-components/BulkSelect';
 import axios from 'axios';
 import { Stack, StackItem } from '@patternfly/react-core/dist/dynamic/layouts/Stack';
 
+interface Bundle {
+  id: string;
+  name: string;
+  displayName: string;
+  children: null | Bundle[];
```
Hyperkid123, line 44: I think children can be optional `children?: Bundle[]`. Or is there is a specific need for the `null` value?

src/components/NotificationsDrawer/DrawerPanelContent.tsx
```diff
@@ -124,7 +136,23 @@ const DrawerPanelBase = ({ innerRef }: DrawerPanelProps) => {
         );
       }
     };
+    const fetchFilterConfig = async () => {
```
Hyperkid123, line 139: Why is this function defined in `useEffect`?

CodyWMitchell: I was following how the permissions were fetched after `mounted` became true in the `fetchPermissions` method.

src/components/NotificationsDrawer/DrawerPanelContent.tsx
CodyWMitchell: Good catch! I think you're right.

src/components/NotificationsDrawer/DrawerPanelContent.tsx
Hyperkid123: I see. The pattern used before is wrong. The mounted variable should be in ref so we don't have to do this. Its something for a future refactoring I guess.

Hyperkid123 (approved):

PR #2880, opened 2024-06-28
Use stricter to rule to patch hac link targets.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-33375

Use stricter regexp rather than just checking the URL includes the `/hac` partial.


aneelac22:
Lint failures `'MemoryRouter' is defined but never used`

PR #2881, opened 2024-06-28
fix(RHINENG-10654): Reimplement Sentry 

adonispuente:
This reconfigures sentry so that it handles multiple frontends properly with chrome being the wrapper.
Currently there are a few changes I need to make in the settings of the repo itself to add secret keys to the environment, as well as a CI/CD addition, but these changes will allow the projects to be better monitored. I dont have perms for this and will ping platform team in regards to the set up aspect, those changes must be merged for this to work at all

This is step 1, but by the end, ideally, the cp-001 project will only report errors sentry deems is from chrome, and all the apps will have their own projects where we can monitor the health of specific apps. 
All the apps would need to do after this is done is have a similar PR like this : https://github.com/RedHatInsights/insights-advisor-frontend/pull/1268/files
and then some CI/CD changes + secret variables added

https://issues.redhat.com/browse/RHINENG-10654


config/webpack.config.js
```diff
@@ -44,6 +44,7 @@ const commonConfig = ({ dev }) => {
       publicPath,
       chunkFilename: dev ? '[name].js' : '[name].[contenthash].js',
     },
+    devtool: process.env.SENTRY_AUTH_TOKEN ? 'source-map' : false,
```
Hyperkid123, line 47: This needs to be `hidden-source-map` so there is no reference in the source code and we do not load sour maps via Akamai.

config/webpack.plugins.js
```diff
@@ -86,6 +87,18 @@ const plugins = (dev = false, beta = false, restricted = false) => {
       __SENTRY_DEBUG__: false,
     }),
     ...(dev ? [new ReactRefreshWebpackPlugin()] : []),
+    // Put the Sentry Webpack plugin after all other plugins
+    sentryWebpackPlugin({
+      authToken: process.env.SENTRY_AUTH_TOKEN,
```
Hyperkid123, line 92: What if the token does not exist?

config/webpack.config.js
adonispuente: I dont have perms yet but the goal would be to build the source maps in the CI/CD set up ,send to sentry every time theres a merge, and then delete it, so the build never makes it to akamai/served to customers.
Similar to this: https://github.com/RedHatInsights/insights-advisor-frontend/blob/cdd23204e7285353094ef1333c12efc13f8983b3/.github/workflows/webpack.yml

config/webpack.plugins.js
adonispuente: I will create a secret variable for this, so itll essentially always be there when we merge, but if a user doesnt want to build it locally every time then it wont build

config/webpack.config.js
Hyperkid123: > and then delete it,

The problem is, that the source map is still referenced in the source file, and the browser will fire a request. The fact the response will be a 404 is even worse than if it was a 200. It will always skip the cache nodes (because the asset does not exist there) and will go directly to our frontend containers.

We don't want Chrome to reference the source maps at all.

adonispuente: I see. Making adjustments

Hyperkid123: It can't be in the build step which is publishing the container. That is the hard limit.

adonispuente: This has been updated to use hidden-source-maps instead

config/webpack.plugins.js
Hyperkid123: Can we just not include the plugin if the env variable does not exist? And gate it behind the `SENTRY_DSN` as well?

src/utils/sentry.ts
```diff
@@ -61,17 +61,49 @@ function initSentry() {
...
+      Sentry.replayIntegration({ maskAllText: false, maskAllInputs: true }),
+      SentryBrowser.moduleMetadataIntegration(),
+    ],
     tracesSampleRate: 0.2,
     debug: !!window.localStorage.getItem('chrome:sentry:debug'),
     replaysOnErrorSampleRate: 1.0,
     replaysSessionSampleRate: 0.3,
+    transport,
+    beforeSend: (event) => {
+      if (event?.exception?.values?.[0]?.stacktrace?.frames) {
```
Hyperkid123, line 82: Are the frames always limited to the first entry? Or can a relevant frame be in any exception? Furthermore, is it possible that more than one event should be modified?

src/utils/sentry.ts
```diff
@@ -61,17 +61,49 @@ function initSentry() {
...
+            [EXTRA_KEY]: routeTo,
+          };
+        }
+      }
+
+      return event;
+    },
+    //These two apps will not be set up as of now. This helps limit transacations
+    beforeSendTransaction: (event) => {
+      if (event?.contexts?.app?.app_name === 'subscriptions' || event.contexts?.app?.app_name === 'image-builder') {
```
Hyperkid123, line 102: I'd like these to be in some variables, potentially configured via Chrome service. Store it in an array for now, and check if the app_name value is included.

Hyperkid123 (changes requested):
Exposed federated modules are failing to be build. Here is the common error:
```
ERROR in ./src/components/FavoriteServices/DashboardFavorites.tsx
Module build failed (from ./node_modules/swc-loader/src/index.js):
Error: Bindings not found.
    at Compiler.<anonymous> (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:227:19)
    at Generator.next (<anonymous>)
    at /home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:31:71
    at new Promise (<anonymous>)
    at __awaiter (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:27:12)
    at Compiler.transform (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:203:16)
    at Object.transform (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:352:21)
    at Object.<anonymous> (/home/martin/insights/insights-chrome/node_modules/swc-loader/src/index.js:97:21)
 @ container entry ./DashboardFavorites[0]

```

src/utils/sentry.ts
adonispuente: I spoke with the sentry folks and they confirmed it should always be the first entry. We wouldnt need to modify any other event. This section is just looking at the last frame to find witch project to send it too.

adonispuente:
> Exposed federated modules are failing to be build. Here is the common error:
> 
> ```
> ERROR in ./src/components/FavoriteServices/DashboardFavorites.tsx
> Module build failed (from ./node_modules/swc-loader/src/index.js):
> Error: Bindings not found.
>     at Compiler.<anonymous> (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:227:19)
>     at Generator.next (<anonymous>)
>     at /home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:31:71
>     at new Promise (<anonymous>)
>     at __awaiter (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:27:12)
>     at Compiler.transform (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:203:16)
>     at Object.transform (/home/martin/insights/insights-chrome/node_modules/@swc/core/index.js:352:21)
>     at Object.<anonymous> (/home/martin/insights/insights-chrome/node_modules/swc-loader/src/index.js:97:21)
>  @ container entry ./DashboardFavorites[0]
> ```

I believe the issue is the tsconfig, "sourceRoot": "/",
This is needed for sentry to map against hidden source maps. Looking into the most non-hacky solution i can come up with.

adonispuente:
ci keeps failing because it cant find the container for this PR

adonispuente:
converting to draft for a bit. I need to find a better way to map the source maps. there are too many with no refs

adonispuente:
it fails because the package.json file is missing from the directory /opt/app-root/src/ inside the container. Has this happened before? @Hyperkid123

Hyperkid123:
@adonispuente you can ignore the Jenkins issues. The pipeline is broken nd will be fixed once we migrate to konflux.

BlakeHolifield:
Just a heads up, this PR seems to be causing a webpack-dev-server issue and [looping forever in some jenkins jobs](https://ci.int.devshift.net/job/RedHatInsights-insights-chrome-pr-check/2067/console). 

``` bash
12:21:01 [webpack-cli] TypeError: Cannot read properties of undefined (reading 'constructor')
12:21:01     at /home/tester/workspace/node_modules/webpack-dev-server/lib/Server.js:1860:22
12:21:01     at Array.find (<anonymous>)
12:21:01     at /home/tester/workspace/node_modules/webpack-dev-server/lib/Server.js:1859:60
12:21:01     at Array.forEach (<anonymous>)
12:21:01     at Server.initialize (/home/tester/workspace/node_modules/webpack-dev-server/lib/Server.js:1846:17)
12:21:01     at Server.start (/home/tester/workspace/node_modules/webpack-dev-server/lib/Server.js:3358:16)
12:21:01     at async Command.<anonymous> (/home/tester/workspace/node_modules/@webpack-cli/serve/lib/index.js:158:21)
12:21:01     at async Command.parseAsync (/home/tester/workspace/node_modules/commander/lib/command.js:935:5)
12:21:01     at async Command.<anonymous> (/home/tester/workspace/node_modules/webpack-cli/lib/webpack-cli.js:1356:13)
12:21:01     at async Command.parseAsync (/home/tester/workspace/node_modules/commander/lib/command.js:935:5)
12:21:03 making HTTP(S) head request to  url:https://127.0.0.1:1337/webpack-dev-server ...
12:21:03 (node:82) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS

```

adonispuente:
@Hyperkid123 I believe we can merge this if we're ignoring the ci test. I connected with @BlakeHolifield about the looping and we both believe its unrelated to the PR

Hyperkid123:
@adonispuente we can't ignore the CI. Only the Jenkins jobs, the rest must be green.

adonispuente:
@Hyperkid12 Everything else is passing other than synk. I dont have access to view why its failing. Can you give me perms?

Hyperkid123:
> @Hyperkid12 Everything else is passing other than synk. I dont have access to view why its failing. Can you give me perms?

It's just a random thing in package.json nothing you should need to be worried about.

config/webpack.plugins.js
```diff
@@ -87,6 +88,19 @@ const plugins = (dev = false, beta = false, restricted = false) => {
       __SENTRY_DEBUG__: false,
     }),
     ...(dev ? [new ReactRefreshWebpackPlugin()] : []),
+    // Put the Sentry Webpack plugin after all other plugins
+    process.env.SENTRY_AUTH_TOKEN &&
```
Hyperkid123, line 92: Adding a plugin conditionally like this will append `false` or `undefined` to the array. The condition should also verify that all required variables exist in the environment.

Hyperkid123 (changes requested):
Something is wrong with the build process. Here is how I can reproduce issues in my machine.

1. run `npm ci`
2. run `npm run build`
3. run `npm i`
4. run `npm run build`

If I try to run build after regular dependency install, the build fails with:
```
assets by status 9.45 KiB [cached] 8 assets
runtime modules 9.36 KiB 14 modules
built modules 237 bytes [built]
  modules by path ./src/components/FavoriteServices/*.tsx 78 bytes
    ./src/components/FavoriteServices/LandingNavFavorites.tsx 39 bytes [built] [code generated] [1 error]
    ./src/components/FavoriteServices/DashboardFavorites.tsx 39 bytes [built] [code generated] [1 error]
  ./src/index.ts 39 bytes [built] [code generated] [1 error]
  container entry 42 bytes [built] [code generated]
  ./src/pdf/DownloadButton.tsx 39 bytes [built] [code generated] [1 error]
  ./src/layouts/SatelliteToken.tsx 39 bytes [built] [code generated] [1 error]

ERROR in ./src/components/FavoriteServices/DashboardFavorites.tsx
Module build failed (from ./node_modules/swc-loader/src/index.js):
Error: Bindings not found.
    at Compiler.<anonymous> (/Users/martin/insights/insights-chrome/node_modules/@swc/core/index.js:227:19)
    at Generator.next (<anonymous>)
    at /Users/martin/insights/insights-chrome/node_modules/@swc/core/index.js:31:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/martin/insights/insights-chrome/node_modules/@swc/core/index.js:27:12)
    at Compiler.transform (/Users/martin/insights/insights-chrome/node_modules/@swc/core/index.js:203:16)
    at Object.transform (/Users/martin/insights/insights-chrome/node_modules/@swc/core/index.js:352:21)
    at Object.<anonymous> (/Users/martin/insights/insights-chrome/node_modules/swc-loader/src/index.js:97:21)
 @ container entry ./DashboardFavorites[0]

...
```

Until this is eliminated, I can't merge the PR.

adonispuente:
@Hyperkid123 After investigating the issue you described, this seems to be an issue with master branch. Currently on master,

Deleteing package.lock file and then running `clear && rm -rf node_modules && npm i && npm run build` and then trying to run cypress test ` npm run test:ct`,  you will encounter the errors you mentioned above as well as some other issues. The swc issue seems to be unrelated to this PR.

Hyperkid123:
@adonispuente we do not have the build issues on other branches. The changes to the build must be the root cause. The cypress component tests have their own webpack configuration. I suggest trying to poking awround there.

adonispuente:
@Hyperkid123 I believe we're ok now. I wasnt able to replicate the errors you mentioned after running your flow with this version of the PR.

Hyperkid123:
@adonispuente when i run `npm i` I still have some changes in the lock file. Can you update it?

Other than that I think we are good now. let me just do one more pass over the code.

config/webpack.config.js
```diff
@@ -52,7 +52,7 @@ const commonConfig = ({ dev }) => {
           },
         }
       : {}),
-    devtool: false,
+    devtool: 'hidden-source-map',
```
Hyperkid123, line 55: I'd like to see this only in production mode. No need to build it during dev.

config/webpack.plugins.js
```diff
@@ -87,6 +88,22 @@ const plugins = (dev = false, beta = false, restricted = false) => {
       __SENTRY_DEBUG__: false,
     }),
     ...(dev ? [new ReactRefreshWebpackPlugin()] : []),
+     // Put the Sentry Webpack plugin after all other plugins
+     ...(process.env.SENTRY_AUTH_TOKEN && process.env.SENTRY_ORG && process.env.SENTRY_PROJECT
```
Hyperkid123, line 92: Do not add this in dev mode. Does nothing but slow the build.

Hyperkid123 (approved):

PR #2885, opened 2024-07-01
Disable VA for itless envs

florkbr:


src/layouts/DefaultLayout.tsx
```diff
@@ -100,7 +101,7 @@ const DefaultLayout: React.FC<DefaultLayoutProps> = ({ hasBanner, selectedAccoun
           </div>
         )}
         <RedirectBanner />
-        <VirtualAssistant />
+        {ITLess() ? null : <VirtualAssistant />}
```
Hyperkid123, line 104: Hmm. OK for now, but we should do this via chrome service configuration in k8s

Hyperkid123 (approved):

PR #2897, opened 2024-07-15
RHCLOUD-27262: Add search feedback buttons

florkbr:
image


src/components/Search/SearchInput.tsx
```diff
@@ -35,13 +37,6 @@ const getMaxMenuHeight = (menuElement?: HTMLDivElement | null) => {
   return bodyHeight - menuTopOffset - 4;
 };
 
-type SearchItem = {
```
florkbr, line 38: Moved to `SearchTypes.ts` to prevent circular dependency in the new `SearchFeedback` component.

src/components/Search/SearchInput.tsx
```diff
@@ -197,8 +192,18 @@ const SearchInput = ({ onStateChange }: SearchInputListener) => {
...
+    menuFooter = (
+      <MenuFooter className="pf-v5-u-px-md">
+        <SearchFeedback query={searchValue} results={searchItems} />
+      </MenuFooter>
+    );
+  }
+
   const menu = (
-    <Menu ref={menuRef} className="pf-v5-u-pt-sm pf-v5-u-px-md chr-c-search__menu">
+    <Menu ref={menuRef} className="pf-v5-u-pt-sm chr-c-search__menu">
```
florkbr, line 206: This width utility class moved to the menu group below (and is also present on the menu footer) to get us closer to the end design in our mockups. Some additional styling will be required later.

src/components/Search/SearchFeedback.tsx
```diff
@@ -0,0 +1,40 @@
...
+  query: string;
+  results: SearchItem[];
+};
+
+const SearchFeedback = ({ query, results }: SearchFeedbackProps) => {
+  const { ready, analytics } = useSegment();
+  const trackFeedback = (type: string | SegmentEvent) => {
+    ready && analytics && analytics.track(type, { query, results });
+  };
+  const throttledTrackFeedback = throttle(trackFeedback, 5000);
```
florkbr, line 23: An alternative to this approach would be to only allow feedback once per query string, and feedback cannot be submitted again until the user types in a new query.

Hyperkid123 (approved):

src/components/Search/SearchFeedback.tsx
florkbr: @Hyperkid123 agree on both points. I will follow up with another PR. Thanks

src/components/Search/SearchFeedback.test.tsx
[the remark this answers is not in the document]
florkbr: @Hyperkid123 I don't completely disagree - I think I like JSDOM tests for ensuring functions are called when expected, do not throw errors or handle errors, have the correct signature, etc. But when it comes to a user clicking a button - I agree cypress component tests actually trigger all events the user would by clicking said button.

PR #2901, opened 2024-07-19
Consume new chrome API itless type in chrome creation

florkbr:
DEPENDS ON: https://github.com/RedHatInsights/frontend-components/pull/2038


florkbr:
Closing as we should remove this logic and not just rename/update.

PR #2908, opened 2024-07-24
Decrease spacing between search item title/description

randomnetcat:
As described by the document linked in the comments of [RHCLOUD-33948](https://issues.redhat.com/browse/RHCLOUD-33948).

This should also fix the bug (described in the same document) where the search title starts spanning three lines (name, pipe, then bundle), which appears to be caused by Javascript loading styles that make the TextContent styling a better match than our overrides.


src/components/Search/SearchTitle.tsx
```diff
@@ -1,18 +1,21 @@
 import React from 'react';
-import { Text, TextContent } from '@patternfly/react-core/dist/dynamic/components/Text';
+import './SearchTitle.scss';
 
-const SearchTitle = ({ title, bundleTitle }: { title: string; bundleTitle: string }) => {
+const SearchTitle = ({ title, bundleTitle, className = '' }: { title: string; bundleTitle: string; className: string }) => {
```
Hyperkid123, line 4: If the classname can be `''` the prop should be optional `className?: string`

Hyperkid123 (commented):
nit

PR #2914, opened 2024-08-07
Use mfe plugin for chrome shell

Hyperkid123:
Experimenting with decoupled module federation from webpack.

Notes:
- Refresh plugin needs to be reconfigured in order to not cause false positive errors during shell init phase
- There is an issue with shared scope between MFE and Vanilla plugins, seems to be some sort of race conditions, sometimes the child modules are unable to access singleton instances. Problematic for React context and hooks


Hyperkid123:
There is an issue with exposed modules coming directly from chrome.

PR #2921, opened 2024-08-20
IQE-3075: Add an OUIA ID for iqe-core to consume

catastrophe-brandon:
iqe-core wants the nav to automagically click the preview switch when `use_beta` is set to true in the framework config (see https://issues.redhat.com/browse/IQE-3075). As such, we want to use an OUIA ID to be able to uniquely interact with the switch. Aria-label is okay, but it could theoretically change and isn't meant to be used as a unique ID for test automation.


Caddyfile
```diff
@@ -1,6 +1,7 @@
 {
     {$CADDY_TLS_MODE}
     auto_https disable_redirects
+    persist_config off
```
Hyperkid123, line 4: Any reason for this Caddy change?

PR #2922, opened 2024-08-22
[DO NOT MERGE] Caddyfile changes to support removing preview

florkbr:
When deploying insights-chrome to ephemeral via bonfire - we are seeing the chrome frontend pods `CrashLoopBackOff`.

I believe this is due to the caddy process looking for the index.html/JS inside of the `preview` and `stable` directories - when now it is at the root of the build dir (and only `operator-generated` appears in those dirs).


florkbr:
Not necessary now as we have reverted the changes to the frontend-build-container

PR #2925, opened 2024-08-30
AAP-29825 add ansible-cloud allowed origin

acosferreira:
As part of onboard ansible-clouds console dot we are adding the the service as allowed-origin

### DIFF
--- a/src/utils/iqeEnablement.ts
+++ b/src/utils/iqeEnablement.ts
@@ -19,8 +19,7 @@ const AUTH_ALLOWED_ORIGINS = [
   /https:\/\/api(?:\.[a-z]+)?\.openshift(?:[a-z]+)?\.com/,
   /https:\/\/api?\.demo-experience(?:\.[a-z]+)?\.demo?\.redhat?\.com/,
   /https:\/\/api?\.aws?\.ap-southeast-1(?:\.[a-z]+)?\.openshift?\.com/,
-  /https:\/\/console-service?\.[a-z]*-[a-z]*\.aws?\.ansiblecloud?\.com\/api/,
-  /https:\/\/console-service?\.[a-z]*-[a-z]*\.aws?\.ansiblecloud?\.redhat?\.com\/api/,
+  /https:\/\/console-service?\.[a-z]*-[a-z]*\.?(?:[a-z]*)\.aws?\.ansiblecloud?(?:\.redhat)?\.com\/api/,
 ];
 const AUTH_EXCLUDED_URLS = [/https:\/\/api(?:\.[a-z]+)?\.openshift(?:[a-z]+)?\.com\/api\/upgrades_info/];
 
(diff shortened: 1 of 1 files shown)


leena-jawale:
@Hyperkid123 Can you review this PR? Thank you!!

karelhala (approved):
This will auto inject auth header to console-service aws enpoint.

src/utils/iqeEnablement.ts
```diff
@@ -19,6 +19,8 @@ const AUTH_ALLOWED_ORIGINS = [
   /https:\/\/api(?:\.[a-z]+)?\.openshift(?:[a-z]+)?\.com/,
   /https:\/\/api?\.demo-experience(?:\.[a-z]+)?\.demo?\.redhat?\.com/,
   /https:\/\/api?\.aws?\.ap-southeast-1(?:\.[a-z]+)?\.openshift?\.com/,
+  /https:\/\/console-service?\.[a-z]*-[a-z]*\.aws?\.ansiblecloud?\.com\/api/,
+  /https:\/\/console-service?\.[a-z]*-[a-z]*\.aws?\.ansiblecloud?\.redhat?\.com\/api/,
```
florkbr, line 23: @acosferreira a few questions
1. Do we have any documentation surrounding these domains? Such as who owns them, what APIs are surfaced, etc? I think this might help clear up some of my additional questions below and help me understand any security implications with sending auth tokens to these endpoints.
2. If these API's are accessed through AWS (guessing by the URL) - are we sure `[a-z]*-[a-z]*` is sufficient for the region? I believe regions may also contain numerics
3. Is the `.redhat` portion of the URL truly optional? If so - do we need 2 regex's or can we simplify to just one where `.redhat` is an optional non-capturing group in the regex (which should catch both cases) IE: 
```suggestion
  /https:\/\/console-service?\.[a-z]*-[a-z]*\.aws?\.ansiblecloud?(?:\.redhat)?\.com\/api/,
```

FWIW this is just a global helper to append the token to HTTP Fetch calls made to known-safe URLs. Nothing is stopping you from sending the token in your application without these changes.

florkbr (changes requested):

src/utils/iqeEnablement.ts
acosferreira: Thanks for the reply @florkbr.

Those aws are part of ansible-as-service project and  `[a-z]*-[a-z]*`  represents our mgt-env (dev, int ,preprod, prod), not the regions.
For our dev and int environments we are not using .redhat.com, only for preprod and prod, Im also adding a validation to our int env because is ` https://console-service.mgt-int.int.aws.ansiblecloud.com/api/v1/subscriptions/ `

florkbr (approved):
@leena-jawale @acosferreira I want to stress that updating this list of regular expressions is not the preferred solution. We can use the `useChrome` auth/token APIs (see https://github.com/RedHatInsights/frontend-components/blob/master/packages/types/index.d.ts#L155) to retrieve the token and send the token as part of the requests you make to these domains in your application.

In the future we will be moving this list into the respective application configuration and exposing through chrome. When that happens we may need your team to migrate these domains into your application itself.

For now in order to make the onboarding process as smooth as possible I will merge this PR.

florkbr:
ci.int.devshift.net PR build failure is not related to this PR. The unit tests are passing - however we are still running into the bonfire deploy issues where the chrome pod is crash looping. CC @catastrophe-brandon 

See https://github.com/RedHatInsights/insights-chrome/pull/2922

PR #2939, opened 2024-09-19
RHCLOUD-27262: Code review comments from search feedback analytics

florkbr:
Code review follow up from https://github.com/RedHatInsights/insights-chrome/pull/2897.

- Makes it so feedback may only be submitted once per query. 
  - _Note this is for the lifecycle of the SearchInput - we do not check against a DB to see if you've submitted feedback for "foobar" in a previous session. If that is required we can follow up_
- Changes the label to display a message on success or failure on feedback submission
- Highlights the corresponding feedback button (thumbsup/down) in blue
- Updates tests accordingly

image
image

### DIFF
--- a/src/components/Search/SearchFeedback.scss
+++ b/src/components/Search/SearchFeedback.scss
@@ -1,4 +1,4 @@
-.chr-c-search-feedback {
+.chr-c-search-feedback.pf-v5-c-menu__group {
     display: flex;
     justify-content: flex-end;
     align-items: center;
--- a/src/components/Search/SearchFeedback.test.tsx
+++ b/src/components/Search/SearchFeedback.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import SearchFeedback from './SearchFeedback';
+import SearchFeedback, { SEARCH_FEEDBACK_NEGATIVE, SEARCH_FEEDBACK_POSITIVE } from './SearchFeedback';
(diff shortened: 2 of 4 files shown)


src/components/Search/SearchInput.tsx
```diff
@@ -61,6 +62,12 @@ const SearchInput = ({ onStateChange }: SearchInputListener) => {
 
   const resultCount = searchItems.length;
 
+  useEffect(() => {
+    if (currentFeedbackType) {
+      setcurrentFeedbackType(undefined);
+    }
+  }, [searchValue]);
```
florkbr, line 69: This change is necessary as when the results list opens/closes - it resets the state of our feedback widget. Therefore we will track if feedback has been sent or not via the parent.

src/components/Search/SearchFeedback.scss
```diff
@@ -1,4 +1,4 @@
-.chr-c-search-feedback {
+.chr-c-search-feedback.pf-v5-c-menu__group {
```
florkbr, line 1: My selector wasn't specific enough causing some weirdness when interacting with the search results. This resolves the issue.

florkbr:
@Hyperkid123 mind reviewing? Thanks!

src/components/Search/SearchFeedback.test.tsx
Hyperkid123, line 1: Just a note for the future (definitely don't do this now). I'd prefer any tests related to React to be written as cypress component tests. JSDOM sucks.

Hyperkid123 (approved):

PR #2943, opened 2024-09-24
Remove unecessary renders after auth refresh.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-35374

### Changes

Lifting calls to the internal auth context to prevent the tree re-renders. Also memoizing some of the public API objects to prevent extra re-renders.

### DIFF
--- a/src/components/QuickStart/useQuickstartsStates.ts
+++ b/src/components/QuickStart/useQuickstartsStates.ts
@@ -1,40 +1,43 @@
 import axios from 'axios';
-import { useContext, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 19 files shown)


src/components/QuickStart/useQuickstartsStates.ts
```diff
@@ -90,13 +96,19 @@ const useQuickstartsStates = () => {
...
   }, []);
 
-  return {
-    activateQuickstart,
-    allQuickStartStates,
-    setAllQuickStartStates,
-    activeQuickStartID,
-    setActiveQuickStartID,
-  };
+  // make sire the API is not created on every render
```
fhlavac, line 99:
```suggestion
  // make sure the API is not created on every render
```

fhlavac (approved):

PR #2955, opened 2024-10-09
Converting GlobalFilter to Jotai

justinorringer:
This will touch a lot of the code base, so I have some questions.


src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -54,18 +49,7 @@ const useLoadTags = (hasAccess = false) => {
...
-    ({ globalFilter: { tags, sid, workloads } }: ReduxState) => ({
-      count: (tags.count || 0) + (sid.count || 0) + (workloads.count || 0),
-      total: (tags.total || 0) + (sid.total || 0) + (workloads.total || 0),
-      tags: tags.items || [],
-      sid: sid.items || [],
-      workloads,
-    }),
-    shallowEqual
-  );
+  const { count, total, tags, sid, workloads, isLoaded } = useGlobalFilterState();
```
justinorringer, line 52: Created a hook for this, may be unnecessary?

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -1,51 +1,46 @@
...
-          })
-        );
-        dispatch(
-          fetchAllWorkloads({
-            registeredWith,
-            activeTags,
-            search,
-          })
-        );
+      // will removing the batch here cause problems? they are all fetches
```
justinorringer, line 28: Did we have race conditions here? Curious why these were batched together. In that discussion, somebody else locked the resources before fetching to imitate the batch. Should we do that?

src/@types/types.d.ts
```diff
@@ -190,6 +190,79 @@ export interface GroupItem {
 
 export type FlagTagsFilter = Record<string, Record<string, boolean | GroupItem>>;
 
+export type GlobalFilterWorkloads = {
```
justinorringer, line 193: Put these variables from the store in types, let me know with a better place. :sweat_smile:

src/redux/redux-config.ts
```diff
@@ -1,25 +0,0 @@
-import * as actionTemplates from './actions';
-import chromeReducer, { chromeInitialState } from '.';
```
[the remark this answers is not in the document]
Hyperkid123: There is the `on` event which leverages the middlewards. It can be and should done via jotai. https://github.com/Hyperkid123/insights-chrome/blob/master/src/chrome/create-chrome.ts#L100.

It was already done for the `APP_NAVIGATION` public event. You can follow similar approach.

src/state/reducers/index.ts
```diff
@@ -37,15 +32,3 @@ const globalFilter = {
   [GLOBAL_FILTER_REMOVE]: onGlobalFilterRemove,
   [GLOBAL_FILTER_UPDATE]: onTagSelect,
 };
-
-export const chromeInitialState: ReduxState = {
```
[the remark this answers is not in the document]
Hyperkid123: Move the types to the jotai atoms and reference it there. Most of the existing jotai atoms are just migrated redux state variables. You can check how its done there.

src/components/GlobalFilter/GlobalFilter.tsx
Hyperkid123: When consolidating state variables, just be careful you are properly memorizing the state output. To prevent extra re-renders.

I don't see an issue with using the jotai atoms directly.

src/components/GlobalFilter/GlobalFilter.tsx
Hyperkid123: You can create a write atom that will set the variables in a single callback. But I do not think removing the batch should be an issue now.

PR #2974, opened 2024-11-11
Update Search behavior/styling

epwinchell:
Hide search behind icon at all breakpoints
Jira: https://issues.redhat.com/browse/RHCLOUD-36133

New
Screenshot 2024-11-11 at 12 22 43 PM
Screenshot 2024-11-11 at 12 22 52 PM


Hyperkid123:
@epwinchell, I've added a commit that changes the Cypress runner to use Chrome, but it also messes a little bit with the configuration. It should fix the issues with different pixel ratios on mac retina display. This way we should not get false positives on the image snapshot reports.

Hyperkid123 (approved):

PR #2980, opened 2024-11-18
sentry: stop avoiding the image-builder app

croissanne:
Image builder now uploads sourcemaps and sends everything into a separate project as we set the DSN in the module metadata.


adonispuente (approved):
LGTM

PR #2987, opened 2025-01-03
fix ansible bundle trial redirects

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-26705

Adds back check that redirects users to specific pages if they are not entitled to the current route.

@himdel can you please verify the route rule? Or do we know who knows to which routes the redirect should apply?

### DIFF
--- a/src/hooks/useTrialRedirect.ts
+++ b/src/hooks/useTrialRedirect.ts
@@ -1,0 +1,42 @@
+import { useContext, useEffect } from 'react';
+import ChromeAuthContext from '../auth/ChromeAuthContext';
+import { matchPath, useLocation, useNavigate } from 'react-router-dom';
+import { ChromeUser } from '@redhat-cloud-services/types';
+import { isAnsibleTrialFlagActive } from '../utils/isAnsibleTrialFlagActive';
+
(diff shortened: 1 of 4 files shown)


src/hooks/useTrialRedirect.ts
```diff
@@ -0,0 +1,42 @@
+import { useContext, useEffect } from 'react';
+import ChromeAuthContext from '../auth/ChromeAuthContext';
+import { matchPath, useLocation, useNavigate } from 'react-router-dom';
+import { ChromeUser } from '@redhat-cloud-services/types';
+import { isAnsibleTrialFlagActive } from '../utils/isAnsibleTrialFlagActive';
+
+const ignoredPaths = ['/ansible/trial/*'];
```
himdel, line 7: Is the path correct? (and on line 16)

Shouldn't it be `/ansible/ansible-dashboard/trial/*`?

himdel:
> Or do we know who knows to which routes the redirect should apply?

Definitely `/ansible/ansible-dashboard/` (except the trial routes), `/ansible/automation-hub` and `/ansible/automation-analytics`. (Not sure about any other ansible routes, such as lightspeed.)

Hyperkid123:
OK I've updated the trial routes.

Hyperkid123:
OK. I've updated the routes.

PR #2990, opened 2025-01-10
Update PatternFly to v6

wise-king-sullyman:
This PR updates the repo to use PatternFly v6 as well as upstream PF consuming packages which have already been migrated.


Hyperkid123:
There are some dependency resolution issues with redux. We will need an override for now, until we remove redux from frontend components as a dependency:

```diff
diff --git a/package.json b/package.json
index d9945d8f..c8fcf234 100644
--- a/package.json
+++ b/package.json
@@ -192,5 +192,8 @@
   "nyc": {
     "report-dir": "cypress-coverage"
   },
+  "overrides": {
+    "react-redux": "^8.1.3"
+  },
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
 ```

File: config/webpack.config.js [hunk 1]
```diff
@@ -34,7 +37,7 @@ const commonConfig = ({ dev }) => {
         }
       : path.resolve(__dirname, '../src/index.ts'),
     output: {
-      path: path.resolve(__dirname, '../build/js'),
+      path: path.resolve(__dirname, '../build/stable/js'),
```
Hyperkid123, line 40: This should not be needed

config/webpack.config.js
```diff
@@ -146,7 +149,7 @@ const commonConfig = ({ dev }) => {
       // HMR flag
       hot: true,
       ...proxy({
-        env: 'stage-beta',
+        env: 'prod-stable',
```
Hyperkid123, line 152: ^^

config/webpack.config.js
```diff
@@ -190,9 +193,10 @@ const commonConfig = ({ dev }) => {
 const pfConfig = {
   entry: {
     'pf4-v5': path.resolve(__dirname, '../src/sass/pf-5-assets.scss'),
+    'pf-v6': path.resolve(__dirname, '../src/sass/pf-6-assets.scss'),
   },
   output: {
-    path: path.resolve(__dirname, '../build/js/pf'),
+    path: path.resolve(__dirname, '../build/stable/js/pf'),
```
Hyperkid123, line 199: This should not be required

src/chrome/create-chrome.ts
```diff
@@ -179,8 +179,9 @@ export const createChromeContext = ({
     clearAnsibleTrialFlag,
     isAnsibleTrialFlagActive,
     setAnsibleTrialFlag,
-    chromeHistory,
-    analytics: analytics!,
+    // FIXME: get rid of these anys
+    chromeHistory: chromeHistory as any,
```
Hyperkid123, line 183: These should no longer be required

src/components/AllServicesDropdown/AllServicesDropdown.scss
```diff
@@ -1,27 +1,31 @@
...
+  &.pf-v6-c-menu-toggle:hover::after,
+  &.pf-m-expanded::after {
+    border-bottom: var(--pf-t--global--border--width--extra-strong) solid var(--pf-t--global--border--color--status--info--default);
   }
 }
 
 .chr-c-page__services-nav-dropdown-menu {
-  @media screen and (min-width: $pf-v5-global--breakpoint--2xl) {
-    top: 70px;
-    height: 100vh;
```
Hyperkid123, line 16: We should take note of this as there were some specific reasons why this property exists. Or is this something that is replaced internal by the page layout in PF now?

Hyperkid123:
The `ChromeUser` typing issues is resolved in `@redhat-cloud-services/types@2.0.0-beta.1`

We should also look at the webpack module sharing config as it seems that for example the learning resources are picking up pf6 components rather that pulling their own version.

config/webpack.config.js
wise-king-sullyman: I'm happy to change this and the other stable addition back, but for context I added these because I otherwise had to manually move the build output into a `/stable` directory before I was able to build an image and get it into an ephemeral env. cc @florkbr

config/webpack.config.js
wise-king-sullyman: I don't have a staging account, do you mind if I keep this here until this is actually ready to go in so that I don't have to change it back every time I want to run this locally?

src/chrome/create-chrome.ts
wise-king-sullyman: I updated the `@redhat-cloud-services/types` version to `2.0.0-beta.1`, but with these `any` castings removed I'm still seeing an error at build time: 
image

src/components/AllServicesDropdown/AllServicesDropdown.scss
wise-king-sullyman: I believe that this was removed based on feedback from design. I want to say an issue with nested scrollable areas & double scroll bars on some of the services pages.

What was the reason the 100vh height was needed originally? I might be able to accomplish that original goal and not introduce the nested scrolling issue with a slightly different approach.

Hyperkid123: I don't remember TBH. But I want to keep this in mind. In case we get some reports.

src/chrome/create-chrome.ts
Hyperkid123: OK. We will can keep it then,

Hyperkid123:
@wise-king-sullyman the cypress component tests seem to be stuck at the build step. They have their own webpack configuration. Could you take a look?

florkbr:
Looking into PR failures - it seems like Konflux is complaining about missing tasks in our tekton config. @wise-king-sullyman can you update this branch with main? At least for the `.tekton` directory files.

Hyperkid123:
@florkbr @wise-king-sullyman we will be merging this soon. Just wanted to remind, you that we should squash the commits in case we need a revert.

florkbr (approved):

PR #2993, opened 2025-01-20
Use FEO generated services tile if flag is present

Hyperkid123:
PR: https://issues.redhat.com/browse/RHCLOUD-35431

For now still uses a hard coded local storage flag. THis flag will exist until we open migration to the platform. Then it will be migrated to unleash to separate prod/stage and other environments.


florkbr (approved):
Code looks good - @Hyperkid123 depending on how long this code will stay in (during the platform migration) we may want to consider adding a UI test when we flip this localstorage flag. CC @catastrophe-brandon @jjaquish

We may also want to indicate somewhere that `GENERATED_SEARCH_FLAG` flag also controls the all services tiles.

Hyperkid123:
@florkbr so this should stay until we open the migration. Then EVERYTHING should be the same on the UI side. So existing tests should not break. I expect that they might, but they should not really be affected. Right now we have just the staging bundle and nothing else, si it would definitely break. Once I am done with the RBAC onboarding, I'll turn on all the bundles and we should have one to one configurations.

PR #3003, opened 2025-01-30
Fix error with caddy config

florkbr:
Fixes:
Error: adapting config using caddyfile: /opt/app-root/src/Caddyfile:4: unrecognized global option: header


ryelo (approved):

florkbr:
Will let merge pipeline run.

PR #3007, opened 2025-02-05
Allow module preloading based on routing information.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-37708


src/chrome/update-shared-scope.ts
```diff
@@ -42,7 +61,15 @@ const updateSharedScope = () => {
         Link: (props: LinkProps) => {
           const Cmp = reactRouter.Link;
           const react = require('react');
-          return react.createElement(Cmp, { ...props, to: calculateTo(props.to) });
+          return react.createElement(Cmp, {
+            ...props,
+            // monkey patch the mouse enter event to preload module if it exists
+            onMouseEnter: (e: any) => {
```
florkbr, line 67: @Hyperkid123 mostly a NIT - but thoughts on adding a small requirement that the user "pauses" over the link for 250 milliseconds (or similar)? I'm just trying to avoid the case where the cursor was dragged "over" the nav item to get to something else - but the nav item was not the "target" if that makes sense.

Hyperkid123: @florkbr way ahead of you: https://github.com/RedHatInsights/insights-chrome/pull/3007/files#diff-ef54dba243af0302f518aae8bfd7165ee24781fe4ccc450de1cbd970568128abR16-R18

I was thinking about cache. But I see we could add the timeout here to not search the routing and matching app. But the loading happes after some time. Let me know what you think.

Hyperkid123: @florkbr I made it more efficient. What do you think now?

Hyperkid123: I think clearing the timeout and canceling the preload is not necessary. I argue that it is even unwanted. 

A preload happens only once for a module. Then the bundle is in the browser and in the module cache. We are also preventing "parallel" preloading for a single module. That is all handled in Scalprum. 

I am not worried about "brushing" over links. Most of the links are not "close" in the UI. And when they are, they lead to the active or same module. That does not trigger pre-loads (I know the landing page is an exception). So in fact we want the module to be pre-loaded even if the final action is not a click on the first "brushed" link because the final action will likely happen on a link for the same module.

This would be a case of premature optimization IMO. The place where this can be an issue is the left main nav. We have a lot of "close links" that lead to different modules. But, this preload has been baked into all Chrome links (including nav) for some years. We never observed any network problems or excessive preloading with it. Even there the rule that links close together usually point to the same module applies.

In the past, I was considering an even more "drastic" approach, that other React frameworks have. And that would be pre-loading modules if a Link is rendered. But for that, I think our entry points need to be even smaller.

If we also take into account browser asset caching, and JS module sharing I don't think we should be worried about some extra network traffic.

florkbr: @Hyperkid123 ack. My main concern is low-bandwidth performance impacts (IE roundtrip overseas, VPN, mobile network, etc). I know that is more of an edge case, but I think it is worth considering.

I agree if we were to preload modules every time a link is rendered - we'd want to make sure our bundles are as small as possible.

I'm fine proceeding with this PR as-is and we can keep an ear out for any performance issues separately as we already have preloading occurring today.

ryelo (approved):
Approving PR based on Bryan's comment: https://github.com/RedHatInsights/insights-chrome/pull/3007#discussion_r1970650570

PR #3011, opened 2025-02-10
Fix bug preventing permission error message from appearing

JetyAdam:
GatewayErrorComponent wasn't rendered in some edge cases.

https://issues.redhat.com/browse/RHCLOUD-37729


src/bootstrap.tsx
```diff
@@ -36,11 +37,14 @@ const useInitializeAnalytics = () => {
...
-  const loaded = useSessionConfig();
+  const { gatewayError, configLoaded } = useSessionConfig();
 
   useInitializeAnalytics();
 
-  if (!loaded) {
+  if (!configLoaded && gatewayError) {
+    return <GatewayErrorComponent error={gatewayError} serviceName="Hybrid Cloud Console" />;
+  }
+  if (!configLoaded) {
```
florkbr, line 47: @JetyAdam this seems a bit untidy to me. We could have:

```tsx
  if (!configLoaded) {
    return gatewayError ? <GatewayErrorComponent error={gatewayError} serviceName="Hybrid Cloud Console" /> : <AppPlaceholder />;
  }
```

florkbr: @JetyAdam this is also missing tests. I think this is a perfect case to have test automation for as it impacts so many users - but is very rare for devs to experience regularly. 

Can we open a follow up ticket to clean this up?

catastrophe-brandon: Just spitballing here - would it be possible to set up some sort of a unit test with a mocked config/response? Essentially mock up a scenario with a fake user that causes a gateway error, then we confirm that the code surfaces the appropriate error? 

If not, we might be able to do this in e2e tests, but it will be prone to breakage.

catastrophe-brandon: I may not be using the right vernacular - if it's not a unit test it could be a component test. Something that doesn't require a fully-built end-to-end environment to execute.

JetyAdam: @florkbr You are right. I updated the if statement.

JetyAdam: @catastrophe-brandon I think that it would be possible to set up a unit test.

florkbr:
Since this is effecting end users and we have our PF6 release tomorrow - I'm going to merge. We can follow up on the test plan ASAP.

florkbr (approved):

src/bootstrap.tsx
Hyperkid123: @catastrophe-brandon @JetyAdam there already is a cypress component test like that. We should just expand the cases.

https://github.com/RedHatInsights/insights-chrome/blob/master/cypress/component/GatewayErrors.cy.tsx

PR #3012, opened 2025-02-11
Add global learning resources config

JetyAdam:


src/locales/Messages.ts
```diff
@@ -330,7 +330,7 @@ export default defineMessages({
   globalLearningResourcesPage: {
     id: 'globalLearningResourcesPage',
     description: 'All learning resources',
-    defaultMessage: 'Global learning resources page',
+    defaultMessage: 'All learning resources',
```
florkbr, line 333: @JetyAdam in order to have this show up I think we need to run `npm run translations` locally so https://github.com/RedHatInsights/insights-chrome/blob/master/src/locales/data.json#L37 will get updated.

JetyAdam: Updated

florkbr (approved):

florkbr:
FWIW the jenkins PR build is failing to deploy chrome frontend via bonfire. I will manually monitor stage to verify it rolls out as expected.

PR #3018, opened 2025-02-20
Check token validity on network and browser window status changes.

Hyperkid123:
Jira: https://issues.redhat.com/browse/RHCLOUD-38189


src/auth/OIDCConnector/OIDCSecured.tsx
```diff
@@ -202,6 +202,46 @@ export function OIDCSecured({
...
+      if (visibilityState !== document.visibilityState && document.visibilityState === 'visible') {
+        // we went from non visible to visible (tab went into focus for various reasons)
+        loginOnTokenExpired();
+      }
+      visibilityState = document.visibilityState;
+    }
+
+    function networkOnlineListener() {
+      loginOnTokenExpired();
+      authRef.current.startSilentRenew();
```
florkbr, line 227: @Hyperkid123 could we add some basic tests to confirm these actions are performed when the event listeners are fired? This is one of those areas we could regress without noticing for awhile.

Otherwise approach looks good to me.

florkbr: I'll also likely need to test this in our itless env

Hyperkid123: This will be very hard to test without a custom KC instance 😢 or we would have to wait for token expiration, turn off then network etc.

I can do basic testing that the pause/resume functionality works, but timing out the tokens might be difficult. I think I should be able to mock it, but this would need a proper e2e testing.

catastrophe-brandon: Would it be possible to have a stubbed version of KC that has functionality for deliberately expiring tokens for the sake of testing? Seems like that's something we could use in future tests, so there would be future ROI on the initial work.

Hyperkid123: We could have a separate e2e test and run the KC instance locally and decrease the timeouts, yes. But we do not have an environment for that right now. Once we have some e2e available in Konflux we can do that.

florkbr: @Hyperkid123 getting back to this PR - I get we will need to wait for E2E - but would it be possible to move these function definitions outside of the `useEffect` and test them with jest tests? I'd like to make sure all the expected functions are called and we consider any undefined/falsy objects.

Hyperkid123: Sure. But I think we should use cypress instead. JSDOM events are most likely just fake.

Hyperkid123: @florkbr added the tests.

florkbr (approved):

PR #3024, opened 2025-02-25
Fix css selector for global filter label group.

Hyperkid123:
Follow up to: https://issues.redhat.com/browse/RHCLOUD-37879

### DIFF
--- a/src/components/GlobalFilter/global-filter-menu.scss
+++ b/src/components/GlobalFilter/global-filter-menu.scss
@@ -1,7 +1,7 @@
 .chr-c-global-filter {
   padding: var(--pf-t--global--spacer--md) var(--pf-t--global--spacer--lg);
   background: var(--pf-t--global--background--color--100);
-  .pf-v6-c-chip-group { margin-right: var(--pf-t--global--spacer--xs); }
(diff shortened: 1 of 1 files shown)


src/components/GlobalFilter/global-filter-menu.scss
```diff
@@ -1,7 +1,7 @@
 .chr-c-global-filter {
   padding: var(--pf-t--global--spacer--md) var(--pf-t--global--spacer--lg);
   background: var(--pf-t--global--background--color--100);
-  .pf-v6-c-chip-group { margin-right: var(--pf-t--global--spacer--xs); }
+  .pf-v6-c-label-group { margin-right: var(--pf-t--global--spacer--xs); }
```
florkbr, line 4: @Hyperkid123 I wonder if the codemods should have picked this up? Or were we using the wrong class prior to PF6?

Change looks fine to me otherwise.

florkbr (approved):
Tested locally and this fixes the spacing between the filters/tags

src/components/GlobalFilter/global-filter-menu.scss
Hyperkid123: No clue why this was missed.

PR #3026, opened 2025-02-26
feat(sentry): Allow other Red Hat Services to be incapsulated by sentry

adonispuente:
This is more of a proof of concept that'd allow testing for the onboarding of other apps wrapped by chrome - to be incorporated by sentry. Once merged openshifts apps will get routed too (2 apps total, still interact-team managed apps)

After this the goal would be to migrate the majority of this set up to Frontend Components


adonispuente:
Bump

src/utils/sentry.ts
```diff
@@ -7,22 +7,27 @@ let sentryInitialized = false;
...
-  let appGroup;
+  let appGroup = 'insights';
   let appName;
-  let betaCheck;
-
+  let environment = 'stable';
   if (pathName[1] === 'beta') {
-    betaCheck = ' Beta';
     appGroup = pathName[2];
-    appName = appGroup === 'landing' ? 'landing' : pathName[3];
```
justinorringer, line 17: This line was weird, but looks like ya lost that landing condition

justinorringer (commented):
Seems fine to me, one quick thing

src/utils/sentry.ts
florkbr: @justinorringer I believe this moved to the else condition down below: https://github.com/RedHatInsights/insights-chrome/pull/3026/files#diff-7cf585363d10fb2f338e86818827620adf01c167e93e8186a776cddf81b8c380R23-R25

src/utils/sentry.ts
```diff
@@ -7,22 +7,27 @@ let sentryInitialized = false;
...
+    appName = pathName[3];
+    environment = 'beta';
   } else {
-    betaCheck = '';
     appGroup = pathName[1];
-    appName = appGroup === 'landing' ? 'landing' : pathName[2];
+    if (appGroup === 'openshift') {
```
florkbr, line 19: @adonispuente we could add a test to exercise these conditionals/functions as we have regressed this without realizing in the past (for example the `beta` logic not working any longer).

florkbr (changes requested):
Overall looks good - just a few comments we should discuss before merging. Tested locally and I see the correct app details being populated when navigating to the various apps/groups.

src/utils/sentry.ts
```diff
@@ -7,22 +7,27 @@ let sentryInitialized = false;
...
+      appName = pathName[3];
+    } else {
+      appName = pathName[2];
+      if (appGroup === 'landing') {
```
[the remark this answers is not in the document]
adonispuente: Nah its all good, I'll add this in the next commit too!

src/utils/sentry.ts
```diff
@@ -7,22 +7,27 @@ let sentryInitialized = false;
 
 function getAppDetails() {
   const pathName = window.location.pathname.split('/');
-  let appGroup;
+  let appGroup = 'insights';
   let appName;
-  let betaCheck;
-
+  let environment = 'stable';
   if (pathName[1] === 'beta') {
```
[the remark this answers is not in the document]
adonispuente: I think this logic can go straight to the trash if we've completely removed beta , removing

src/utils/sentry.ts
adonispuente: This will get removed now but yeah it was just lower

src/utils/sentry.ts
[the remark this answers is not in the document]
adonispuente: Look at this now, im not sure what we'd want the appName to resolve to in sentry. Essentially theres no 'app' here. Theres no project for that page in sentry sp it would end up in cpin-001-insights but im not sure what we'd want the make the app name be if we want errors from that page to have a specific name

src/utils/sentry.test.js
```diff
@@ -0,0 +1,53 @@
+import { getAppDetails } from './sentry';
+
+describe('getAppDetails function', () => {
+  beforeEach(() => {
+    Object.defineProperty(global, 'location', {
```
florkbr, line 5: @adonispuente Jest does not reset global variables automatically after a test has executed, so with this change we run the risk of contaminating tests that run after this spec.

There are a few ways to avoid global variables in the spec. In lieu of accessing `window.location` in the source code for `sentry.ts` you can add a helper function that simply returns the value of window.location, and then spy on that function in your test.

An alternative would be to clean up after this test in an afterEach.

adonispuente: I sent up the change in the latest commit!

florkbr (approved):

PR #3038, opened 2025-03-06
FEC - Bump PF6 packages to latest stable versions

florkbr:
This updates chrome's dependencies to pull in the stable releases of our FEC PF6 enabled packages.


florkbr:
Quick sanity check locally and everything appears fine: 
image

PR #3040, opened 2025-03-11
Enable scope updates with iframe flow.

Hyperkid123:
### Description

An experiment that enables session scope expansion without a full browser refresh. The method uses the IFrame OIDC flow.

### Changes
Since the iframe flow was not used, we had some misconfiguration of the OIDC client which surfaced once the flow was "forced".
- removed the different output path of the `silent-check-sso.html` between dev and prod build
- fix the actual `silent-check-sso.html` post message payload (it was still using the keycloack-js signature)
- `reAuthWithScopes` will try the iframe flow and fallbacks to `signInRedirect` if it fails
- `useUserSSOScopes` is now using `reAuthWithScopes` instead of the `auth.login` directly

### Caveats
The OIDC client we use will **ignore** the iframe flow if a refresh_token was provided. This is ok unless we want the scopes to be updated. There is no supported way of disabling the refresh_token refresh unless the refresh_token is disabled directly in the KC settings. It's not something we want.

However, we can patch the condition in the node_modules  and disable it via global variable (simplest possible change). We can leverage https://github.com/ds300/patch-package to do so.  Ideally, we would try to work with the client maintainers to find a proper solution. But if we decide to accept this change, we at least can do so before we manage to update the client code.


florkbr:
@Hyperkid123 have we heard back from the maintainer of this package on contributing an enhancement?

Hyperkid123:
@florkbr not yet.

PR #3042, opened 2025-03-13
Do not require user to show all services dropdown.

Hyperkid123:
Just a small change, the user data is not necessary to render the new all services dropdown.


florkbr:
Tested locally against stage and everything still works as expected.

florkbr (approved):

PR #3044, opened 2025-03-17
Add additional token/session error checks.

Hyperkid123:
Part of: https://issues.redhat.com/browse/RHCLOUD-37897

Add a new error boundary that should explicitly deal with thrown errors form the OICD user manager. We have no callbacks API in the lib that could handle these exception so extra boundary is needed.

If listed events occur, we force re-authenticate the user. If we thee error does not match, we bubble the error up to the generic boundary.

### DIFF
--- a/cypress/component/OIDCConnector/OIDCUserManagerErrorBoundary.cy.tsx
+++ b/cypress/component/OIDCConnector/OIDCUserManagerErrorBoundary.cy.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import OIDCUserManagerErrorBoundary, { SESSION_NOT_ACTIVE, TOKEN_NOT_ACTIVE } from '../../../src/auth/OIDCConnector/OIDCUserManagerErrorBoundary';
+import { UserManager } from 'oidc-client-ts';
+import ErrorBoundary from '../../../src/components/ErrorComponents/ErrorBoundary';
+import { IntlProvider } from 'react-intl';
+
+const ThrowAbleComponent = ({ error }: { error: any }) => {
+  throw error;
+};
+
+describe('OIDCUserManagerErrorBoundary', () => {
+  let basicFakeManager: UserManager;
+
+  beforeEach(() => {
+    basicFakeManager = new UserManager({
+      authority: '',
+      client_id: '',
+      redirect_uri: '',
+    });
+  });
+
+  it('should render children if no error is thrown', () => {
+    cy.mount(
+      <OIDCUserManagerErrorBoundary userManager={basicFakeManager}>
+        <h1>Child</h1>
+      </OIDCUserManagerErrorBoundary>
+    );
+
+    cy.get('h1').should('exist');
+  });
+
+  it('Should bubble unrelated OIDC timeout error to parent error boundary', () => {
+    cy.on('uncaught:exception', () => {
+      return false;
+    });
+
+    cy.mount(
+      <IntlProvider locale="en">
+        <ErrorBoundary>
+          <OIDCUserManagerErrorBoundary userManager={basicFakeManager}>
+            <ThrowAbleComponent error="Fake error" />
+          </OIDCUserManagerErrorBoundary>
+        </ErrorBoundary>
+      </IntlProvider>
+    );
+
+    cy.contains('Fake error').should('exist');
+  });
+
+  [SESSION_NOT_ACTIVE, TOKEN_NOT_ACTIVE].forEach((error) => {
+    it('should try redirect to signin page if error is thrown', () => {
+      cy.intercept('GET', '/authorityUrl', {
+        statusCode: 200,
+        body: {
+          success: true,
+          error,
+        },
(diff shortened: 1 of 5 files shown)


src/auth/OIDCConnector/OIDCUserManagerErrorBoundary.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const SESSION_NOT_ACTIVE = 'Session not active';
+export const TOKEN_NOT_ACTIVE = 'Token not active';
+
+function isInvalidAuthStateError(error: any): boolean {
+  return error?.error_description === SESSION_NOT_ACTIVE || error?.error_description === TOKEN_NOT_ACTIVE;
+}
+/**
+ *  The session not active and token not active are not handled by the oidc-client library and will "bubble up".
+ *  We need a specific error boundary to handle these error. There is no way of preventing the errors from being thrown. And they will always show up in Sentry.
```
florkbr, line 20: 👍 for the quick explanation

cypress/component/OIDCConnector/OIDCUserManagerErrorBoundary.cy.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <OIDCUserManagerErrorBoundary userManager={fakeManager}>
+        <h1>Child</h1>
+      </OIDCUserManagerErrorBoundary>
+    );
+
+    cy.get('h1').should('exist');
+  });
+
+  it('Should bubble unrelated OIDC timeout error to parent error boundary', () => {
+    const fakeManager = new UserManager({
```
florkbr, line 28: @Hyperkid123 some of this test setup is duplicated between our specs. Can we move it into a `beforeEach`?

florkbr:
FWIW this will cause merge conflicts with https://github.com/RedHatInsights/insights-chrome/pull/3040, however everything is working locally for me. Just one small comment.

Hyperkid123:
@florkbr can you take another look?

cypress.config.ts
```diff
@@ -49,10 +49,10 @@ export default defineConfig({
   },
   e2e: {
     blockHosts: ['consent.trustarc.com'],
-    baseUrl: 'https://stage.foo.redhat.com:1337/',
+    baseUrl: 'https://ee-vfwknanu.apps.crc-eph.r9lp.p1.openshiftapps.com/',
```
josejulio, line 52: Is this change on purpose or was it part of a local test?

cypress.config.ts
```diff
@@ -49,10 +49,10 @@ export default defineConfig({
...
     env: {
-      E2E_USER: process.env.E2E_USER,
-      E2E_PASSWORD: process.env.E2E_PASSWORD,
+      E2E_USER: 'jdoe', //process.env.E2E_USER,
+      E2E_PASSWORD: 'jyJSIx6vo9BlW3b5', // process.env.E2E_PASSWORD,
```
josejulio, line 55: Is this change on purpose or was it part of a local test?

Hyperkid123: Oh crap! Thanks for spitting these.

cypress/component/OIDCConnector/OIDCUserManagerErrorBoundary.cy.tsx
catastrophe-brandon, line 1: Thank you for adding tests!

cypress/e2e/ephemeral/ee.cy.ts
```diff
@@ -0,0 +1,10 @@
+describe('Ephemeral test', () => {
+  beforeEach(() => {
+    // cy.login();
```
catastrophe-brandon, line 3: Is there any reason not to remove the entire `beforeEach`?

cypress/e2e/ephemeral/ee.cy.ts
```diff
@@ -0,0 +1,10 @@
...
+  });
+  it('should find all services page', () => {
+    cy.visit('/allservices');
+    cy.contains('All Services').should('exist');
+    cy.get('.pf-v6-c-masthead__brand > div > .pf-v6-c-menu-toggle > .pf-v6-c-menu-toggle__text').click();
```
catastrophe-brandon, line 8: I'm assuming this works, but is there a way we can avoid using selectors? Not a blocker for this PR, but if an OUIA ID can avoid using selectors, I'd favor that.

cypress/e2e/ephemeral/ee.cy.ts
Hyperkid123: Sorry this was not supposed to be here, I was doing some POC with Matt about Konflux PR checks. I'll remove this file.

cypress/e2e/ephemeral/ee.cy.ts
Hyperkid123: Same here, this file is not supposed to be here.

PR #3060, opened 2025-04-08
fix: :bug: remove "Welcome to Preview" popover text - RHCLOUD-36415

ryelo:
https://issues.redhat.com/browse/RHCLOUD-36415

This just removes the popover component


ryelo:
Checked with @catastrophe-brandon, we don't think there's any tests in IQE for this, so it should be fine.

Hyperkid123 (approved):

PR #3075, opened 2025-04-28
Try signing silent on expired token

Hyperkid123:


justinorringer (approved):
LGTM! (naively possibly)

PR #3077, opened 2025-04-28
update segment modules to include insights

CodyWMitchell:
[RHCLOUD-37946](https://issues.redhat.com/browse/RHCLOUD-37946)
- Update Segment configuration to include Insights

## Summary by Sourcery

New Features:
- Include the 'insights' module in the Segment analytics provider.


florkbr:
We will be adding these values to the chrome service backend static fed-modules config (eventually migrated to each individual apps frontend.yaml FEO config).

PR #3089, opened 2025-05-14
fix(all-services): permissions on new redesign to be used

karelhala:
### Description

As a user if there are any permissions applied to any navigation this link is visible on the all services redesign. This PR fixes such issue by checking the permissions and filtering out those links which should not be visible for current user.

## Summary by Sourcery

Filter out navigation links in the All Services redesign based on user permissions so that unauthorized links are hidden.

Bug Fixes:
- Hide navigation links for which the current user lacks permissions in the All Services redesign.

Enhancements:
- Introduce a recursive filterItem utility that marks and removes hidden navItems and routes based on evaluateVisibility.
- Update fetchBundles to apply permission-based filtering to all bundle navigations before rendering.


justinorringer (approved):
LGTM!

PR #3093, opened 2025-05-16
feat: add workspace label to settings menu

radekkaluzik:
feat: add workspace label to settings menu

https://issues.redhat.com/browse/RHCLOUD-34249

A slight visual discrepancy compared to the mockups is due to a different implementation in PatternFly 6.

workspace label


karelhala:
@radekkaluzik looks like it's failing on some lint error.

src/components/Header/SettingsToggle.tsx
```diff
@@ -54,6 +55,7 @@ const SettingsToggle = (props: SettingsToggleProps) => {
                 : ({ className: itemClassName }) => (
                     <ChromeLink {...rest} className={itemClassName} href={url} rel={rel} isBeta={isPreview}>
                       {title}
+                      {titleLabel}
```
Hyperkid123, line 58: I believe we should add this to the dropdown item description prop: https://www.patternfly.org/components/menus/dropdown#with-item-descriptions

I know we use custom dropdown component definition, but I'd like to see if we can use the `description` prop first on the dropdown item.

Hyperkid123: If we pass the title as a child to the dropdown item itself, we should be able to just pass the props to the ChromeLink (it should get the children prop).

But if the looks off we will have to stick with this.

radekkaluzik: Done

src/components/Header/Tools.tsx
```diff
@@ -25,6 +25,8 @@ import { notificationDrawerExpandedAtom } from '../../state/atoms/notificationDr
 import useSupportCaseData from '../../hooks/useSupportCaseData';
 import { ScalprumComponent, ScalprumComponentProps } from '@scalprum/react-core';
 import { drawerPanelContentAtom } from '../../state/atoms/drawerPanelContentAtom';
+import { Label } from '@patternfly/react-core/dist/dynamic/components/Label';
+import UsersIcon from '@patternfly/react-icons/dist/esm/icons/users-icon';
```
Hyperkid123, line 29: This should be `/dist/dynamic` as well.

Hyperkid123:
One small import change then we can merge.

src/components/Header/Tools.tsx
radekkaluzik: Fixed

PR #3095, opened 2025-05-16
Jotai migration - globalFilter

JetyAdam:


src/redux/actions.test.js
catastrophe-brandon, line 1: Was this moved somewhere else?

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -1,123 +1,176 @@
...
-    debounce((activeTags: any, search: any) => {
-      storeFilter(activeTags, hasAccess && !isDisabled, navigate);
-      batch(() => {
-        dispatch(
-          fetchAllTags({
+    debounce(async (activeTags: any, search: any) => {
+      // Set loading state to false before fetching
+      setTags((prev) => ({ ...prev, isLoaded: false }));
+      setSids((prev) => ({ ...prev, isLoaded: false }));
+      setWorkloads((prev) => ({ ...prev, isLoaded: false }));
```
Hyperkid123, line 39: When we do multiple "set calls", each of them means one render. I'd suggest creating a write atom that would set the loading with the same value (in this case false) state to all of these individual atoms.

It can be re-used down bellow.

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -1,123 +1,176 @@
...
-      });
+          }),
+        ]);
+      } catch (error) {
+        console.error('Failed to load global filter tags:', error);
+      } finally {
+        // Set loaded property to true even if an error occurs
+        setTags((prev) => ({ ...prev, isLoaded: true }));
+        setSids((prev) => ({ ...prev, isLoaded: true }));
+        setWorkloads((prev) => ({ ...prev, isLoaded: true }));
```
Hyperkid123, line 65: ^^ batch write aotm would be great

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -1,123 +1,176 @@
...
-      tags: tags.items || [],
-      sid: sid.items || [],
-      workloads,
-    }),
-    shallowEqual
-  );
+  const isLoaded = useAtomValue(isLoadedAtom);
+  const tagsData = useAtomValue(tagsAtom);
+  const sidsData = useAtomValue(sidsAtom);
+  const workloadsData = useAtomValue(workloadsAtom);
```
Hyperkid123, line 77: Here I would create a common "read atom" that would give all of these. WE don't have to retrieve them individually, and the re-render count will decrease.

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -1,123 +1,176 @@
...
-      chips: GlobalFilterDropdownProps['chips'];
-      selectedTags: FlagTagsFilter;
-      setValue: GlobalFilterDropdownProps['setValue'];
-      filterTagsBy: string;
-    }
-  )(
-    [workloads, ...sid, ...tags],
+  const setTags = useSetAtom(tagsAtom);
+  const setSids = useSetAtom(sidsAtom);
+  const setWorkloads = useSetAtom(workloadsAtom);
```
Hyperkid123, line 82: Another bacthw rite atom would improve the performance and quality of the code.

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -1,123 +1,176 @@
...
+
+  const count = (tagsData.count || 0) + (sidsData.count || 0) + (workloadsData.count || 0);
+  const total = (tagsData.total || 0) + (sidsData.total || 0) + (workloadsData.total || 0);
+
+  const filterData: AllTag[] = [
```
Hyperkid123, line 87: Creating this array seems quite heavy. Usememo would be good here.

src/components/GlobalFilter/GlobalFilter.tsx
```diff
@@ -129,17 +182,16 @@ const GlobalFilter = ({ hasAccess }: { hasAccess: boolean }) => {
...
   const { getUserPermissions } = useContext(InternalChromeContext);
 
   // FIXME: Clean up the global filter display flag
   const isLanding = pathname === '/';
   const isAllowed = isGlobalFilterAllowed();
-  const globalFilterHidden = useSelector(({ globalFilter: { globalFilterHidden } }: ReduxState) => globalFilterHidden);
-  const activeModule = useAtomValue(activeModuleAtom);
-  const isDisabled = globalFilterHidden || !activeModule;
+
+  const isDisabled = useAtomValue(isDisabledAtom);
```
Hyperkid123, line 194: The name of this atom `isDisabledAtom` is very generic. It can be interpreted that any element can be disabled with this. Can you rename it to match its purpose better please?

src/components/GlobalFilter/GlobalFilterMenu.tsx
```diff
@@ -89,10 +88,9 @@ export const GlobalFilterDropdown: React.FunctionComponent<GlobalFilterDropdownP
    * We are unable to test it in any local development environment
    * */
   const hotjarEventEmitter = typeof window.hj === 'function' ? window.hj : () => undefined;
-  const registeredWith = useSelector(({ globalFilter: { scope } }: ReduxState) => scope);
+  const registeredWith = useAtomValue(registeredWithAtom);
```
Hyperkid123, line 91: Again the atom name seems to be a bit too generic

src/components/GlobalFilter/TagsModal.tsx
```diff
@@ -23,13 +30,16 @@ export type TagsModalProps = {
...
-    const selected = globalFilter[key];
-    return [selected?.isLoaded, selected?.total || 0, selected?.page || 1, selected?.perPage || 10];
-  }, shallowEqual);
+export const useMetaSelector = (key: 'tags' | 'workloads' | 'sid') => {
+  const tags = useAtomValue(tagsAtom);
+  const workloads = useAtomValue(workloadsAtom);
+  const sids = useAtomValue(sidsAtom);
+  const selected = { tags, workloads, sid: sids }[key];
+
+  return [selected?.isLoaded ?? false, selected?.total ?? 0, selected?.page ?? 1, selected?.perPage ?? 10] as [boolean, number, number, number];
```
Hyperkid123, line 41: Instead of typecasting the return value, you can define the return value of the function. Its a bit cleaner:

```tsx
export const useMetaSelector = (key: 'tags' | 'workloads' | 'sid'): [boolean, number, number, number]
```

This will ensure that TS will not complain down bellow. and you wont need type casting 🙂

src/components/GlobalFilter/TagsModal.tsx
```diff
@@ -100,14 +107,15 @@ const TagsModal = ({
...
+  const sidsData = useAtomValue(sidsAtom);
   const [tagsLoaded, tagsCount, tagsPage, tagsPerPage] = useMetaSelector('tags');
   const [sidLoaded, sidCount, sidPage, sidPerPage] = useMetaSelector('sid');
-  const tags = useSelector<ReduxState, GlobalFilterTag[]>(({ globalFilter: { tags } }) => tags?.items || []);
-  const sids = useSelector<ReduxState, SID[]>(({ globalFilter: { sid } }) => sid?.items || []);
-  const filterScope = useSelector(({ globalFilter: { scope } }: ReduxState) => scope || undefined);
-  const debounceGetTags = useDebounce(fetchAllTags, tagsPerPage, selectedTags);
-  const debounceGetSIDs = useDebounce(fetchAllSIDs, sidPerPage, selectedTags);
+  const tags = tagsData.items;
+  const sids = sidsData.items;
```
Hyperkid123, line 115: The new variables are not necessary, you can rename the `items` attribute while "deconstructing" the value of the atom:

```tsx

  const { items: tags } = useAtomValue(tagsAtom);
  const { items: sids } = useAtomValue(sidsAtom);
```

src/components/GlobalFilter/filterApi.ts
```diff
@@ -6,22 +6,32 @@ export const storeFilter = (tags: FlagTagsFilter, isEnabled: boolean, navigate:
...
+
+    if (SIDs && SIDs.length > 0) {
+      searchParams.append('SIDs', SIDs.join(','));
+    }
+
+    if (mappedTags && mappedTags.length > 0) {
+      searchParams.append('tags', mappedTags.join(','));
+    }
+
+    const finalUrl = {
```
Hyperkid123, line 25: Just a nit, this is not a URL but a so-called `to` argument or prop in react router.

src/components/GlobalFilter/globalFilterApi.test.js
```diff
@@ -33,8 +33,8 @@ describe('flatTags', () => {
 
   it('should create flat array of global filter', () => {
     const data = flatTags(globalFilter);
-    expect(data.length).toBe(2);
-    expect(data).toMatchObject(['someTag%2Fslash/someKey=[someValue%3Dvalue]', 'Workloads/SAP']);
+    expect(data.length).toBe(3);
+    expect(data).toEqual(expect.arrayContaining(['someTag%2Fslash/someKey=[someValue%3Dvalue]', 'Workloads/SAP', 'SAP ID (SID)/SOMEVAL']));
```
Hyperkid123, line 37: Why did these change? Ideally, the output should be exactly the same after refactoring.

I just want to make sure we are not adapting the tests to a "buggy" code.

src/components/GlobalFilter/globalFilterApi.test.js
```diff
@@ -45,16 +45,18 @@ describe('flatTags', () => {
 
   it('with encode enabled', () => {
     const data = flatTags(globalFilter, true);
-    expect(data.length).toBe(2);
-    expect(data).toMatchObject(['someTag%252Fslash/someKey=%5BsomeValue%253Dvalue%5D', 'Workloads/SAP']);
+    expect(data.length).toBe(3);
+    expect(data).toEqual(
+      expect.arrayContaining(['someTag%252Fslash/someKey=%5BsomeValue%253Dvalue%5D', 'Workloads/SAP', 'SAP%20ID%20(SID)/SOMEVAL'])
+    );
```
Hyperkid123, line 51: ^^

src/components/GlobalFilter/globalFilterApi.ts
```diff
@@ -5,7 +5,7 @@ import memoize from 'lodash/memoize';
 import { FlagTagsFilter, GroupItem } from '../../@types/types';
 import { getUrl } from '../../hooks/useBundle';
 
-export const SID_KEY = 'SAP ID (SID)';
+export const SID_KEY = 'SAP IDs (SID)';
```
Hyperkid123, line 8: We should not be changing the output. Especially if the URL is in question. We will break bookmarks

src/components/GlobalFilter/globalFilterApi.ts
```diff
@@ -137,15 +137,10 @@ export const flatTags = memoize(
...
-      ? [
-          Workloads,
-          Object.entries<any>(SID || {})
-            .filter(([, { isSelected }]: [unknown, GroupItem]) => isSelected)
-            .reduce<any>((acc, [key]) => [...acc, key], []),
-          mappedTags,
-        ]
-      : mappedTags;
+    const sidArray = Object.entries<any>(SID || {})
+      .filter(([, { isSelected }]: [unknown, GroupItem]) => Boolean(isSelected))
```
Hyperkid123, line 141: Consistency nit. Above on line 129 you use `===` and here you are using the `Boolean` func. Use one or the other in both places..

src/components/GlobalFilter/tagsApi.ts
```diff
@@ -70,38 +73,109 @@ export function getAllTags({ search, activeTags, registeredWith }: TagFilterOpti
     // @ts-ignore
     registeredWith ? registeredWith : undefined,
     undefined,
-    {
-      query: generateFilter(buildFilter(workloads, SID)),
-    }
+    { query: generateFilter(buildFilter(workloads, SID)) }
   );
+
+  // @ts-ignore
```
Hyperkid123, line 79: Please do not use them, as they will be ignored. Only as a last resort.

src/components/GlobalFilter/tagsApi.ts
```diff
@@ -70,38 +73,109 @@ export function getAllTags({ search, activeTags, registeredWith }: TagFilterOpti
...
+    isLoaded: true,
+    // Update pagination state
+    page: pagination?.page || 1,
+    perPage: pagination?.perPage || 10,
+    // Transform API results into the required GlobalFilterTag[] structure for TagsModal
+    items: [
+      {
+        id: 'tags-group',
+        name: 'Tags',
+        tags: ((response as any).data?.results || (response as any).results || [])
```
Hyperkid123, line 91: Please do not use `as any`. You should typeguard the result variable somwhere above instead. And you wont have to constantly cast tit as any

src/components/GlobalFilter/tagsApi.ts
```diff
@@ -70,38 +73,109 @@ export function getAllTags({ search, activeTags, registeredWith }: TagFilterOpti
...
+              },
+              count: result.count || 0,
+            };
+          })
+          .filter((item: any) => item.tag.key && item.tag.id), // Additional safety check
+      },
+    ],
+    // @ts-ignore
+    count: (response as any).data?.count || (response as any).count,
+    // @ts-ignore
```
Hyperkid123, line 112: ❗

src/components/GlobalFilter/tagsApi.ts
```diff
@@ -70,38 +73,109 @@ export function getAllTags({ search, activeTags, registeredWith }: TagFilterOpti
...
+    undefined,
     // @ts-ignore
     registeredWith ? registeredWith : undefined,
     undefined,
     {
       query: generateFilter(buildFilter(workloads, SID)),
     }
   );
+
+  // @ts-ignore
```
Hyperkid123, line 134: Same here, do not use this annotation if you don't have and cast the response variable, so you don't have to constantly cast it to any.

src/components/GlobalFilter/tagsApi.ts
```diff
@@ -70,38 +73,109 @@ export function getAllTags({ search, activeTags, registeredWith }: TagFilterOpti
...
   const [workloads, SID, selectedTags] = flatTags(activeTags, false, true);
+
   const [SAP, AAP, MSSQL] = await Promise.all([
     sap.apiSystemProfileGetSapSystem(
-      selectedTags, // tags
-      (pagination && pagination.perPage) || 10,
-      (pagination && pagination.page) || 1,
-      undefined, // staleness,
+      selectedTags,
+      1,
```
Hyperkid123, line 176: Why was the pagination removed from here? I think this will be a regression.

src/components/GlobalFilter/tagsApi.ts
```diff
@@ -147,7 +213,7 @@ export async function getAllWorkloads({ activeTags, registeredWith }: TagFilterO
       undefined,
       undefined,
       undefined,
-      1, // number of items per page
```
Hyperkid123, line 150: These comments are quite useful due to the structure of the arguments. Can you keep them?

src/components/Header/MastheadMenuToggle.tsx
```diff
@@ -24,7 +21,6 @@ const MastheadMenuToggle = ({
         isSidebarOpen={isNavOpen}
         onSidebarToggle={() => {
           setIsNavOpen?.((prev) => !prev);
-          dispatch(onToggle());
```
Hyperkid123, line 27: Just for reference, the onToggle was deprecated and no longer did anything. It can be safely removed.

src/state/atoms/globalFilterAtom.ts
```diff
@@ -1,38 +1,73 @@
...
+export const tagsAtom = atom<GlobalFilterTags>({
+  isLoaded: false,
+  items: [],
+  total: 0,
+  count: 0,
+  page: 1,
+  perPage: 10,
+});
+
+export const sidsAtom = atom<GlobalFilterSIDs>({
```
Hyperkid123, line 25: The GlobalFilterSIDs and GlobalFilterTags are the same. I suggest creating a generic type and then creating these specific aliases from it.

The GlobalFilterWorkloads type can then use that as a base and extend it.

Hyperkid123 (changes requested):

src/@types/types.d.ts
```diff
@@ -201,7 +200,7 @@ export type ChromeNavItemProps = {
   className?: string;
   active?: boolean;
   appId: string;
-  notifier?: string;
+  notifier?: atom<boolean>;
```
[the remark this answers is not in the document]
JetyAdam: You're right - there is no reason for this change, thus making it a string again.

src/chrome/__tests__/globalFilterToggle.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    it('should set globalFilterHiddenAtom to true when called with true', () => {
+      const originalSet = chromeStore.set;
+      chromeStore.set = store.set.bind(store);
+
+      chromeStore.set(globalFilterHiddenAtom, true);
+
+      const result = store.get(globalFilterHiddenAtom);
+      expect(result).toBe(true);
+
+      chromeStore.set = originalSet;
```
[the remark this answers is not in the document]
JetyAdam: Great suggestion! I switched it to jest.

PR #3098, opened 2025-05-26
Return back api.console scope

kovayur:
Reference tickets: RHCLOUD-29135, ROX-29455
Returning the scope back, since ACS is ready to consume both `sub` and `deprecated_sub` claims.


Hyperkid123:
I've tested the issues in stage, clicked trough apps, things seem to be working.

Hyperkid123 (approved):

PR #3106, opened 2025-05-29
fix(all-services): fix all services redesign filter

karelhala:
### Description

Currently filtering by bundle name breaks the filtered item, this PR fixes it by sending whole subtree if it matches title. Also add empty state when no items are found

### Before

Screenshot 2025-05-29 at 10 12 01

### After

Screenshot 2025-05-29 at 10 12 16

### JIRA

[RHCLOUD-40295](https://issues.redhat.com/browse/RHCLOUD-40295)

## Summary by Sourcery

Fix filtering in All Services redesign to return full matching subtrees and show an empty state when no items match the filter.

Bug Fixes:
- Return the entire item subtree when its title matches the filter in filterNavItemsByTitle, preventing broken navigation items.

Enhancements:
- Display a PatternFly empty state with a clear filters button when no services match the search in both redesign and legacy layouts.


riccardo-forina (approved):
LGTM

Hyperkid123 (approved):

PR #3120, opened 2025-06-12
added ouia id to User Access menu

radekkaluzik:
added ouia id to User Access menu

https://issues.redhat.com/browse/RHCLOUD-40378

### DIFF
--- a/src/components/Header/SettingsToggle.tsx
+++ b/src/components/Header/SettingsToggle.tsx
@@ -48,6 +48,7 @@ const SettingsToggle = (props: SettingsToggleProps) => {
             onClick={onClick}
             key={title}
             ouiaId={rest.ouiaId ?? title}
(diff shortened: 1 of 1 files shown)


src/components/Header/SettingsToggle.tsx
```diff
@@ -48,6 +48,7 @@ const SettingsToggle = (props: SettingsToggleProps) => {
             onClick={onClick}
             key={title}
             ouiaId={rest.ouiaId ?? title}
+            data-ouia-component-id={rest.ouiaId ?? (title === 'User Access' ? 'user-access' : undefined)}
```
Hyperkid123, line 51: We should extend the config of the dropdown items. The settings toggle is supposed to be generic.

Hyperkid123 (approved):

PR #3136, opened 2025-07-07
Fix GLOBAL_FILTER_UPDATE event to dispatch on mount

JetyAdam:
https://issues.redhat.com/browse/RHCLOUD-41000

Fixed the GLOBAL_FILTER_UPDATE event to dispatch on mount. Added unit test: "should dispatch GLOBAL_FILTER_UPDATE event immediately when listener is added".

### DIFF
--- a/src/chrome/create-chrome.ts
+++ b/src/chrome/create-chrome.ts
@@ -130,6 +130,14 @@ export const createChromeContext = ({
       // Add the callback to the listeners (cast to GenericCB for GLOBAL_FILTER_UPDATE)
       eventListeners.get('GLOBAL_FILTER_UPDATE')!.set(listenerId, callback as GenericCB);
 
+      // Dispatch current state to new listener immediately (on mount)
+      try {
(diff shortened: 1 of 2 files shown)


src/chrome/create-chrome.ts
```diff
@@ -130,6 +130,14 @@ export const createChromeContext = ({
...
       eventListeners.get('GLOBAL_FILTER_UPDATE')!.set(listenerId, callback as GenericCB);
 
+      // Dispatch current state to new listener immediately (on mount)
+      try {
+        const currentSelectedTags = chromeStore.get(selectedTagsAtom);
+        (callback as GenericCB)({ data: currentSelectedTags });
+      } catch (error) {
+        console.error('Error dispatching initial GLOBAL_FILTER_UPDATE state:', error);
+      }
+
```
Hyperkid123, line 140: I think the placement here is a bit off. I think this should be triggered automatically when the atom is changed.

I don't think we know yet know what is the expected outcome here.

JetyAdam:
Closing the PR. The issue seems to be in Malware.

PR #3138, opened 2025-07-08
Add ask RH help panel link

apinkert:
For [RHCLOUD-40983](https://issues.redhat.com/browse/RHCLOUD-40983). Need to add the flag, but ready to go otherwise.

image


src/components/Header/Tools.tsx
```diff
@@ -187,6 +188,12 @@ const Tools = () => {
         },
       ]
     : [
+        {
```
karelhala, line 191: This should also be enabled if Help panel is enabled. We should move it and show/hide it similarly as to how `enableGlobalLearningResourcesPage` is used.

apinkert:
@karelhala is this the correct dropdown logic for `helpPanelEnabled` WRT the ask Redhat item?

karelhala (approved):

PR #3144, opened 2025-07-22
Konflux cypress run

Hyperkid123:


.tekton/insights-chrome-pull-request.yaml
```diff
@@ -670,6 +733,16 @@ spec:
       optional: true
   taskRunTemplate:
     serviceAccountName: build-pipeline-insights-chrome
+    podTemplate:
+      hostAliases:
+      - ip: "127.0.0.1"
```
catastrophe-brandon, line 738: I'm unsure how the network stack is set up inside the container/pod. I've occasionally bumped into the issue where locahost doesn't work but the internal adapter address (0.0.0.0) does.

PR #3150, opened 2025-07-31
fix(va): do not render VA component when conditions are not met

Hyperkid123:
## Summary by Sourcery

Add comprehensive unit tests for the VirtualAssistant component and clean up its showAssistant state logic.

Enhancements:
- Simplify showAssistant state update to directly assign the boolean match result instead of using an if-statement.

Tests:
- Add unit tests covering VirtualAssistant rendering and prop passing under both open-config and routes modes, with mocked Scalprum components and feature flags.


Hyperkid123:
@sourcery-ai review

PR #3164, opened 2025-08-21
adding media queries to fix styling issues in small dimentions in the feedback modal

tahmidefaz:
fixes: [RHCLOUD-40959](https://issues.redhat.com/browse/RHCLOUD-40959)

Adding CSS media query to fix button overlap issues.

Before:

image

After (modal becomes scrollable, instead of the buttons overlapping):

image

## Summary by Sourcery

Bug Fixes:
- Fix button overlap in the feedback modal on short or narrow screens by adjusting the button container to static positioning and adding top margin via a media query to enable scrolling.

### DIFF
--- a/src/components/Feedback/Feedback.scss
+++ b/src/components/Feedback/Feedback.scss
@@ -88,7 +88,7 @@ button.pf-v6-c-button.chr-c-button-feedback {
   }
 }
 
-@media (max-height: 450px), (max-width: 700px) {
+@media (max-height: $pf-v6-global--breakpoint--sm), (max-width: $pf-v6-global--breakpoint--md) {
   .chr-c-feedback-buttons {
     margin-top: var(--pf-t--global--spacer--md);
     position: static;
--- a/src/components/Feedback/Feedback.scss
+++ b/src/components/Feedback/Feedback.scss
@@ -1,3 +1,5 @@
+@import '~@patternfly/patternfly/sass-utilities/scss-variables';
+
 button.pf-v6-c-button.chr-c-button-feedback {
   svg {
     margin-right: var(--pf-t--global--spacer--sm);
--- a/src/components/Feedback/Feedback.scss
+++ b/src/components/Feedback/Feedback.scss
@@ -87,3 +87,11 @@ button.pf-v6-c-button.chr-c-button-feedback {
     border-bottom: 1px solid var(--pf-t--global--border--color--default);
   }
 }
+
+@media (max-height: 450px), (max-width: 700px) {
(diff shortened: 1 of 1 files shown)


catastrophe-brandon (approved):

src/components/Feedback/Feedback.scss
```diff
@@ -87,3 +87,11 @@ button.pf-v6-c-button.chr-c-button-feedback {
     border-bottom: 1px solid var(--pf-t--global--border--color--default);
   }
 }
+
+@media (max-height: 450px), (max-width: 700px) {
```
florkbr, line 91: @tahmidefaz if possible we should use patternfly's baked in break points for this? https://www.patternfly.org/developer-resources/global-css-variables/#breakpoint-variables-and-class-suffixes

image

tahmidefaz: @florkbr Thanks! Didn't know that was a things. Updated the media query condition with breakpoint vars. 👍

Hyperkid123:
Do we even need the Sass variables? All these should be usable as regular CSS variables.

tahmidefaz:
> @tahmidefaz I checked this branch out locally and tried running the UI. I see the following error during the build:
> 
> ```
> ERROR in ./src/components/Feedback/Feedback.scss (./src/components/Feedback/Feedback.scss.webpack[javascript/auto]!=!./node_modules/css-loader/dist/cjs.js!./node_modules/resolve-url-loader/index.js??ruleSet[1].rules[1].use[2]!./node_modules/sass-loader/dist/cjs.js??ruleSet[1].rules[1].use[3]!./src/components/Feedback/Feedback.scss)
> Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
> Undefined variable.
>    ╷
> 91 │ @media (max-height: $pf-v6-global--breakpoint--sm), (max-width: $pf-v6-global--breakpoint--md) {
> ```
> 
> In order to use these patternfly variables, you'd need to import them: `@import '~@patternfly/patternfly/sass-utilities/scss-variables';`
> 
> Once I added that import, the styles do look better.
> 
> Past that, I also just realized we may get this for free without requiring custom SCSS if we use a [ModalFooter](https://www.patternfly.org/components/modal/#modalfooter) for the [buttons in the modal](https://github.com/RedHatInsights/insights-chrome/blob/master/src/components/Feedback/FeedbackForm.tsx#L132-L149).
> 
> Either way however, we will need to fix the build errors then resolve the pipeline issues before we can merge.

@florkbr Updated with the scss import statement.

PR #3166, opened 2025-08-26
feat(search): add support for external links in index

florkbr:
## Summary by Sourcery

Add support for external links in the local search by introducing an isExternal flag, including external URLs in the search index, and propagating this flag to the UI for proper rendering.

New Features:
- Introduce isExternal flag on search entries and schema, removing filtering to include external URLs in the index
- Propagate isExternal into search results and pass to ChromeLink component for proper external link rendering

### DIFF
--- a/src/utils/localSearch.ts
+++ b/src/utils/localSearch.ts
@@ -21,6 +21,7 @@ type ResultItem = {
   bundleTitle: string;
   pathname: string;
   id: string;
+  isExternal?: boolean;
 };
 
 const resultCache: {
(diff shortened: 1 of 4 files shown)


src/utils/localSearch.ts
```diff
@@ -155,13 +156,14 @@ export const localQuery = async (db: any, term: string, env: ReleaseEnv = Releas
...
     for (let i = 0; i < Math.min(10, validResults.length); i += 1) {
-      const { title, description, bundleTitle, pathname, id } = validResults[i];
+      const { title, description, bundleTitle, pathname, id, isExternal } = validResults[i];
       results.push({
         title: highlightText(term, title, 'title'),
         description: highlightText(term, description, 'description'),
         bundleTitle,
         pathname,
         id,
+        isExternal,
```
florkbr, line 166: @Hyperkid123 looking for your 2 cents here:
1. should we surface `isExternal` on the frontend CRD search entries themselves? That would give tenants complete control over specifying if a link should open in a new tab etc.
2. OR should we use the check you originally had for `.startsWith('/')`?

If we do the former, we will need all my changes to pass `isExternal` to the local database itself. If not, I can probably remove the changes to the atom/database and push the `.startsWith('/')` check deeper into the search component and avoid some diff.

PR #3173, opened 2025-09-03
fix(transformScalprumManifest): temp fix to override FEO public path

florkbr:


shannon-donahue (approved):
lgtm

src/components/RootApp/transformScalprumManifest.ts
```diff
@@ -11,13 +11,15 @@ function transformScalprumManifest(manifest: PluginManifest, config: ScalprumCon
       loadScripts: [],
     };
   }
+  let baseUrl = manifest.baseURL;
+  if (baseUrl === 'auto') {
```
florkbr, line 15: Actually, if we want to deliver this in commercial we would need to add `&& !manifest.cdnPath` check

PR #3174, opened 2025-09-03
fix(transformScalprumManifest): update root path to be relative to /apps

florkbr:
## Summary by Sourcery

Bug Fixes:
- Set baseURL of manifests with ‘auto’ to `/apps/<name>/` rather than `/`

Follow up to https://github.com/RedHatInsights/insights-chrome/pull/3173


shannon-donahue (approved):
LGTM

PR #3175, opened 2025-09-05
fix: Exclude trustarc url from auth header injection

charlesmulder:
## Summary by Sourcery

Extend the auth header injection logic by adding TrustArc analytics URLs to the exclusion list, exporting the isExcluded helper, and updating relevant tests and dependencies

Bug Fixes:
- Exclude TrustArc analytics endpoints from auth header injection

Enhancements:
- Add TrustArc analytics URL pattern to the exclusion list and export the isExcluded function

Build:
- Introduce @faker-js/faker as a devDependency
- Move pf-5-styles from devDependencies to dependencies

Tests:
- Add unit tests for isExcluded to cover excluded URLs, non-excluded URLs, and edge cases


src/utils/iqeEnablement.ts
```diff
@@ -207,6 +210,7 @@ export function init(chromeStore: ReturnType<typeof createStore>, authRef: React
 
 const qe = {
   init,
+  isExcluded,
```
florkbr, line 213: @charlesmulder not blocking, but the name `isExcluded` is not very descriptive on _what_ exactly we are excluding and why (outside the context of this file). We may eventually want to update the name of this function to be a bit more clear to others if we expose it. 

Past that, @catastrophe-brandon could you forsee any reason this would cause issues with IQE? I don't think it would, but I just wanted to check.

src/utils/iqeEnablement.test.js
```diff
@@ -18,3 +18,48 @@ describe('iqeEnablement', () => {
...
+    // random URLs
+    'https://example.com',
+    'https://api.redhat.com/some/endpoint',
+    'https://different-domain.com/analytics',
+    'https://api.example.com/upgrades_info',
+  ])('does not exclude %s', (url) => {
+    expect(iqeEnablement.isExcluded(url)).toBe(false);
+  });
+
+  // edge cases
```
florkbr, line 56: 👍 for tests with the code!

src/utils/iqeEnablement.ts
charlesmulder: @florkbr That's good thinking. Any name suggestions?

charlesmulder: @florkbr How about `isAuthHeaderDisabled`? It's a quick addition to this PR.

PR #3183, opened 2025-09-22
testing memory leak potential fix

shannon-donahue:
## Summary by Sourcery

Use a factory function and useMemo to provide a stable transformPluginManifest reference and update its dependencies to prevent unnecessary re-creations.

Enhancements:
- Introduce createTransformFunction factory to generate a stable transformScalprumManifest reference based on config
- Memoize the transformManifest function in ChromeApiRoot using useMemo with config as a dependency
- Add transformManifest to the dependency array of the pluginSDKOptions useMemo hook


karelhala:
@shannon-donahue is this PR still required?

PR #3188, opened 2025-09-25
FEO migration

tahmidefaz:
## Summary by Sourcery

Migrate the frontend to the FEO platform by updating related dependencies and enabling the FEO configuration flag in the deployment spec.

Enhancements:
- Bump @redhat-cloud-services/frontend-components-config-utilities to 4.7.0
- Upgrade @redhat-cloud-services/frontend-components to 7.0.9
- Upgrade @redhat-cloud-services/frontend-components-utilities to 7.0.8
- Enable feoConfigEnabled in the Chrome frontend.yml deployment specification


Hyperkid123:
@tahmidefaz we need the custom webpack config here as it's a shell app. There may be issue with the frontend resource validation as it may not be in the default location.

tahmidefaz:
Closing this PR for now. We decided to tackle this under a tech debt later on. Migrating FEO for `insights-chrome` isn't high prio at the moment.

The build was hanging because the location of the CRD is hardcoded under the proxy config for the `config-utils` package: [Here](https://github.com/RedHatInsights/frontend-components/blob/71b0702228109bcbbad2a441a463b14dac97494d/packages/config-utils/src/proxy.ts#L137).

The proxy tries to watch for changes to the CRD under that default location (`deploy/frontend.yaml`) forever. Since, insights-chrome has the CRD defined under the `./frontend.yml`, the proxy doesn't pick it up and hangs forever.

PR #3294, opened 2025-11-13
fix(notentitled): do not hard code console domain

florkbr:
## Summary by Sourcery

Bug Fixes:
- Replace absolute console.redhat.com image URLs with relative paths in notEntitledData


cypress/snapshots/cypress/component/NotEntitledModal/NotEntitledModal.cy.tsx/NonEntitledModal -- should render modal for ansible.snap.png
florkbr: FWIW this is not a great snapshot to keep (since the relative URL won't load unless you run chrome locally while running the tests). For now I've updated them so they should pass and we can avoid the hard-coded domains.

PR #3299, opened 2025-11-18
removing appendTo inline prop to fix z-index issues

tahmidefaz:
`appendTo: 'inline'` is making us ignore the z-index for this dropdown.

## Summary by Sourcery

Bug Fixes:
- Remove appendTo: 'inline' from Dropdown popperProps in UserToggle to fix z-index issues.


justinorringer:
Mind showing a screenshot for before/after?

tahmidefaz:
> Mind showing a screenshot for before/after?

@justinorringer this on its own will not have any visible changes. However, once https://github.com/RedHatInsights/astro-virtual-assistant-frontend/pull/348 is merged, it will look something like this:

Untitled3

PR #3302, opened 2025-11-25
fix: use css classes, rules, and variables to use PF PF values

Hyperkid123:
## Summary by Sourcery

Align layout and search UI styling with PatternFly utility classes and classnames-based conditional styling.

Enhancements:
- Replace custom search feedback icon color toggling with conditional PatternFly utility classes using classnames.
- Update global filter chip group and search input toggle to use conditional PatternFly utility and custom classes instead of inline conditionals or bespoke classes.
- Adjust 'viewing as account' banner and ChromeRoutes scope to rely on PatternFly flex and spacing utilities instead of custom layout classes.
- Simplify or remove redundant custom search and layout SCSS in favor of existing utilities.
- Ensure search title text displays correctly inline by adding a PatternFly display utility class.

### DIFF
--- a/.tekton/insights-chrome-dev-pull-request.yaml
+++ b/.tekton/insights-chrome-dev-pull-request.yaml
@@ -43,7 +43,7 @@ spec:
         - name: name
           value: show-sbom
         - name: bundle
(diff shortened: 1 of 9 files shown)


Hyperkid123:
e2e tests are down cause the stage refresh. We can wait till its back.

PR #3478, opened 2026-04-21
chore(types): apply types-react-codemod preset-19 for React 19 compat

charlesmulder:
## Why                                                                                                                                                                     
                                                                                                                                                                             
  Prepares insights-chrome for the React 19 type system by fixing two categories of breaking changes that `@types/react@19` introduces. Applied via `npx types-react-codemod@latest preset-19 ./src`.
                                                                                                                                                                             
  ## What changed                                                

  ### `useRef<T>()` → `useRef<T>(undefined)`                                                                                                                                 
  React 19 removes the zero-argument `useRef` overload. Passing `undefined` explicitly is backward-compatible with `@types/react@18` and will be required under
  `@types/react@19`.                                                                                                                                                         
                                                                 
  Files: `SegmentProvider.tsx`, `ChromeLink.tsx`, `ScalprumRoot.tsx`, `useChromeServiceEvents.ts`                                                                            
                                                                 
  ### `VoidFunctionComponent` → `FunctionComponent`                                                                                                                          
  `VFC`/`VoidFunctionComponent` is removed entirely from `@types/react@19`. `FunctionComponent` is the correct replacement and is available in all React versions.
                                                                                                                                                                             
  Files: `Stratosphere/Footer.tsx`, `Stratosphere/ProductCard.tsx`, `Stratosphere/RedirectBanner.test.tsx`                                                                   
                                                                                                                                                                             
  ## What was intentionally excluded                                                                                                                                         
                                                                 
  `RefObject<T>` → `RefObject<T | null>` prop type changes (also emitted by `preset-19`) were reverted. Those are only valid after the `@types/react@19` bump — applying them
   now causes build failures against v18 JSX intrinsic types. They will be re-applied in RHCLOUD-44216.
                                                                                                                                                                             
  ## Test plan                                                   
  - [x] `npm run verify` passes
  - [x] No TypeScript errors (`npx tsc --noEmit`)                                                                                                                            
                                                                                                                                                                             
  Refs: RHCLOUD-44375

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Refactor**
  * Internal TypeScript and initialization cleanups across hooks, components and tests: explicit ref initializations and component typing adjustments. No functional or UI changes; improves type safety and consistency.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/analytics/SegmentProvider.tsx
+++ b/src/analytics/SegmentProvider.tsx
@@ -103,7 +103,7 @@ const SegmentProvider: React.FC<React.PropsWithChildren> = ({ children }) => {
   const isITLessEnv = ITLess();
   const isDisabled = localStorage.getItem('chrome:segment:disable') === 'true' || isITLessEnv;
   const disableIntegrations = localStorage.getItem('chrome:analytics:disable') === 'true' || isITLessEnv;
-  const analytics = useRef<AnalyticsBrowser>();
+  const analytics = useRef<AnalyticsBrowser>(undefined);
   const analyticsLoaded = useRef(false);
   const { user } = useContext(ChromeAuthContext);
   const isPreview = useAtomValue(isPreviewAtom);
--- a/src/components/ChromeLink/ChromeLink.tsx
+++ b/src/components/ChromeLink/ChromeLink.tsx
@@ -33,7 +33,7 @@ const LinkWrapper: React.FC<LinkWrapperProps> = memo(
     const moduleRoutes = useAtomValue(moduleRoutesAtom);
     const triggerNavListener = useSetAtom(triggerNavListenersAtom);
     const moduleEntry = useMemo(() => moduleRoutes?.find((route) => href?.includes(route.path)), [href, appId]);
-    const preloadTimeout = useRef<NodeJS.Timeout>();
+    const preloadTimeout = useRef<NodeJS.Timeout>(undefined);
     let actionId = href.split('/').slice(2).join('/');
     if (actionId.includes('/')) {
       actionId = actionId.split('/').pop() as string;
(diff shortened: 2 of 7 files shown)


justinorringer (approved):
LGTM! Weird that setting undefined will be required but w/e

charlesmulder:
Good question! It's about what TypeScript return type `useRef` gives you depending on the initial value.

`useRef` returns a ref object — just `{ current: T }`. React defines two variants:

- `MutableRefObject<T>` — `.current` is **writable** (for instance variables you manage yourself)
- `RefObject<T>` — `.current` is **read-only** (for DOM refs React manages)

In React 18, `useRef` had three type overloads:

```typescript
function useRef<T>(initialValue: T): MutableRefObject<T>;
function useRef<T>(initialValue: T | null): RefObject<T>;
function useRef<T>(): MutableRefObject<T | undefined>;
```

These were ambiguous — for example, `useRef<HTMLDivElement>(null)` could match overload 1 (with `T = HTMLDivElement | null`, mutable) or overload 2 (with `T = HTMLDivElement`, read-only). And the zero-argument form looked similar to passing `null` — both mean "starts empty" — yet gave completely different mutability.

React 19 removes the zero-argument overload and makes intent explicit:

- `useRef<T>(undefined)` returns `MutableRefObject<T | undefined>` — mutable `.current`, starts as `undefined`
- `useRef<T>(null)` returns `RefObject<T | null>` — read-only `.current`, meant for DOM refs

If we passed `null` instead of `undefined`, `.current` would become read-only and assignments like `analytics.current = new AnalyticsBrowser(...)` would fail at the type level.

So `undefined` explicitly signals "this is a mutable ref that starts empty, not a DOM ref."

Hyperkid123 (approved):

PR #3542, opened 2026-05-27
feat: add high contrast toggle, add tests

OlhaTmlk:
### Description
Adds high contrast mode support (`pf-v6-theme-high-contrast`) to the platform, gated behind the `platform.chrome.high-contrast` feature flag. 

When enabled, users can choose via a ToggleGroup in the settings dropdown between 
- **System** (follows `browser prefers-contrast: more` preference), 
- **Default** (no high contrast), 
- **High** contrast 

The preference is persisted in localStorage under `chrome:high-contrast.`

[RHCLOUD-47434](https://redhat.atlassian.net/browse/RHCLOUD-47434)

---

### Screenshots
Screenshot 2026-05-27 at 10 48 51
Screenshot 2026-05-27 at 10 49 00

[RHCLOUD-47434]: https://redhat.atlassian.net/browse/RHCLOUD-47434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a feature-flagged high-contrast control in Settings with **System**, **Default**, and **High contrast** options, including persistence and system-preference handling.
  * Added localized strings for the contrast options.
  * Enhanced Settings dropdown groups to support mutually exclusive **custom group content** (vs items) and improved divider placement for visible groups.
* **Tests**
  * Added unit tests for high-contrast behavior and hook setters (including feature-flag gating and system preference).
  * Added component test coverage for the Settings contrast toggle, plus updated existing Tools/Settings mocks for custom group content.

### DIFF
--- a/src/components/Header/Tools.tsx
+++ b/src/components/Header/Tools.tsx
@@ -177,30 +177,30 @@ const Tools = () => {
       ),
     },
     {
-      title: 'Contrast',
+      title: intl.formatMessage(messages.contrast),
       isHidden: !isHighContrastEnabled,
       customContent: (
-        <ToggleGroup aria-label="Contrast mode" className="pf-v6-u-mx-md pf-v6-u-my-sm">
+        <ToggleGroup aria-label={intl.formatMessage(messages.contrast)} className="pf-v6-u-mx-md pf-v6-u-my-sm">
           <ToggleGroupItem
-            text="System"
+            text={intl.formatMessage(messages.contrastSystem)}
             buttonId="contrast-system"
             isSelected={contrastMode === HighContrastVariants.system}
             onChange={() => setSystemContrast()}
-            aria-label="System contrast"
+            aria-label={intl.formatMessage(messages.contrastSystem)}
           />
           <ToggleGroupItem
-            text="Default"
+            text={intl.formatMessage(messages.contrastDefault)}
             buttonId="contrast-default"
             isSelected={contrastMode === HighContrastVariants.default}
             onChange={() => setDefaultContrast()}
-            aria-label="Default contrast"
+            aria-label={intl.formatMessage(messages.contrastDefault)}
           />
           <ToggleGroupItem
-            text="High contrast"
+            text={intl.formatMessage(messages.contrastHigh)}
             buttonId="contrast-high"
             isSelected={contrastMode === HighContrastVariants.high}
             onChange={() => setHighContrast()}
-            aria-label="High contrast"
+            aria-label={intl.formatMessage(messages.contrastHigh)}
           />
         </ToggleGroup>
       ),
--- a/src/components/Header/Tools.tsx
+++ b/src/components/Header/Tools.tsx
@@ -6,6 +6,8 @@ import { Switch } from '@patternfly/react-core/dist/dynamic/components/Switch';
 import { DropdownItem } from '@patternfly/react-core/dist/dynamic/components/Dropdown';
 import { ToolbarItem } from '@patternfly/react-core/dist/dynamic/components/Toolbar';
 import { Tooltip } from '@patternfly/react-core/dist/dynamic/components/Tooltip';
+import { ToggleGroup } from '@patternfly/react-core/dist/dynamic/components/ToggleGroup';
+import { ToggleGroupItem } from '@patternfly/react-core/dist/dynamic/components/ToggleGroup';
 import QuestionCircleIcon from '@patternfly/react-icons/dist/dynamic/icons/question-circle-icon';
 import CogIcon from '@patternfly/react-icons/dist/dynamic/icons/cog-icon';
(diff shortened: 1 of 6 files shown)


src/hooks/useHighContrast.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    // Default to system mode when no preference is saved
+    localStorage.setItem('chrome:high-contrast', 'system');
+    const prefersHighContrast = window.matchMedia('(prefers-contrast: more)').matches;
+    applyHighContrast(prefersHighContrast);
+    return HighContrastVariants.system;
+  };
+
+  const [contrastMode, setContrastMode] = useState<HighContrastVariants>(getInitialMode);
+
```
karelhala, line 45: `getInitialMode()` has side effects — DOM mutations via `applyHighContrast()` and `localStorage.setItem()` — but it's used as a `useState` initializer, which React expects to be a [pure function](https://react.dev/reference/react/useState#parameters).

Consider splitting this into:
1. A **pure** function that only computes and returns the `HighContrastVariants` value
2. Moving the DOM class toggling and `localStorage` writes into the `useEffect` below

This would also simplify issue below about the effect dependencies.

src/hooks/useHighContrast.ts
```diff
@@ -0,0 +1,80 @@
...
+  useEffect(() => {
+    const newMode = getInitialMode();
+    setContrastMode(newMode);
+  }, [isHighContrastEnabled]);
```
karelhala, line 50: `getInitialMode` is defined inside the hook body, so it's recreated every render and closes over `isHighContrastEnabled` — making it an implicit dependency of this effect that ESLint's `exhaustive-deps` rule would flag.

If you extract the mode-computation logic as a pure function **outside** the hook (taking `isHighContrastEnabled` as a parameter), the data flow becomes explicit and the function isn't recreated each render:

```ts
const computeInitialMode = (flagEnabled: boolean): HighContrastVariants => { ... };

export const useHighContrast = () => {
  const isHighContrastEnabled = useFlag('platform.chrome.high-contrast');
  const [contrastMode, setContrastMode] = useState(() => computeInitialMode(isHighContrastEnabled));

  useEffect(() => {
    const mode = computeInitialMode(isHighContrastEnabled);
    setContrastMode(mode);
    applyHighContrast(/* derive from mode */);
  }, [isHighContrastEnabled]);
};
```

src/components/Header/Tools.tsx
```diff
@@ -152,6 +157,36 @@ const Tools = () => {
         },
       ],
     },
+    {
+      title: 'Contrast',
+      isHidden: !isHighContrastEnabled,
```
karelhala, line 181: Passing `items: []` alongside `customContent` works but is a bit of a code smell — a future caller could accidentally provide both and not realize `items` gets silently ignored.

Consider making `SettingsToggleDropdownGroup` a discriminated union so a group has *either* `items` or `customContent`:

```ts
type SettingsToggleDropdownGroup = {
  title?: string;
  isHidden?: boolean;
} & (
  | { items: SettingsToggleDropdownItem[]; customContent?: never }
  | { customContent: ReactNode; items?: never }
);
```

Low priority — fine as a follow-up.

src/components/Header/Tools.tsx
```diff
@@ -152,6 +157,36 @@ const Tools = () => {
         },
       ],
     },
+    {
```
karelhala, line 179: `'Contrast'`, `'System'`, `'Default'`, and `'High contrast'` are hardcoded English strings. Since `intl` is already available in this component (line 97), consider wrapping these in `intl.formatMessage()` for i18n consistency — especially since this is new code and a good opportunity to set the right pattern.

src/hooks/useHighContrast.ts
```diff
@@ -0,0 +1,80 @@
...
+    applyHighContrast(true);
+    localStorage.setItem('chrome:high-contrast', 'high');
+  };
+
+  const setSystemContrast = () => {
+    setContrastMode(HighContrastVariants.system);
+    const prefersHighContrast = window.matchMedia('(prefers-contrast: more)').matches;
+    applyHighContrast(prefersHighContrast);
+    localStorage.setItem('chrome:high-contrast', 'system');
+  };
```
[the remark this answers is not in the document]
OlhaTmlk: Intentionally left as read-on-load — useTheme (dark mode) works the same way and we'd rather keep the two hooks consistent. The addEventListener/removeEventListener stubs in tests are there to satisfy the matchMedia mock shape, not to signal a planned subscription.

karelhala:
@coderabbitai full review

karelhala:
@coderabbitai full review

karelhala (approved):

PR #3560, opened 2026-06-05
fix: report original errors to Sentry from DefaultErrorComponent

johnsonm325:
### Description
Pass props.error directly to captureException instead of wrapping it in a new Error, which replaced the real stack trace with the error boundary's. Use Sentry's contexts.react.componentStack and extra fields so bundle, app, and the React component tree are actually recorded instead of being dropped.

### Checklist

### AI disclosure
Assisted by: Claude Code

## Summary by CodeRabbit

* **Tests**
  * Added tests for error-message extraction and for error-reporting behavior across various error types, including no-report when no error exists.

* **Bug Fixes**
  * More robust error-to-string handling, consistent "Something went wrong:" messaging, improved chunk-load detection and safer gateway error checks, and stricter rendering of error details.

* **Refactor**
  * Centralized error-message logic and updated error reporting to include React component stack and structured extra metadata.

### DIFF
--- a/src/components/ErrorComponents/DefaultErrorComponent.test.tsx
+++ b/src/components/ErrorComponents/DefaultErrorComponent.test.tsx
@@ -89,21 +89,26 @@ describe('DefaultErrorComponent', () => {
       });
     });
 
-    it('reports string errors to Sentry without wrapping in a new Error', async () => {
+    it('reports string errors to Sentry with a Something went wrong prefix', async () => {
       const error = 'Cannot destructure property future of useContext as it is null';
 
       renderWithProviders(<DefaultErrorComponent error={error} />);
 
       await waitFor(() => {
-        expect(Sentry.captureException).toHaveBeenCalledWith(error, {
-          contexts: {
-            react: { componentStack: undefined },
-          },
-          extra: {
-            bundle: 'insights',
-            app: 'insights',
-          },
-        });
+        expect(Sentry.captureException).toHaveBeenCalledWith(
+          expect.objectContaining({
+            message: 'Something went wrong: Cannot destructure property future of useContext as it is null',
+          }),
+          {
+            contexts: {
+              react: { componentStack: undefined },
+            },
+            extra: {
+              bundle: 'insights',
+              app: 'insights',
+            },
+          }
+        );
       });
     });
 
--- a/src/components/ErrorComponents/DefaultErrorComponent.test.tsx
+++ b/src/components/ErrorComponents/DefaultErrorComponent.test.tsx
@@ -112,6 +112,34 @@ describe('DefaultErrorComponent', () => {
       });
     });
 
+    it('reports plain object errors with message property', async () => {
+      const error = { message: 'cross-frame error' };
+
+      renderWithProviders(<DefaultErrorComponent error={error} />);
+
+      await waitFor(() => {
+        expect(Sentry.captureException).toHaveBeenCalledWith(
+          expect.objectContaining({
+            message: 'Something went wrong: cross-frame error',
+          }),
+          expect.any(Object)
+        );
+      });
+    });
+
+    it('reports non-standard error types with fallback message', async () => {
+      renderWithProviders(<DefaultErrorComponent error={42} />);
+
+      await waitFor(() => {
+        expect(Sentry.captureException).toHaveBeenCalledWith(
+          expect.objectContaining({
+            message: 'Something went wrong: Unhandled UI runtime error',
+          }),
+          expect.any(Object)
+        );
+      });
+    });
+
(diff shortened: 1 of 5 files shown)


src/components/ErrorComponents/DefaultErrorComponent.tsx
```diff
@@ -39,17 +39,17 @@ const DefaultErrorComponent = (props: DefaultErrorComponentProps) => {
...
     const sentryId =
       props.error &&
-      Sentry.captureException(new Error(exceptionMessage), {
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        bundle: getUrl('bundle'),
-        app: getUrl('app'),
-        error: (props.error instanceof Error && props.error?.message) || props.error,
-        trace: props.errorInfo?.componentStack || (props.error instanceof Error && props.error?.stack) || props.error,
+      Sentry.captureException(props.error, {
```
adonispuente: Id like to to add something like 'exceptionMsg = 'Something went wrong + props.error' just so we can still distinguish SWW page crash errors in sentry

johnsonm325:
@coderabbitai review

src/components/ErrorComponents/DefaultErrorComponent.tsx
```diff
@@ -34,36 +34,45 @@ export type DefaultErrorComponentProps = {
...
+    return error;
+  }
+  if (error instanceof Error) {
+    return error.message;
+  }
+  if (typeof error === 'object' && error !== null && 'message' in error) {
+    return String((error as { message: unknown }).message);
+  }
+  return 'Unhandled UI runtime error';
+};
```
karelhala: **DRY: `getErrorMessage` is duplicated from `chunkLoadErrorUtils.ts:6-12`.**

An identical-purpose function already exists in `src/utils/chunkLoadErrorUtils.ts`. The two implementations diverge slightly:

| | This version | `chunkLoadErrorUtils` |
|---|---|---|
| `Error` instance | Explicit `instanceof Error` branch | Falls through to `'message' in error` |
| Fallback | `'Unhandled UI runtime error'` | `''` |

The `instanceof Error` branch here is redundant since `Error` objects have a `message` property and would be caught by the `'message' in error` check anyway.

Suggestion: extract a single `getErrorMessage(error: unknown, fallback?: string)` to a shared utility (e.g. `src/utils/common.ts`) and import it in both places, so they don't drift apart.

src/components/ErrorComponents/DefaultErrorComponent.test.tsx
```diff
@@ -68,6 +69,59 @@ describe('DefaultErrorComponent', () => {
...
+      await waitFor(() => {
+        expect(screen.getByText('Something went wrong', { exact: false })).toBeInTheDocument();
+      });
+      expect(Sentry.captureException).not.toHaveBeenCalled();
+    });
+  });
+
   describe('renders error page', () => {
     it('renders Something went wrong message', () => {
       renderWithProviders(<DefaultErrorComponent error={new Error('test error')} />);
```
karelhala, line 155: **Missing test coverage: plain object errors and fallback message.**

The `getErrorMessage` helper handles three cases but only two are tested here:

1. `Error` instance — covered by the first test
2. `string` — covered by the second test
3. **Plain object with `message`** — e.g. `{ message: 'cross-frame error' }`. This is a real scenario when errors cross iframe/realm boundaries where `instanceof Error` fails. Not tested.
4. **Non-standard error types** (number, boolean, `Symbol`) — should hit the `'Unhandled UI runtime error'` fallback. Not tested.

Suggesting two additional tests:
```typescript
it('reports plain object errors with message property', async () => {
  const error = { message: 'cross-frame error' };
  renderWithProviders(<DefaultErrorComponent error={error} />);
  await waitFor(() => {
    expect(Sentry.captureException).toHaveBeenCalledWith(
      expect.objectContaining({
        message: 'Something went wrong: cross-frame error',
      }),
      expect.any(Object)
    );
  });
});

it('reports non-standard error types with fallback message', async () => {
  renderWithProviders(<DefaultErrorComponent error={42} />);
  await waitFor(() => {
    expect(Sentry.captureException).toHaveBeenCalledWith(
      expect.objectContaining({
        message: 'Something went wrong: Unhandled UI runtime error',
      }),
      expect.any(Object)
    );
  });
});
```

karelhala (commented):
Good fix — the core change (preserving original stack traces and using proper Sentry `CaptureContext` keys) is correct and addresses a real observability gap. A few suggestions inline.

karelhala (approved):

karelhala:
@coderabbitai review

PR #3561, opened 2026-06-05
Add help panel drawer panel to allservices bundle

apinkert:
Fixes [RHCLOUD-43510](https://redhat.atlassian.net/browse/RHCLOUD-43510) where the help panel would not toggle on the allservices page.

Before:
Screenshot 2026-06-05 at 2 28 25 PM

After:
Screenshot 2026-06-05 at 2 28 16 PM

[RHCLOUD-43510]: https://redhat.atlassian.net/browse/RHCLOUD-43510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added an optional notification drawer with improved focus handling for accessibility; visibility is controlled by feature flags.
* **Tests**
  * Added tests covering drawer rendering for different flag combinations and verifying open/close state behavior and state clearing.


catastrophe-brandon:
We may want to add a simple user regression test that navigates to a few different pages and confirms that the help panel drawer is visible when clicked and usable by the user. 

We don't need to cover all the pages, just the most commonly used ones such as:
* All services
* Landing Page
* User Access
* Integrations
* Notifications

I'm not sure that the test belongs here; it might be more appropriate in the learning-resources repo. What do you think?

apinkert:
@catastrophe-brandon I'm not sure that should live in the learning-resources. The drawer itself and its state management live in this repo while learning-resources only defines the help panel content that's displayed in the drawer. This drawer also displays notifications via the Notifications toggle in the header and this change also fixes the notifications drawer toggle for the allservices page.

catastrophe-brandon:
> @catastrophe-brandon I'm not sure that should live in the learning-resources. The drawer itself and its state management live in this repo while learning-resources only defines the help panel content that's displayed in the drawer. This drawer also displays notifications via the Notifications toggle in the header and this change also fixes the notifications drawer toggle for the allservices page.

Thank you, that helps provide more context. I'll review the coverage for the all services page we already have to see if we can squeeze this into an existing test.

catastrophe-brandon:
While I was testing locally, I noticed that direct navigation to /allservices still exhibits the behavior (nothing happens when I click the help button). If I navigate via the menu link things work as-expected.

apinkert:
@catastrophe-brandon I'm not seeing that running it locally. Are you navigating to `https://stage.foo.redhat.com:1337/allservices` or `https://console.stage.redhat.com:1337/allservices`? When I navigate to `https://stage.foo.redhat.com:1337/allservices` after running `npm run dev`, the drawers work as expected

src/layouts/AllServices.tsx
```diff
@@ -37,13 +40,40 @@ export type AllServicesProps = {
...
+
+  const isNotificationsEnabled = useFlag('platform.chrome.notifications-drawer');
+  const isHelpPanelEnabled = useFlag('platform.chrome.help-panel');
+  const isDrawerEnabled = isNotificationsEnabled || isHelpPanelEnabled;
+
+  useEffect(() => {
+    if (drawerPanelRef.current !== null) {
+      focusDrawer();
+    }
+  }, []);
```
karelhala: The empty dependency array means this effect runs once on mount — at that point the drawer is collapsed and `drawerPanelRef.current` is `null`, so `focusDrawer` is never called and focus never moves when the drawer opens.

The dependency should be `isNotificationsDrawerExpanded`, matching the pattern in `DefaultLayout.tsx`:

```tsx
useEffect(() => {
  if (isNotificationsDrawerExpanded && drawerPanelRef.current !== null) {
    focusDrawer();
  }
}, [isNotificationsDrawerExpanded]);
```

apinkert: @karelhala updated! good catch

catastrophe-brandon:
I'm no longer able to reproduce the issue I observed Friday in a local developer environment. Perhaps it was an environmental issue?

I have an e2e test in progress in another PR sans code fix, which fails. Once this branch is merged, I'll update and see if the test passes.

karelhala (approved):

PR #3583, opened 2026-07-02
feat(routing): update routing and footer for new appp

karelhala:
### Description

Fix nested route matching for `/lightwell` and remove footer from lightwell pages.

The `/lightwell` route was defined as an exact path match, so navigating to any subroute (e.g., `/lightwell/foo`) fell through to the `*` catch-all, rendering `DefaultLayout` instead of the Lightwell layout — resulting in a 404. Changed to `/lightwell/*` so React Router matches all nested paths and the `contentSources` federated module can handle its own sub-routing. Also removed the `Footer` prop from the Lightwell route.

---

### Screenshots

N/A — no visual UI changes beyond removing the footer on lightwell pages.

---

This mirrors how other routes like `*` (DefaultLayout) handle nested paths. The `contentSources` ScalprumComponent inside `Lightwell` manages its own routing, so it needs the wildcard to receive sub-paths.

---

### Checklist

### AI disclosure

Assisted by: Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the Lightwell page routing so nested page paths load correctly.
  * Updated page loading behavior to display a fallback while content is being prepared, resulting in a smoother experience.

### DIFF
--- a/src/components/RootApp/ScalprumRoot.tsx
+++ b/src/components/RootApp/ScalprumRoot.tsx
@@ -91,10 +91,10 @@ const ScalprumRoot = memo(
           <Route path="/security" element={<DefaultLayout />} />
           {/* TODO: Temporary hardcoded route for content-sources-frontend authed experience (RHCLOUD-48921). Revisit for a longer-term approach. */}
           <Route
-            path="/lightwell"
+            path="/lightwell/*"
             element={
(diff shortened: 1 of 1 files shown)


ochosi (approved):
I know I'm not a code owner, but this is so trivial... 😅

Hyperkid123 (approved):

PR #3592, opened 2026-07-07
fix(lightwell): eliminate theme and banner flicker on /lightwell routes (RHCLOUD-49049)

ochosi:
## Summary

Eliminates visible flicker on `/lightwell` routes caused by atoms and theme classes being initialized too late in the React lifecycle.

- **Banner flash:** `layoutBannerHiddenAtom` and `layoutForceGlassThemeAtom` started as `false` and relied on `useLayoutEffect` inside the lazy-loaded Lightwell component to flip them — but the chunk arrived after the first paint, so the BetaSwitcher banner and unstyled header briefly rendered. Fixed by initializing both atoms from `window.location.pathname` at module-eval time and eager-importing the Lightwell layout (which is small).
- **Theme flash:** Felt/glass theme classes were applied via `useEffect` in separate hooks, leaving a gap during the AppPlaceholder → Lightwell transition. Fixed by introducing `useLightwellRouteSetup`, which applies theme classes at module-eval time and reinforces them via `useLayoutEffect`.
- **Glass override bug:** `useGlassTheme` checked `!isEnabled` before `forceEnabled`, so when the `platform.chrome.glass-theme` flag was off, force-enable was silently ignored. Reordered so `forceEnabled` takes priority — "force" means unconditional. This only affects callers passing `forceEnabled=true` (currently only Lightwell); all other routes are unchanged.

## Screencast

See the related Jira.

Note that in addition to what's in this pull request, a small follow-up in content-sources-frontend will fix a resize that is already addressed in this screen capture.

## Changes

| File | What changed |
|------|-------------|
| `src/state/atoms/releaseAtom.ts` | Initialize `layoutBannerHiddenAtom` and `layoutForceGlassThemeAtom` from pathname |
| `src/hooks/useLightwellRouteSetup.ts` | New hook: applies felt + glass classes at module-eval time and via `useLayoutEffect` |
| `src/hooks/useGlassTheme.ts` | `forceEnabled` now takes priority over `!isEnabled` in both init and effect |
| `src/layouts/Lightwell.tsx` | Uses `useLightwellRouteSetup`; drops manual `useFeltTheme` + `layoutForceGlassThemeAtom` management |
| `src/components/RootApp/ScalprumRoot.tsx` | Eager-import Lightwell, remove `Suspense`/`lazyWithRetry` wrapper |
| `src/hooks/useGlassTheme.test.ts` | Updated test to match new `forceEnabled` semantics |
| `src/layouts/Lightwell.test.tsx` | Added glass class to `afterEach` cleanup |

## Non-Lightwell impact

The only shared code change is the `forceEnabled` priority reorder in `useGlassTheme`. No caller outside Lightwell passes `forceEnabled=true`, so behavior on all other routes is identical. The atom initializations are no-ops on non-Lightwell pathnames.

Fixes https://redhat.atlassian.net/browse/RHCLOUD-49049

## Test plan

- [x] `useGlassTheme` tests pass (14/14) — including updated `forceEnabled` semantics test
- [x] `Lightwell` tests pass (8/8)
- [ ] Manual: navigate to `/lightwell` — no banner flash, no unstyled → styled transition
- [ ] Manual: navigate to any non-Lightwell route — no regressions in theme toggle or BetaSwitcher

## Summary by CodeRabbit

* **New Features**
  * Improved Lightwell styling so the correct felt/glass theme is applied immediately when entering the route, reducing any visual theme flash.

* **Bug Fixes**
  * Refined Lightwell layout initialization to ensure the banner and header behavior is consistent from the start.
  * Updated Lightwell test cleanup and theme handling to reliably remove theme classes after tests and route transitions.

### DIFF
--- a/src/components/RootApp/ScalprumRoot.tsx
+++ b/src/components/RootApp/ScalprumRoot.tsx
@@ -40,9 +40,7 @@ import useDPAL from '../../analytics/useDpal';
 import { selectedTagsAtom } from '../../state/atoms/globalFilterAtom';
 import useAmplitude from '../../analytics/useAmplitude';
 import usePf5Styles from '../../hooks/usePf5Styles';
-
 const ProductSelection = lazyWithRetry(() => import('../Stratosphere/ProductSelection'));
-// TODO: Temporary hardcoded route for content-sources-frontend authed experience (RHCLOUD-48921). Revisit for a longer-term approach.
 const Lightwell = lazyWithRetry(() => import('../../layouts/Lightwell'));
 
 const useGlobalFilter = (callback: (selectedTags?: FlagTagsFilter) => any) => {
--- a/src/hooks/useLightwellRouteSetup.ts
+++ b/src/hooks/useLightwellRouteSetup.ts
@@ -1,0 +1,44 @@
+import { useLayoutEffect } from 'react';
+import { useSetAtom } from 'jotai';
+import { layoutForceGlassThemeAtom } from '../state/atoms/releaseAtom';
+import { LIGHTWELL_PATH } from '../utils/common';
+
+const FELT_THEME_CLASS = 'pf-v6-theme-felt';
+const GLASS_THEME_CLASS = 'pf-v6-theme-glass';
+
+const isLightwellRoute = window.location.pathname === LIGHTWELL_PATH || window.location.pathname.startsWith(`${LIGHTWELL_PATH}/`);
+
+if (isLightwellRoute) {
+  document.documentElement.classList.add(FELT_THEME_CLASS, GLASS_THEME_CLASS);
+}
+
+type UseLightwellRouteSetupOptions = {
+  enabled?: boolean;
+};
+
+/**
+ * Applies Lightwell felt + glass themes synchronously before paint.
+ * Also forces glass via layoutForceGlassThemeAtom for Header toolbar state.
+ *
+ * Theme classes are also applied eagerly at module-eval time (above) so they
+ * survive the AppPlaceholder → Lightwell component transition without a gap.
+ */
+const useLightwellRouteSetup = ({ enabled = true }: UseLightwellRouteSetupOptions = {}) => {
+  const setLayoutForceGlassTheme = useSetAtom(layoutForceGlassThemeAtom);
+
+  useLayoutEffect(() => {
+    if (!enabled) {
+      return;
+    }
+
+    document.documentElement.classList.add(FELT_THEME_CLASS, GLASS_THEME_CLASS);
+    setLayoutForceGlassTheme(true);
+
+    return () => {
+      document.documentElement.classList.remove(FELT_THEME_CLASS, GLASS_THEME_CLASS);
+      setLayoutForceGlassTheme(false);
+    };
+  }, [enabled, setLayoutForceGlassTheme]);
+};
+
+export default useLightwellRouteSetup;
--- a/src/state/atoms/releaseAtom.ts
+++ b/src/state/atoms/releaseAtom.ts
@@ -6,6 +6,9 @@ import { SearchPermissionsCache } from './localSearchAtom';
 import { atom } from 'jotai';
 import { userConfigAtom } from './userConfigAtom';
 import { ChromeUserConfig } from '../../utils/initUserConfig';
+import { LIGHTWELL_PATH } from '../../utils/common';
+
+const isLightwellPath = window.location.pathname.startsWith(LIGHTWELL_PATH);
 
 export const previewModalOpenAtom = atomWithToggle(false);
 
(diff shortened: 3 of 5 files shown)


src/components/RootApp/ScalprumRoot.tsx
```diff
@@ -90,14 +89,7 @@ const ScalprumRoot = memo(
...
           {/* TODO: Temporary hardcoded route for content-sources-frontend authed experience (RHCLOUD-48921). Revisit for a longer-term approach. */}
-          <Route
-            path={`${LIGHTWELL_PATH}/*`}
-            element={
-              <Suspense fallback={LoadingFallback}>
-                <Lightwell />
-              </Suspense>
-            }
-          />
+          <Route path={`${LIGHTWELL_PATH}/*`} element={<Lightwell />} />
```
karelhala: Moving Lightwell to an eager import would include its code in the bundle for all pages. It's small today, but if the Lightwell application grows over time, users on other routes would end up downloading files they don't need. Worth keeping the lazy loading pattern to stay consistent with how we handle other route-specific layouts.

src/hooks/useGlassTheme.ts
karelhala, line 1: These changes are already covered in #3591 — we can drop them from this PR to avoid merge conflicts and keep the scope focused.

src/hooks/useLightwellRouteSetup.ts
karelhala, line 1: The current theming approach is intentional — we have a theme hook that applications can use to apply theming from their side, which gives customers the flexibility to switch between glass theme and other options. Hardcoding the theme classes at module-eval time would bypass that customization layer. We want to preserve a certain level of theming flexibility for customers on consoledot.

karelhala (commented):
Thanks for digging into the flicker issue.

  A couple of things to keep in mind as we iterate on the approach:

  **Async loading in ScalprumRoot:** The route components in ScalprumRoot are loaded asynchronously by design — this keeps them out of the main chrome bundle so users only download code for the routes they
  actually visit. We'd want to preserve that pattern here.

  **Theme customization:** We need to be careful about forcing a specific theme. We provide a theme hook that applications can use to set a default, but customers should still be able to switch to their
  preferred theme to maintain UX consistency across the console. Any flicker fix should work within that constraint rather than bypassing it.

  Good catch on the `layoutBannerHiddenAtom` and `layoutForceGlassThemeAtom` initialization from the pathname!

src/components/RootApp/ScalprumRoot.tsx
ochosi: I did a bundle analyze and there was _no measurable size increase in the bundle today_ (compared with `master`) .
You're right though that if the Lightwell application grows, this could become problematic. :confused: 

Unfortunately, it's also one of the bigger sources of the flickering and re-rendering. Do you have any alternative suggestions on how to address this?

src/hooks/useLightwellRouteSetup.ts
ochosi: I get that design concern from a platform perspective, that's why I only hardcoded it into the Lightwell route. In fact for Lightwell, we **don't** want to offer customers a way to switch. All they should be able to do is switch between light and dark mode (that's per the UX design). That's why I hardcoded it into Lightwell...

src/hooks/useGlassTheme.ts
ochosi: Rebased and resolved the conflict :+1:

src/state/atoms/releaseAtom.ts
```diff
@@ -58,16 +59,21 @@ export const hidePreviewBannerAtom = atomWithToggle(initialHidePreviewBanner, as
...
 /**
  * Atom for layouts to signal that the glass theme should be force-enabled.
- * Used by the Lightwell layout to auto-enable glass theme without hardcoding URL checks.
+ * Initialized from the current pathname so the glass theme is already active
+ * before Header/Tools first renders on Lightwell routes.
  */
-export const layoutForceGlassThemeAtom = atom(false);
+export const layoutForceGlassThemeAtom = atom(
+  window.location.pathname === LIGHTWELL_PATH || window.location.pathname.startsWith(`${LIGHTWELL_PATH}/`)
+);
```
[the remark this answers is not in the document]
ochosi: The requested tests already exist and are passing. I hope coderabbitai will pick it up and mark this as resolved.

src/components/RootApp/ScalprumRoot.tsx
karelhala: The production build should be performant enough to handle this. Moreover it's mostly caused because we are going against the regular user flow, usually customer lands on landing page console.redhat.com and from that they navigate either by searching or using masthead to specific application.

ochosi: Fair, let's try without this for now and see how things shake out in prod! Thanks for the review :)

karelhala (approved):

PR #3593, opened 2026-07-07
fix(useDarkMode): export default hook

karelhala:
### Description

Add a default export for the `useDarkModeStore` hook. The hook was only available as a named export, which is inconsistent with how consuming apps typically import hooks and prevents usage via `import useDarkModeStore from '...'` syntax.

[RHCLOUD-XXXXX](https://issues.redhat.com/browse/RHCLOUD-XXXXX)

---

This is a non-breaking additive change — the existing named export is preserved. The default export is added alongside it for convenience.

---

### Checklist

### AI disclosure
Assisted by: Claude Code

## Summary by CodeRabbit

* **Chores**
  * Updated the dark mode store export to support both default and named imports. This should not change app behavior, but makes the module easier to use in other parts of the app.

### DIFF
--- a/src/state/stores/darkModeStore.ts
+++ b/src/state/stores/darkModeStore.ts
@@ -51,3 +51,5 @@ export const useDarkModeStore = () => {
     isDark: state.isDark,
   };
 };
+
+export default useDarkModeStore;


src/state/stores/darkModeStore.ts
```diff
@@ -51,3 +51,5 @@ export const useDarkModeStore = () => {
     isDark: state.isDark,
   };
 };
+
+export default useDarkModeStore;
```
eliranmal, line 55: it's hard to tell how/if this will resolve the issue i was having – importing this hook via scalprum inside an HCC app (see [this PR][2] for reference).

looking at docs and other apps' usages, we should be able to consume this module by its named export.

what confuses me is the neighboring hook in this codebase, [openShiftIntercomStore][1], do not have a default export, but presumably still work.

@karelhala, should openShiftIntercomStore be changed too?  or is it proved to work without a default export?  and if so, how?



[1]: https://github.com/karelhala/insights-chrome/blob/f4268622bbc8405ae28fadce9f7c039155bebae8/src/state/stores/openShiftIntercomStore.ts
[2]: https://github.com/RedHatInsights/uhc-portal/pull/617/changes#diff-04637ae594622d56c01d470442a9c26ac21f491df8c1a41acf898ac6bded502eR31-R35

eliranmal: hmm, i tested locally and was able to load `useOpenShiftIntercomStore` using exactly the same method (and using its named export):

```ts
const { hookResult, loading } = useRemoteHook({
  scope: 'chrome',
  module: './analytics/intercom/useOpenShiftIntercomStore',
  importName: 'useOpenShiftIntercomStore',
});
```

that makes me think that the export method isn't the problem, and this PR might be redundant.

karelhala: I think the problem might be with caching issues, but I haven't invested that deep. When I added this default export it started to work with your PR actually. Once I removed the export it was still working, so maybe forcing the chrome chunk to reload might be the issue here.

eliranmal: oh ok, i see what you're trying to do 👌 

btw, i suspected it was cache, so i've removed all container-images cache, restarted podman daemon, and re-spun the dev-server.  that didn't help for me.

eliranmal (approved):

florkbr:
@eliranmal @karelhala I wanted to give y'all some more context as to what I think happened here:

 After investigating, the export default change likely didn't fix the underlying issue. The PR discussion already confirmed this — once the export was removed, things still worked.

### Most likely root cause: type/runtime drift between repos

  The Chrome types (from frontend-components) and the Chrome runtime (from the insights-chrome-dev container image) are versioned independently. This means:

  1. The @redhat-cloud-services/types package gets updated to include `useDarkModeStore` types — so TypeScript compiles fine, no red squiggles in IDE
  2. But the actual Chrome runtime is served from the insights-chrome-dev container image running in the background via local dev tools
  3. If that background image was built from a commit before `361f57eb` (the one that added the ./theme/useDarkModeStore MF expose), then the store simply doesn't exist in the running Chrome's remoteEntry — `useRemoteHook` would return ` { error: true, hookResult: null }` at runtime with no obvious explanation
  4. There was a later PR that triggered a new push to master, which built a fresh insights-chrome-dev image. Once the local tooling pulled the newer image, the store
  became available — making it look like export default was the fix, when really it was just the dev image catching up

 ### Why cache clearing didn't help:
Clearing podman caches and restarting the dev server wouldn't help if the latest available insights-chrome-dev image in quay still predated the store. The image had to be rebuilt from a newer commit first.

 ### This Change Itself
It's harmless and brings darkModeStore.ts in line with the other MF-exposed files (useSearch, useOramaSearch, etc.) which all use default exports. Note that openShiftIntercomStore.ts also lacks a default export — if we want full consistency, that file should get the same treatment.

### Future debugging
If this happens again in the future, it would be helpful if you check the version of chrome-dev running on your machine:

```
 To check the running version:
  # Find the running container
  podman ps | grep chrome
  
  # What digest is running locally?
  podman inspect <container-id> --format '{{.ImageDigest}}'
```

And you can resolve by just running: 
```
podman pull quay.io/redhat-user-workloads/hcc-platex-services-tenant/insights-chrome-dev:latest
```
Then kill/restart your local dev tools

eliranmal:
thanx for elaborating @florkbr !

ok, so all reasoning aside, how do we solve the core issue?  should i just wait for the next update from chrome runtime?  (and when will it happen?)

btw, IRT this:
>  useRemoteHook would return  { error: true, hookResult: null } at runtime with no obvious explanation

for what it's worth, this is not what i'm seeing in the app runtime - i just see `error: null` and `loading: true` forever.

florkbr:
@eliranmal 

That points to processManifest — the manifest fetch itself is hanging. This would happen if the Chrome container's manifest URL unreachable or timing out, not if the module is simply missing from it. If the module was missing, scalprum would throw and you'd see a different error.

  So the root cause might be simpler than what I thought: the Chrome dev container wasn't running or wasn't reachable when uhc-portal tried to load the remote hook. The manifest fetch just never completed — no error, no timeout, just an indefinitely pending promise.

eliranmal:
> So the root cause might be simpler than what I thought: the Chrome dev container wasn't running or wasn't reachable when uhc-portal tried to load the remote hook. The manifest fetch just never completed — no error, no timeout, just an indefinitely pending promise.

hmm that still doesn't add up, because a similar remote-hook request (for openshift-intercom) returned ok - hook-result was populated and the loading/error fields were falsy 🤔 

btw, both remote-hook requests were done side-by-side, at the same time.

PR #3595, opened 2026-07-07
Allow dark theme selection in Lightwell

marusak:
This is currently only in Preview mode. Lightwell does not have preview mode but we want to enable dark theme.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dark mode now activates automatically on Lightwell-related pages, even when the usual dark-mode settings are not enabled.
  * Dark mode settings continue to work as before on all other pages.

### DIFF
--- a/src/components/Header/Tools.tsx
+++ b/src/components/Header/Tools.tsx
@@ -15,7 +15,7 @@ import UserToggle from './UserToggle';
 import ToolbarToggle from './ToolbarToggle';
 import SettingsToggle, { SettingsToggleDropdownGroup } from './SettingsToggle';
 import cookie from 'js-cookie';
-import { ITLess, getSection } from '../../utils/common';
+import { ITLess, getSection, LIGHTWELL_PATH } from '../../utils/common';
 import { useIntl } from 'react-intl';
 import { useFlag } from '@unleash/proxy-client-react';
 import messages from '../../locales/Messages';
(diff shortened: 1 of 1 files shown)


florkbr:
@marusak chatted with @ochosi offline about this. There are a few concerns:

```
Your concern is valid. PR #3595 has this at module scope:

  const isLightwellRoute = window.location.pathname === LIGHTWELL_PATH || window.location.pathname.startsWith(`${LIGHTWELL_PATH}/`);

  This is evaluated once when the module loads. It's not reactive to navigation — it's a static snapshot of the URL at import time. So:

  - If a user lands on /lightwell, isLightwellRoute is true for the entire session. They see the dark mode toggle. They navigate to /insights/dashboard — the toggle is still visible because the module-level const doesn't re-evaluate. This is actually the opposite of your concern —
  the toggle persists everywhere, not just on Lightwell.
  - If a user lands on /insights/dashboard first, isLightwellRoute is false. They navigate to /lightwell — no dark mode toggle, because the const was captured at initial load.

  Both directions are broken. The behavior depends entirely on which page the user first loaded.

  **But even if this were made reactive (e.g., using useLocation inside the component), your core concern still stands: a user enables dark mode on /lightwell, navigates to another app where the platform.chrome.dark-mode flag is off, and is now stuck in dark mode with no toggle to
  switch back**.

  PR #3594 doesn't have this problem — it uses the feature flags as-is (useFlag('platform.chrome.dark-mode')), so the dark mode toggle is governed globally by Unleash, not per-route. The toggle is either available everywhere or nowhere.

  Bottom line: #3595 has a real regression risk. #3594 is the safer approach. If dark mode needs to be Lightwell-only, it should be controlled via Unleash targeting (e.g., enable the flag for Lightwell users/contexts) rather than a client-side route check.
```

**TL;DR a user enables dark mode on /lightwell, navigates to another app where the platform.chrome.dark-mode flag is off, and is now stuck in dark mode with no toggle to switch back**

**For that reason I think we should go with #3594 and revisit dark mode later.**

florkbr:
@marusak can this be closed? Thx

karelhala:
Another PR adds option to turn on/off specific tools https://github.com/RedHatInsights/insights-chrome/pull/3644

PR #3600, opened 2026-07-09
fix: centralize bundle visibility evaluation to prevent Kessel API storm

riccardo-forina:
## Summary

- Centralizes bundle and service tile visibility evaluation into a single initializer (`useInitVisibleBundles`) that runs once after feature flags are ready
- Stores pre-evaluated results in shared Jotai atoms (`visibleBundlesAtom`, `visibleServiceTilesAtom`)
- All consumers (`useAllLinks`, `useAllServices`, `useNavigation`, `useAppFilter`, `AllServices`, `useFavoritedServices`) now read pre-evaluated data instead of independently calling `evaluateVisibility`
- Removes hidden nav items from the tree (not just marking `isHidden`) so downstream consumers like the All Services redesign page render correctly
- Adds `filterHiddenItems` with unit tests

### Before
Multiple hooks independently fetched bundle data and ran `evaluateVisibility`, each triggering `loosePermissionsKessel` API calls. Combined with re-renders from Unleash flag changes and Scalprum initialization, this produced 16+ identical POST requests per page load — breaching Akamai's rate limit (5 non-GET hits/sec over 5 seconds) and causing "Access Denied" errors for users.

### After
One initializer evaluates visibility once. Zero caching layers, zero dedup — the architecture prevents duplicate evaluations structurally.

**16 → 2 Kessel API calls from Chrome per page load.**

## Test plan

- [x] `npm test` — 79 suites, 687 tests pass
- [x] `npm run lint` — no new errors
- [ ] Load homepage — verify sidebar navigation renders correctly
- [ ] Verify All Services dropdown shows correct permission-gated items
- [ ] Verify All Services page (redesign) shows only V2 links for V2 users
- [ ] Verify favorited services still work
- [ ] Verify Network tab shows only 2 Kessel calls from chrome (checkself + checkselfbulk)
- [ ] Navigate between bundles — no re-evaluation storms
- [ ] Test with a restricted user — hidden nav items should not appear

## Summary by CodeRabbit

* **New Features**
  * Navigation, All Services, and filter views now build their contents from “visible” bundles/tiles state, so users see only relevant items.
* **Bug Fixes**
  * Hidden entries are filtered consistently across menus, nested groups, and service sections.
  * Active navigation state updates more reliably using the updated visible content flow.
* **Chores**
  * Updated unit and Cypress tests to seed visible bundles/tiles state directly (reducing network mocking and improving determinism).
  * Added coverage for visibility filtering and initialization behavior.

### DIFF
--- a/src/components/AppFilter/useAppFilter.ts
+++ b/src/components/AppFilter/useAppFilter.ts
@@ -1,11 +1,8 @@
-import axios from 'axios';
 import { useEffect, useState } from 'react';
 import { BundleNavigation, ChromeModule, NavItem } from '../../@types/types';
-import { getChromeStaticPathname } from '../../utils/common';
-import { evaluateVisibility } from '../../utils/isNavItemVisible';
 import { useAtomValue } from 'jotai';
 import { chromeModulesAtom } from '../../state/atoms/chromeModuleAtom';
-import { navigationAtom } from '../../state/atoms/navigationAtom';
+import { useVisibleBundles } from '../../state/atoms/visibleBundlesAtom';
 
 export type AppFilterBucket = {
   id: string;
@@ -96,10 +93,10 @@ const useAppFilter = () => {
       },
     },
   });
-  const existingSchemas = useAtomValue(navigationAtom);
+  const visibleBundles = useVisibleBundles();
   const modules = useAtomValue(chromeModulesAtom);
 
-  const handleBundleData = async ({ data: { id, navItems, title } }: { data: BundleNavigation }) => {
+  const handleBundleData = ({ id, navItems, title }: BundleNavigation) => {
     const links: (NavItem & {
       costLinks: NavItem[];
       subscriptionsLinks: NavItem[];
@@ -119,28 +116,22 @@ const useAppFilter = () => {
       cost: [],
       subs: [],
     };
-    const promises = links.map(async ({ costLinks, subscriptionsLinks, ...rest }) => {
-      const nextIndex = bundleLinks.length;
+    links.forEach(({ costLinks, subscriptionsLinks, ...rest }) => {
       if (costLinks.length > 0) {
-        extraLinks.cost = await costLinks.filter(evaluateVisibility);
+        extraLinks.cost.push(...costLinks.filter((item) => !item.isHidden));
       }
 
       if (subscriptionsLinks.length > 0) {
-        extraLinks.subs = await subscriptionsLinks.filter(evaluateVisibility);
+        extraLinks.subs.push(...subscriptionsLinks.filter((item) => !item.isHidden));
       }
       if (rest.filterable !== true && (subscriptionsLinks.length > 0 || costLinks.length > 0)) {
         return;
       }
 
-      /**
-       * We have to create a placeholder for the link item, in order to preserver the links order
-       */
-      bundleLinks.push({ ...rest, isHidden: true });
-      const link = await evaluateVisibility(rest);
-
-      bundleLinks[nextIndex] = link;
+      if (!rest.isHidden) {
+        bundleLinks.push(rest);
+      }
     });
-    await Promise.all(promises);
 
     setState((prev) => ({
       ...prev,
@@ -165,27 +156,18 @@ const useAppFilter = () => {
   };
 
   useEffect(() => {
-    if (state.isOpen && !state.isLoading && !state.isLoaded) {
+    if (state.isOpen && !state.isLoading && !state.isLoaded && visibleBundles.length > 0) {
       setState((prev) => ({
         ...prev,
         isLoading: true,
       }));
-      const bundles = requiredBundles.filter((app) => !Object.keys(existingSchemas).includes(app));
-      bundles.map((fragment) =>
-        axios
-          .get<BundleNavigation>(`${getChromeStaticPathname('navigation')}/${fragment}-navigation.json?ts=${Date.now()}`)
-          // fallback static CSC for EE env
-          .catch(() => {
-            return axios.get<BundleNavigation>(`$/config/chrome/${fragment}-navigation.json?ts=${Date.now()}`);
-          })
-          .then(handleBundleData)
-          .then(() => Object.values(existingSchemas).map((data) => handleBundleData({ data } as { data: BundleNavigation })))
-          .catch((err) => {
-            console.error('Unable to load appfilter bundle', err, fragment);
-          })
-      );
+      visibleBundles.forEach(handleBundleData);
+      setState((prev) => ({
+        ...prev,
+        isLoading: false,
+      }));
     }
-  }, [state.isOpen]);
+  }, [state.isOpen, visibleBundles, modules]);
 
   const setIsOpen = (isOpen: boolean) => {
     setState((prev) => ({
--- a/src/hooks/useAllServices.ts
+++ b/src/hooks/useAllServices.ts
@@ -1,24 +1,11 @@
-import axios, { AxiosResponse } from 'axios';
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { NavItem } from '../@types/types';
 import { AllServicesGroup, AllServicesLink, AllServicesSection, isAllServicesGroup, isAllServicesLink } from '../components/AllServices/allServicesLinks';
-import { getChromeStaticPathname } from '../utils/common';
-import { evaluateVisibility } from '../utils/isNavItemVisible';
-import useFeoConfig from './useFeoConfig';
+import { useVisibleServiceTiles } from '../state/atoms/visibleBundlesAtom';
 
 export type AvailableLinks = {
   [key: string]: NavItem;
 };
-
-const allServicesFetchCache: {
-  [qeury: string]: Promise<
-    AxiosResponse<
-      (Omit<AllServicesSection, 'links'> & {
-        links: (AllServicesLink | AllServicesGroup)[];
-      })[]
-    >
-  >;
-} = {};
 
 const matchStrings = (value = '', searchTerm: string): boolean => {
   // convert strings to lowercase and remove any white spaces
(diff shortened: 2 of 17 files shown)


src/state/atoms/visibleBundlesAtom.ts
```diff
@@ -0,0 +1,155 @@
...
+      .catch((error) => {
+        console.error('Failed to fetch and evaluate service tiles', error);
+        if (!cancelled) {
+          setTilesError(true);
+          setTilesReady(true);
+        }
+      });
+
+    return () => {
+      cancelled = true;
```
karelhala, line 134: **suggestion:** The cleanup only prevents stale state writes but doesn't cancel in-flight HTTP requests. The old `useAllServices` code used `AbortController` to actually abort network requests on unmount/re-render. Consider adding one here — otherwise if `feoGenerated` toggles or the component remounts, zombie requests continue consuming bandwidth.

```suggestion
    const controller = new AbortController();
    return () => {
      cancelled = true;
      controller.abort();
    };
```

You'd then need to pass `{ signal: controller.signal }` to the axios call on line 115 and thread it through `fetchNavigationFiles` (which already uses axios internally).

src/state/atoms/visibleBundlesAtom.ts
```diff
@@ -0,0 +1,155 @@
...
+          setBundles(result);
+          setBundlesReady(true);
+        }
+      })
+      .catch((error) => {
+        console.error('Failed to fetch and evaluate bundles', error);
+        if (!cancelled) {
+          setBundlesError(true);
+          setBundlesReady(true);
+        }
```
karelhala: **suggestion:** If `fetchNavigationFiles` fails, `visibleBundlesAtom` stays `[]` permanently. Downstream consumers that guard on `visibleBundles.length > 0` (`useNavigation`, `useAppFilter`, `AllServices.tsx`) will silently stay stuck in a loading/empty state with no error UI — the user sees a blank page with no feedback.

The old code handled this per-consumer (e.g., `useNavigation` called `setNoNav(true)` in its catch). Now that error handling is centralized here, consumers need a way to react.

Consider either:
- Having consumers also read `visibleBundlesErrorAtom` and render an error state, or
- Setting `setBundles` to a sentinel (e.g. empty array is fine but also set a ready flag like you do for tiles) so consumers can distinguish "not loaded yet" from "loaded but empty/failed".

src/hooks/useAllServices.ts
```diff
@@ -1,25 +1,11 @@
...
+import { useMemo, useState } from 'react';
 import { AllServicesGroup, AllServicesLink, AllServicesSection, isAllServicesGroup, isAllServicesLink } from '../components/AllServices/allServicesLinks';
-import { getChromeStaticPathname } from '../utils/common';
-import { evaluateVisibility } from '../utils/isNavItemVisible';
-import useFeoConfig from './useFeoConfig';
+import { useVisibleServiceTiles } from '../state/atoms/visibleBundlesAtom';
 
 export type AvailableLinks = {
-  [key: string]: NavItem;
+  [key: string]: unknown;
```
karelhala: **nit:** This type is exported — weakening the value from `NavItem` to `unknown` could break any external consumer that imports `AvailableLinks` and relies on the value type. Quick grep shows it's not used elsewhere in this repo, but if any consuming app imports it, they'd get a type error. Consider keeping `NavItem` or removing the export if it's truly unused.

src/state/atoms/visibleBundlesAtom.test.ts
```diff
@@ -0,0 +1,191 @@
+import { evaluateServiceTilesVisibility, filterHiddenItems } from './visibleBundlesAtom';
+import { NavItem } from '../../@types/types';
+import { AllServicesSection } from '../../components/AllServices/allServicesLinks';
```
karelhala, line 3: **nit:** `filterHiddenItems` is well tested here (7 cases, good coverage). However, `useInitVisibleBundles` — the most critical new code — has no direct test coverage. Its error paths (fetch failure → atoms stay empty), cancellation behavior (cleanup on unmount), and the guard on `flagsReady`/`flagsError` are all untested. Consider adding a few `renderHook` tests for the initializer, especially the error → consumer-stuck scenario flagged above.

karelhala (commented):
Solid architectural change — centralizing visibility evaluation eliminates the API storm by construction rather than by caching. The PR also incidentally fixes three pre-existing bugs:

1. **`getNavLinks` `concat` bug** (`useAllLinks.ts`): `links.concat(...)` returns a new array without mutating `links`, so nested nav items were silently dropped. Fixed with `links.push(...)`.
2. **Async predicate in `Array.filter`** (`useAppFilter.ts`): `costLinks.filter(evaluateVisibility)` — `evaluateVisibility` is async, returns a Promise (always truthy), so nothing was ever filtered. Fixed by checking `isHidden` on pre-evaluated data.
3. **`test.only` left in test file** (`useNavigation.test.js`): Was skipping sibling tests in CI.

Four items worth addressing below (two suggestions, two nits).

src/state/atoms/visibleBundlesAtom.ts
riccardo-forina: Added `AbortController` for the service tiles fetch — the axios call now receives `{ signal: controller.signal }` and the cleanup aborts it. For the bundles fetch, `fetchNavigationFiles` has its own internal caching layer that doesn't accept a signal, so we keep the `cancelled` flag there. Threading a signal through `fetchNavigationFiles` is a deeper change better suited for a follow-up.

src/state/atoms/visibleBundlesAtom.ts
riccardo-forina: Addressed. Added `visibleBundlesReadyAtom` and `visibleBundlesErrorAtom` (mirroring the service tiles pattern). `useNavigation` now reads `useVisibleBundlesError()` and sets `noNav: true` on failure so the user sees the no-nav state instead of a blank page. Error atoms are also reset at the start of each fetch cycle so stale errors don't persist across retries.

src/hooks/useAllServices.ts
riccardo-forina: Fixed — restored the value type to `NavItem`.

src/state/atoms/visibleBundlesAtom.test.ts
riccardo-forina: Agree this would be valuable. The hook needs a full provider setup (Jotai store, Unleash mock, axios mock, cleanup verification) which makes it a heavier lift. Deferring to a follow-up to keep this PR focused on the structural fix. The critical paths (evaluation, filtering, error propagation) are covered by the `filterHiddenItems` and `evaluateServiceTilesVisibility` unit tests plus the Cypress integration tests.

riccardo-forina:
@coderabbit review

src/components/AppFilter/useAppFilter.ts
```diff
@@ -165,27 +156,14 @@ const useAppFilter = () => {
...
-            return axios.get<BundleNavigation>(`$/config/chrome/${fragment}-navigation.json?ts=${Date.now()}`);
-          })
-          .then(handleBundleData)
-          .then(() => Object.values(existingSchemas).map((data) => handleBundleData({ data } as { data: BundleNavigation })))
-          .catch((err) => {
-            console.error('Unable to load appfilter bundle', err, fragment);
-          })
-      );
+      visibleBundles.forEach(handleBundleData);
     }
```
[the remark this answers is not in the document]
riccardo-forina: Already addressed — `setState({ isLoading: false })` is now called after `visibleBundles.forEach(handleBundleData)` in the current code.

src/layouts/AllServices.test.tsx
```diff
@@ -36,9 +36,8 @@ jest.mock('../hooks/useAllServices', () => ({
   }),
 }));
 
-jest.mock('../hooks/useAllLinks', () => ({
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  fetchBundles: jest.fn<() => Promise<any[]>>().mockResolvedValue([]),
+jest.mock('../state/atoms/visibleBundlesAtom', () => ({
+  useVisibleBundles: jest.fn().mockReturnValue([]),
```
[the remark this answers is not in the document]
riccardo-forina: will do in a future PR

karelhala:
@coderabbitai full review

karelhala:
## Navigation Comparison: `stage.foo:1337` (PR) vs `console.stage.redhat.com` (production)

### Sidebar Navigation — All 5 bundles match ✅

| Bundle | Result |
|--------|--------|
| `/ansible` | Identical |
| `/openshift` | Identical |
| `/insights` | Identical |
| `/settings` | Identical |
| `/iam` | Identical |

The left sidebar navigation is fully consistent across both environments for all tested bundles.

<details>
<summary><b>Ansible Navigation Screenshots</b></summary>

| stage.foo (PR) | console.stage (production) |
|---|---|
| stage-foo-ansible | console-stage-ansible |

</details>

<details>
<summary><b>OpenShift Navigation Screenshots</b></summary>

| stage.foo (PR) | console.stage (production) |
|---|---|
| stage-foo-openshift | console-stage-openshift |

</details>

<details>
<summary><b>Insights (RHEL) Navigation Screenshots</b></summary>

| stage.foo (PR) | console.stage (production) |
|---|---|
| stage-foo-insights | console-stage-insights |

</details>

<details>
<summary><b>Settings Navigation Screenshots</b></summary>

| stage.foo (PR) | console.stage (production) |
|---|---|
| stage-foo-settings | console-stage-settings |

</details>

<details>
<summary><b>IAM Navigation Screenshots</b></summary>

| stage.foo (PR) | console.stage (production) |
|---|---|
| stage-foo-iam | console-stage-iam |

</details>

---

### `/allservices` Page — Missing Items ⚠️

The following **13 items** are visible on `console.stage.redhat.com/allservices` (production) but **missing** from `stage.foo.redhat.com:1337/allservices` (PR build):

**OpenShift section (11 items):**

| # | Item | Sub-section |
|---|------|-------------|
| 1 | 60-Day Product Trial \| OpenShift AI | OpenShift AI > Products |
| 2 | ACS Instances | Advanced Cluster Security > Products |
| 3 | Advanced Cluster Management | Products |
| 4 | Cluster List | Cluster Management |
| 5 | Developer Sandbox | Resources |
| 6 | Developer Sandbox \| OpenShift AI | OpenShift AI > Products |
| 7 | Documentation (ACS) | Advanced Cluster Security > Products |
| 8 | Getting Started | Advanced Cluster Security > Products |
| 9 | OpenShift Virtualization | Products |
| 10 | Operators | Products |
| 11 | Overview (ACS) | Advanced Cluster Security > Products |

**IAM section (2 items):**

| # | Item | Sub-section |
|---|------|-------------|
| 12 | Service Accounts | *(top-level)* |
| 13 | Workspaces | User Access |

Most of the missing OpenShift items are under "Products" sub-sections (ACS, OpenShift AI) and external-link entries. These appear to be getting filtered out by the centralized Kessel API visibility evaluation where they shouldn't be.

<details>
<summary><b>All Services Page Screenshots</b></summary>

**stage.foo (PR):**
allservices


**console.stage (production):**
allservices


</details>

riccardo-forina:
## Re: Missing AllServices items

The 13 missing items are **not a filtering bug** — they are a data source difference.

The old code had an inconsistency: `fetchBundles()` in `useAllLinks.ts` hardcoded `feoGenerated = false`:

```typescript
export const fetchBundles = async (feoGenerated = false) => { ... }
```

`AllServices.tsx` called it without arguments, so the AllServices redesign page **always** fetched from the legacy `*-navigation.json` files. Meanwhile, every other consumer (sidebar, useAllLinks, useAllServices, useNavigation) used `useFeoConfig()` and fetched from FEO when the flag was on.

This PR unified all consumers to use the same FEO data source (`bundles-generated.json`). The 11 missing OpenShift items (ACS Instances, Developer Sandbox, Cluster List, etc.) simply do not exist in the FEO-generated bundles — they have not been migrated to FEO yet.

The 2 missing IAM items (Service Accounts, Workspaces) do exist in the FEO data but are gated by permissions (`featureFlag("platform.rbac.workspaces")` and `scope("api.iam.service_accounts")`), so their visibility depends on the user's flags and scopes.

**TL;DR:** the old AllServices page was accidentally bypassing FEO by hardcoding `false`. This PR makes it consistent with everything else. The missing items are a FEO data migration gap, not a Chrome code issue.

karelhala (approved):

PR #3601, opened 2026-07-09
test: add e2e tests for theming

OlhaTmlk:
### Description

Add Playwright e2e tests for the three theme settings panels: 
- color scheme (light/dark/system), 
- contrast mode (default/high contrast/glass), 
- theme toggle (default/project felt) 

Tests cover switching between options, verifying localStorage persistence across page reloads, and mutual exclusivity of contrast modes.

Feature flags are mocked via Unleash API interception to ensure tests are deterministic regardless of environment flag state.

[RHCLOUD-49097](https://redhat.atlassian.net/browse/RHCLOUD-49097)

[RHCLOUD-49097]: https://redhat.atlassian.net/browse/RHCLOUD-49097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for theme toggles, Color Scheme (Light/Dark/System), and Contrast Mode (Default/High Contrast/Glass/System).
  * Verified the settings UI renders the expected options and that selections update the correct active/pressed state.
  * Confirmed the correct theme styling is applied and that selections persist in browser storage across reloads (including switching and mutual exclusivity for contrast modes).
  * Introduced test-side feature flag mocking to reliably exercise the relevant UI paths.

### DIFF
--- a/playwright/e2e/theme-toggle.spec.ts
+++ b/playwright/e2e/theme-toggle.spec.ts
@@ -1,11 +1,13 @@
 import { expect, test } from '../setup/test-setup';
 import { ChromeTopbar } from './pages/chrome-topbar';
+import { mockFeatureFlags } from '../helpers/feature-flags';
 
 const FELT_THEME_CLASS = 'pf-v6-theme-felt';
 const FELT_STORAGE_KEY = 'chrome:felt-theme';
 
 test.describe('Theme Toggle — Default / Project Felt', () => {
   test.beforeEach(async ({ page }) => {
+    await mockFeatureFlags(page, ['platform.chrome.felt-theme']);
     await page.goto('/');
     await page.evaluate((key) => localStorage.removeItem(key), FELT_STORAGE_KEY);
     await page.reload();
@@ -39,6 +41,20 @@ test.describe('Theme Toggle — Default / Project Felt', () => {
     expect(stored).toBe('true');
   });
 
+  test('should switch from Felt back to Default', async ({ page }) => {
+    await page.evaluate((key) => localStorage.setItem(key, 'true'), FELT_STORAGE_KEY);
+    await page.reload();
+    const topbar = new ChromeTopbar(page);
+    await topbar.openSettings();
+    await expect(page.locator('html')).toHaveClass(new RegExp(FELT_THEME_CLASS));
+
+    await page.locator('#theme-default').click();
+    await expect(page.locator('html')).not.toHaveClass(new RegExp(FELT_THEME_CLASS));
+
+    const stored = await page.evaluate((key) => localStorage.getItem(key), FELT_STORAGE_KEY);
+    expect(stored).toBe('false');
+  });
+
   test('should persist Felt theme after page reload', async ({ page }) => {
     const topbar = new ChromeTopbar(page);
     await topbar.openSettings();
@@ -46,6 +62,7 @@ test.describe('Theme Toggle — Default / Project Felt', () => {
     await expect(page.locator('html')).toHaveClass(new RegExp(FELT_THEME_CLASS));
 
     await page.reload();
+    await topbar.openSettings();
 
     await expect(page.locator('html')).toHaveClass(new RegExp(FELT_THEME_CLASS));
   });
--- a/playwright/e2e/theme-toggle.spec.ts
+++ b/playwright/e2e/theme-toggle.spec.ts
@@ -1,0 +1,52 @@
+import { expect, test } from '../setup/test-setup';
+import { ChromeTopbar } from './pages/chrome-topbar';
+
+const FELT_THEME_CLASS = 'pf-v6-theme-felt';
+const FELT_STORAGE_KEY = 'chrome:felt-theme';
+
+test.describe('Theme Toggle — Default / Project Felt', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/');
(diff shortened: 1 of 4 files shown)


karelhala (commented):
## Missing feature flag mock in `theme-toggle.spec.ts`

The Theme section in `Tools.tsx` is gated behind `useFlag('platform.chrome.felt-theme')` ([Tools.tsx:112](https://github.com/RedHatInsights/insights-chrome/blob/master/src/components/Header/Tools.tsx#L112), [line 149](https://github.com/RedHatInsights/insights-chrome/blob/master/src/components/Header/Tools.tsx#L149): `isHidden: !isFeltThemeEnabled`).

The other two test files (`color-scheme.spec.ts`, `contrast-mode.spec.ts`) correctly mock their required flags via Unleash API route interception, but `theme-toggle.spec.ts` has no flag mocking. If the stage environment has `platform.chrome.felt-theme` disabled, `#theme-default` and `#theme-felt` won't render and all assertions will fail with element-not-found errors.

**Fix:** Add the same Unleash route interception pattern to the `beforeEach`:

```typescript
const REQUIRED_FLAGS = ['platform.chrome.felt-theme'];

test.beforeEach(async ({ page }) => {
  await page.route('**/api/featureflags/v0**', async (route) => {
    let toggles: object[] = [];
    try {
      const response = await route.fetch();
      toggles = ((await response.json()) as { toggles?: object[] }).toggles ?? [];
    } catch {
      // noop
    }
    const filtered = toggles.filter((t: { name?: string }) => !REQUIRED_FLAGS.includes(t.name ?? ''));
    for (const name of REQUIRED_FLAGS) {
      filtered.push({ name, enabled: true, impressionData: false, variant: { name: 'disabled', enabled: false } });
    }
    await route.fulfill({ json: { toggles: filtered } });
  });
  await page.goto('/');
  await page.evaluate((key) => localStorage.removeItem(key), FELT_STORAGE_KEY);
  await page.reload();
});
```

Optional improvement: the flag mock block is now duplicated three times — consider extracting it into a shared helper in `playwright/helpers/feature-flags.ts`.

playwright/helpers/feature-flags.ts
```diff
@@ -0,0 +1,18 @@
...
+export async function mockFeatureFlags(page: Page, flags: string[]): Promise<void> {
+  await page.route('**/api/featureflags/v0**', async (route) => {
+    let toggles: object[] = [];
+    try {
+      const response = await route.fetch();
+      toggles = ((await response.json()) as { toggles?: object[] }).toggles ?? [];
+    } catch {
+      // noop
+    }
+    const filtered = toggles.filter((t: { name?: string }) => !flags.includes(t.name ?? ''));
```
karelhala: Nit: `toggles` is typed as `object[]` but the filter callback asserts `t` as `{ name?: string }`. This type assertion is technically unsafe — `object` doesn't guarantee a `name` property.

Consider defining a minimal interface:

```typescript
interface FeatureToggle {
  name: string;
  enabled: boolean;
  impressionData: boolean;
  variant: { name: string; enabled: boolean };
}
```

Then use `FeatureToggle[]` instead of `object[]` on line 5 and 8. Not blocking — the current code works fine at runtime since the Unleash API does return `name` on every toggle.

playwright/e2e/theme-toggle.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const stored = await page.evaluate((key) => localStorage.getItem(key), FELT_STORAGE_KEY);
+    expect(stored).toBe('true');
+  });
+
+  test('should persist Felt theme after page reload', async ({ page }) => {
+    const topbar = new ChromeTopbar(page);
+    await topbar.openSettings();
+    await page.locator('#theme-felt').click();
+    await expect(page.locator('html')).toHaveClass(new RegExp(FELT_THEME_CLASS));
```
karelhala, line 62: Consider adding a test for switching **back** from Felt to Default — verifying the class is removed and localStorage is updated. The color-scheme spec has a "toggle from Dark to Light" test covering this pattern, but the theme-toggle spec is missing it.

Something like:
```typescript
test('should switch from Felt back to Default', async ({ page }) => {
  await page.evaluate((key) => localStorage.setItem(key, 'true'), FELT_STORAGE_KEY);
  await page.reload();
  const topbar = new ChromeTopbar(page);
  await topbar.openSettings();
  await expect(page.locator('html')).toHaveClass(new RegExp(FELT_THEME_CLASS));

  await page.locator('#theme-default').click();
  await expect(page.locator('html')).not.toHaveClass(new RegExp(FELT_THEME_CLASS));
});
```

Not blocking — the Default state test implicitly covers the class-absent case, just not the transition.

karelhala (commented):
## Review

Clean, well-structured E2E tests for the theming features. Good coverage of switching, persistence, and mutual exclusivity.

Strengths:
- Feature flag mocking via route interception is a solid deterministic approach
- Using `aria-pressed` for state assertions (proper accessible selectors)
- localStorage verification alongside visual class assertions gives confidence at two levels
- Mutual exclusivity test for Glass/High Contrast is valuable
- All CI checks pass

Two minor suggestions below — neither is blocking.

karelhala (approved):

PR #3602, opened 2026-07-10
feat(analytics): add enriched user properties to Amplitude autocapture

catastrophe-brandon:
## Summary

Enhances Amplitude autocapture functionality to include enriched user properties for better analytics segmentation and analysis.

Note: Fixes for flaky tests need to be merged first. See #3603

## Required Properties
- ✅ **internal** (boolean) - User's internal status, matches Segment-sourced events

## Stretch Goal Properties
- ✅ **isBeta** (boolean) - Preview mode status
- ✅ **isOrgAdmin** (boolean) - Organization admin status  
- ✅ **org_id** (string) - Organization identifier

## Additional Context Properties
- Organization: account_id, account_number, organization_name
- User: locale, email_domain
- Application: current_bundle, current_app
- Entitlements: entitlement_* (boolean) - all entitlements with trial flags (40 total properties)

### User Properties Collection
- Built from ChromeAuthContext user data
- Property names match Segment analytics conventions (camelCase for booleans, snake_case for IDs)
- Filtered out undefined values

### Amplitude Integration
- Used Amplitude Identify() API with .set() for each property
- Wait for amplitude.init() promise to complete before calling identify()
- Properties sent in $set operation of $identify event

### Unit Tests
- ✅ Updated `useAmplitude.test.tsx` to verify Identify() called with enriched properties
- ✅ All 12 unit tests passing
- ✅ Validates each required and stretch property is set correctly

### E2E Tests  
- ✅ New `playwright/e2e/amplitude-autocapture.spec.ts`
- ✅ Intercepts FEO config to inject test Amplitude API keys
- ✅ Intercepts Unleash feature flags to enable autocapture
- ✅ Captures and decompresses gzipped Amplitude requests (using gunzipSync)
- ✅ Uses page.waitForRequest() to wait for actual Amplitude event (no blind timeouts)
- ✅ Validates required `internal` property and stretch goals
- ✅ Verifies all entitlement properties included
- ✅ Tests feature flag gating (negative test)

### Test Results
```
2 passed (32.2s)
```

## Technical Notes

- Amplitude SDK batches multiple events in a single request
- E2E test searches for the specific $identify event containing our custom properties
- Properties appear in user_properties.$set object (Amplitude's operation-based structure)
- Init timing is critical - identify() must be called after init() completes

## Related Work
- Aligns with existing Segment analytics property conventions
- Complements visual labeling autocapture functionality

## Summary by CodeRabbit

* **New Features**
  * Enhanced Amplitude autocapture with enriched context, including account, organization, locale, application, preview state, module context, and entitlement properties.
  * Added entitlement/trial indicators and improved autocapture initialization as app context changes.

* **Documentation**
  * Added guidance for enriched autocapture user properties, tenant app coverage/behavior, privacy constraints, and example analysis queries.

* **Tests**
  * Added a Playwright E2E suite validating enriched identify payloads and enabled/disabled autocapture behavior.
  * Improved reliability of tooltip and auth-related UI visibility assertions using shared timeout constants.

### DIFF
--- a/docs/amplitude-autocapture-tenant-app-coverage.md
+++ b/docs/amplitude-autocapture-tenant-app-coverage.md
@@ -35,8 +35,9 @@ The Amplitude autocapture plugin uses **browser-level event listeners** on the `
 
 1. ✅ **It captures events from the ENTIRE DOM** - not just Chrome shell elements
 2. ✅ **It captures events from dynamically loaded Module Federation apps** (tenant applications)
-3. ✅ **It captures events from iframes** (if same-origin)
-4. ✅ **User properties set via `identify()` persist across ALL captured events**
+3. ✅ **User properties set via `identify()` persist across ALL captured events**
+
+**Note on iframes:** Events from within iframes (even same-origin) are NOT automatically captured as they have separate document contexts. Module Federation apps render in the main document, not iframes, so this limitation doesn't affect tenant applications.
 
 ### What Gets Captured Automatically
 
@@ -196,6 +197,7 @@ Some enriched properties are **reactive** and update automatically as the user n
 
 | Scenario | Captured? | Notes |
 |----------|-----------|-------|
+| **Same-origin iframes** | ❌ NO | Separate document context; events don't bubble to parent |
 | **Cross-origin iframes** | ❌ NO | Browser security restrictions |
 | **Events before SDK init** | ❌ NO | Autocapture only starts after Chrome initializes |
 | **Custom events** | ❌ NO | Tenant apps must manually track via Segment/Amplitude |
@@ -246,7 +248,7 @@ The enriched properties (`internal`, `isOrgAdmin`, etc.) are **automatically inc
 ### Cost Management Team
 
 **"How many org admins created a budget this week?"**
-```
+```text
 Event: [Amplitude] Element Clicked
   WHERE [Amplitude] Element Text = "Create Budget"
   AND isOrgAdmin = true
@@ -254,7 +256,7 @@ Event: [Amplitude] Element Clicked
 ```
 
 **"Do trial users interact with the cost optimization recommendations?"**
-```
+```text
 Event: [Amplitude] Element Clicked
   WHERE [Amplitude] Page URL CONTAINS "/cost-management"
   AND entitlement_cost_management_trial = true
@@ -264,7 +266,7 @@ Event: [Amplitude] Element Clicked
 ### Ansible Team
 
 **"Are beta users clicking the new playbook builder?"**
-```
+```text
 Event: [Amplitude] Element Clicked
   WHERE [Amplitude] Element Text = "Create Playbook"
   AND isBeta = true
@@ -274,7 +276,7 @@ Event: [Amplitude] Element Clicked
 ### Subscriptions Team
 
 **"Which organizations are most engaged with subscription management?"**
-```
+```text
 Event: [Amplitude] Page Viewed
   WHERE [Amplitude] Page URL CONTAINS "/subscriptions"
   GROUP BY org_id, organization_name
@@ -321,19 +323,26 @@ Therefore, autocapture works seamlessly across all federated apps.
 
 ## Performance Impact
 
-### Zero Additional Load for Tenant Apps
+### No Additional Load for Tenant Apps
 
+Tenant applications benefit from Chrome-level initialization:
 - Amplitude SDK is loaded **once** by Chrome shell
 - Autocapture plugin is loaded **once** by Chrome shell  
 - Event listeners are registered **once** at the document level
-- Tenant apps add **zero additional analytics overhead**
+- Tenant apps incur no additional SDK loading or initialization overhead
 
 ### Minimal Runtime Impact
 
+The autocapture implementation uses efficient patterns to minimize performance impact:
 - Event listeners use **event delegation** (single listener per event type)
 - Event batching reduces network requests
 - Gzip compression on event payloads
-- Typical overhead: **< 50KB initial load, < 5KB per event batch**
+
+**Measured typical overhead** (varies by browser, network, and payload size):
+- Initial SDK/plugin load: ~50KB (gzipped)
+- Per event batch: ~2-5KB (varies with property count and batching)
+
+Note: Actual overhead depends on build configuration, browser environment, network conditions, and the number of properties included in each event.
 
 ---
 
--- a/docs/amplitude-autocapture-tenant-app-coverage.md
+++ b/docs/amplitude-autocapture-tenant-app-coverage.md
@@ -1,0 +1,409 @@
+# Amplitude Autocapture - Tenant Application Coverage
+
+## TL;DR: **YES** ✅ 
+
+**All tenant applications automatically benefit from these enriched user properties with ZERO additional work required.**
+
+---
+
+## How It Works
+
+### Chrome-Level Initialization
+
+The Amplitude autocapture SDK is initialized **once at the Chrome/platform level** in `src/analytics/useAmplitude.ts`:
+
+```typescript
+// Initialized in Chrome shell
+amplitude.add(autocapturePlugin());
+amplitude.init(autocaptureKeyToUse, userId, {
+  deviceId: deviceId,
+  defaultTracking: {
+    sessions: true,      // ✅ Tracks all page sessions
+    pageViews: true,     // ✅ Tracks all page navigations
+    formInteractions: true,  // ✅ Tracks all form interactions
+    fileDownloads: true, // ✅ Tracks all file downloads
+  },
+});
+
+// Set enriched user properties ONCE
+amplitude.identify(identifyEvent); // Contains all 40+ properties
+```
+
+### DOM-Level Event Capture
+
+The Amplitude autocapture plugin uses **browser-level event listeners** on the `document` or `window` object. This means:
+
+1. ✅ **It captures events from the ENTIRE DOM** - not just Chrome shell elements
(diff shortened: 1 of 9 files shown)


catastrophe-brandon:
@coderabbit-ai all critical findings have been resolved

catastrophe-brandon:
@coderabbit-ai resolve

src/analytics/useAmplitude.ts
```diff
@@ -186,7 +253,7 @@ function useAmplitude() {
```
karelhala, line 264: **Bug: `current_app`/`current_bundle` never update after init**

`activeModule` and `isPreview` added to deps, but `amplitudeSdkInitialized.current` guard (line 70) blocks re-entry. User navigates from Dashboard to Cost Management — `identify()` never re-fires. Properties stay frozen at first-visited app.

The tenant coverage doc (line 444) claims these update dynamically — code doesn't support that.

**Fix:** Extract identify logic into a separate `useEffect` that depends on `activeModule`/`isPreview` and calls `amplitude.identify()` directly (SDK already initialized). Keep init guard only for the `amplitude.init()` path.

src/analytics/useAmplitude.ts
```diff
@@ -20,6 +23,9 @@ function useAmplitude() {
   const forwardHandlerRef = useRef<((event: string, properties: Record<string, unknown>) => void) | null>(null);
   const activeModuleDefinition = useAtomValue(activeModuleDefinitionReadAtom);
   const chromeModules = useAtomValue(chromeModulesAtom);
```
karelhala, line 25: **Minor: `user` object reference in effect deps**

`user` from `useContext(ChromeAuthContext)` likely gets new reference each render. Effect fires every render, exits early via guard — harmless but wasteful. Consider comparing stable scalar values or memoizing upstream.

playwright/e2e/amplitude-autocapture.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+      // Navigate to the page
+      page.goto('/insights/dashboard'),
+    ]);
+
+    // Wait for event batching to complete
+    await page.waitForTimeout(EVENT_BATCH_DELAY);
+
+    // Verify that Amplitude requests were made
+    expect(amplitudeRequests.length).toBeGreaterThan(0);
+
```
karelhala, line 166: **Flakiness risk: hardcoded timeouts**

`EVENT_BATCH_DELAY = 2000` after already waiting via `waitForRequest` — likely unnecessary. For the negative test (line 330), `NEGATIVE_TEST_WAIT = 3000` is inherently racy. Consider `page.waitForLoadState('networkidle')` + shorter grace period instead.

playwright/e2e/release-gate/landing-page.spec.ts
```diff
@@ -18,11 +20,11 @@ test.describe('Landing page', () => {
     // Wait for the settings button to be present before interacting
     const settingsButton = page.getByRole('button', { name: 'Settings menu' });
```
karelhala, line 20: **Nit: inconsistent timeout pattern**

Local `TOOLTIP_TIMEOUT = 10000` while `auth.spec.ts` imports `UI_VISIBILITY_TIMEOUT` from shared constants. Same value, different source. Should use shared constant for consistency.

docs/amplitude-autocapture-tenant-app-coverage.md
```diff
@@ -0,0 +1,418 @@
+# Amplitude Autocapture - Tenant Application Coverage
```
karelhala, line 1: **Docs: consider trimming**

673 lines across two new markdown files. Suggest running `/caveman-compress` on these docs to cut the fluff while preserving technical substance. Also contains claims about dynamic property updates that don't match implementation (see comment on useAmplitude.ts line 253).

src/analytics/useAmplitude.ts
```diff
@@ -83,19 +89,80 @@ function useAmplitude() {
...
+              org_id: user.identity.internal?.org_id,
+
+              // Additional organization context
+              account_id: user.identity.internal?.account_id,
+              account_number: user.identity.account_number,
+              organization_name: user.identity.organization?.name,
+
+              // Additional user context
+              locale: user.identity.user?.locale,
+              email_domain: user.identity.user?.email ? user.identity.user.email.split('@')[1]?.toLowerCase() : undefined,
```
karelhala: **Question: `organization_name` sent to analytics**

Sending org display names to Amplitude. `email_domain` extraction (line 120) is good — verify `organization_name` also aligns with data classification policy. Org names can be sensitive depending on the customer.

karelhala (commented):
Review: 6 inline comments. Main concern is stale `current_app`/`current_bundle` — properties set once at init, never updated on navigation. Docs claim otherwise. Rest looks solid.

playwright/e2e/amplitude-autocapture.spec.ts
catastrophe-brandon: networkidle is an anti-pattern in our tests because there is often background activity in the application that means the idle event will never happen.

playwright/e2e/release-gate/auth.spec.ts
```diff
@@ -31,7 +31,7 @@ test.describe('Authentication', () => {
     await login(page);
 
     const userMenu = page.getByRole('button', { name: /User Avatar/ });
-    await expect(userMenu).toBeVisible();
+    await expect(userMenu).toBeVisible({ timeout: UI_VISIBILITY_TIMEOUT });
```
florkbr, line 34: nit: we could just set `expect: { timeout: 10_000 }` in `playwright.config.ts` instead? That applies globally and removes the need for per-assertion UI_VISIBILITY_TIMEOUT overrides. Not blocking at all and we can follow up if true.

src/analytics/useAmplitude.ts
```diff
@@ -83,19 +92,41 @@ function useAmplitude() {
...
+                } else {
+                  console.warn('No user properties to set for Amplitude autocapture');
+                }
+              })
+              .catch((error) => {
+                // Handle initialization promise rejection
+                amplitudeSdkInitialized.current = false;
+                console.error('Error during Amplitude SDK initialization promise', error);
+              });
           } catch (error) {
```
florkbr, line 177: nit: The three error handlers in initializeAmplitudeAutocapture all do the same thing (reset flag + log). Converting to async/await would collapse them into a single try/catch.

src/analytics/useAmplitude.ts
```diff
@@ -108,6 +139,77 @@ function useAmplitude() {
...
 
+  // Build enriched user properties - extracted to helper function for reuse
+  const buildUserProperties = function (): Record<string, unknown> {
+    if (!user) {
+      return {};
+    }
+
+    // Build enriched user properties from ChromeUser context
+    // Property names match Segment conventions (camelCase for booleans, snake_case for IDs)
+    const userProperties: Record<string, unknown> = {
```
florkbr: nit: it might be worth following up and tightening these typescript types a bit - though I realize that couples us more closely with the amplitude SDK.

src/analytics/useAmplitude.ts
justinorringer, line 1: The  changes here look good. Can you confirm a couple things?

```
  1. buildUserProperties uses stale closures — defined inside the hook body, captures activeModule, isPreview, user from render scope. Called from inside analytics.ready()
  → .then() chains, so by execution time those values may be stale. The updateAmplitudeUserProperties effect mitigates this for activeModule/isPreview, but the initial
  identify in initializeAmplitudeAutocapture could send stale data if the async chain resolves after a re-render.

  2. Entitlement key enumeration is unbounded — Object.entries(user.entitlements || {}) spreads N entitlement keys into user properties. If entitlements grow, this
  balloons the identify payload. Not a bug today, but worth noting for cardinality awareness on Amplitude's side.

  3. Missing useCallback/ref for buildUserProperties — duplicated identify logic appears in both initializeAmplitudeAutocapture and updateAmplitudeUserProperties. Both
  build an Identify object the same way. Could extract the "build + send identify" into one shared function. Minor DRY issue.

  4. updateAmplitudeUserProperties effect deps are incomplete — effect depends on [activeModule, isPreview] but the function also reads user. If user changes without
  activeModule/isPreview changing, properties won't update. The userAccountNumber dep in the autocapture init effect suggests awareness of this pattern but it wasn't
  applied here.
```

src/analytics/useAmplitude.test.tsx
```diff
@@ -334,4 +414,64 @@ describe('useAmplitude', () => {
     errorSpy.mockRestore();
     warnSpy.mockRestore();
   });
+
+  it('updates current_app and current_bundle properties on navigation', async () => {
```
florkbr, line 418: nit: A few negative paths in the new code are untested — notably user being null, initPromise.promise rejection, and missing entitlements/email. Would be good to add coverage for those since they're real scenarios during initial page load.

florkbr (commented):
Nits/non blocking comments only @catastrophe-brandon

src/analytics/useAmplitude.test.tsx
catastrophe-brandon: Added test coverage for this nit, take a look. The other nits may remain open for a future PR.

src/analytics/useAmplitude.ts
catastrophe-brandon: All good finds!

florkbr (approved):
I'd like to let this simmer in stage before promotion. Changes seem fine to me and should all be behind the `platform.chrome.analytics.amplitude.autocapture` feature flag.

PR #3605, opened 2026-07-14
fix(navigation): fix navigation after FEO and visibility evaluations

karelhala:
### Description

Fix navigation crashes caused by nav items with `expandable: true` but no `routes` array. This was introduced by recent navigation restructuring where some bundle nav items have the expandable flag set without providing a routes array, causing `TypeError: Cannot read properties of undefined (reading 'length')` in `findNavLeafPath` and `TypeError: Cannot read properties of undefined (reading 'map')` in `evaluateVisibility`.

Three files fixed, all addressing the same root cause:
- `isExpandableNav` type guard now verifies `routes` actually exists (not just `expandable`)
- `findNavLeafPath` adds a defensive early return for non-array input
- `evaluateVisibility` guards `.map()` calls on `routes` and `navItems` with `Array.isArray()`

---

#### After:

---

The `isExpandableNav` type guard was claiming `routes` is `Required` via the return type but only checked `!!item.expandable` — the type narrowing was a lie. The fix makes the runtime check match the type assertion. The `evaluateVisibility` fix is the same pattern: it used `result.routes!.map(...)` trusting the non-null assertion, which crashed when `expandable` was true without routes.

---

### Checklist

### AI disclosure
Assisted by: Claude Code (Opus)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation robustness by requiring valid route arrays for expandable items and by safely handling missing/malformed navigation data during leaf-path resolution and recursive visibility evaluation.

* **Tests**
  * Added a Playwright “Navigation rendering stability” release gate to verify key navigations and fail on navigation JS errors or bundle-evaluation fetch failures.
  * Expanded unit tests for expandable behavior, leaf-path matching, and recursive visibility edge cases.
  * Updated Playwright checks for Settings submenu titles and refined tooltip assertions; removed the dev-server overlay on page load to reduce rendering instability.

### DIFF
--- a/playwright/e2e/release-gate/favorite-services.spec.ts
+++ b/playwright/e2e/release-gate/favorite-services.spec.ts
@@ -6,11 +6,14 @@ test.describe('Favorite Services (E2E User Flow)', () => {
   });
 
   test('should favorite on the page and unfavorite from the header dropdown', async ({ page }) => {
+    test.slow();
     const serviceToTest = 'Groups';
     const quickstartIdSelector = '[data-quickstart-id="iam_user-access_groups"]';
 
     await page.goto('/allservices');
     await page.waitForLoadState('load');
+    // Wait for all services to render — bundle visibility evaluation can be slow in CI
+    await expect(page.getByRole('heading', { name: 'All Services', level: 2 })).toBeVisible({ timeout: 45000 });
     await page.getByText(serviceToTest).scrollIntoViewIfNeeded();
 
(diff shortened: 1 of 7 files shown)


Hyperkid123 (dismissed):

karelhala:
@coderabbitai full review

charlesmulder (dismissed):

karelhala:
@coderabbitai full review

Hyperkid123 (approved):

PR #3608, opened 2026-07-14
fix: prevent x-rh-frontend-origin header duplication on token renewals (RHCLOUD-40410)

catastrophe-brandon:
## Summary

Fixes RHCLOUD-40410

The `x-rh-frontend-origin: hcc` header was growing in size with each silent token renewal, resulting in headers like:
```
x-rh-frontend-origin: hcc, hcc, hcc, hcc, hcc, ...
```

The number of duplicates was proportional to how many token renewals had occurred in a long-lived browser tab.

## Root Cause

The `init()` function in `src/utils/iqeEnablement.ts` was being called on every silent token renewal from `onUserAuthenticated()`, causing layered monkey-patching of `XMLHttpRequest` and `fetch`. This resulted in:

1. **Duplicate headers**: `x-rh-frontend-origin: hcc, hcc, hcc, ...` (proportional to number of renewals)
2. **Performance degradation**: N-deep nested function wrappers on every request

Each call to `init()` would capture the current (already-patched) implementations and wrap them again, creating nested layers. For XHR requests, the native `setRequestHeader()` appends to existing header values (comma-joined), so N layers resulted in N duplicates of `hcc`.

## Solution

Made `init()` idempotent with a module-level `initialized` flag:

```typescript
let initialized = false;

export function init(chromeStore, authRef) {
  if (initialized) return;
  initialized = true;
  // ... rest of initialization
}
```

Now `init()` runs exactly once per page load. The wrapper functions close over `authRef` (a mutable ref), so they automatically see updated tokens without needing to re-wrap.

## Changes

- **src/utils/iqeEnablement.ts**: Added idempotency guard and test helper `_resetInitialization()`
- **src/utils/iqeEnablement.test.js**: Added comprehensive tests for idempotency

## Test Plan

- [x] All 704 existing tests pass
- [x] New idempotency tests pass (23 total in iqeEnablement.test.js)
- [x] Build completes successfully with no new errors/warnings
- [x] Manual verification: inspect network requests after multiple token renewals, confirm single `x-rh-frontend-origin: hcc` header

### Before
- Long-lived tabs accumulated duplicate headers and nested wrappers
- Backend logs showed growing header values (wasted bandwidth)
- Performance degradation over time (N function calls per request)

### After
- Clean, single-level API wrapping
- No duplicate headers
- No performance degradation
- Backend logs see consistent single-value headers

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate request instrumentation when initialization runs multiple times.
  * Improved stability during token renewals by ensuring request handling is set up only once and renewed tokens are used for subsequent requests.

* **Tests**
  * Added/extended coverage verifying initialization is idempotent for both `fetch` and `XMLHttpRequest`.
  * Added coverage confirming renewed tokens are applied after re-initialization.
  * Updated component tests to reset initialization state before each run for consistent results.


src/utils/iqeEnablement.test.ts
karelhala, line 1: May be worth migrating this completely to typescript while you're at it.

catastrophe-brandon: https://github.com/RedHatInsights/insights-chrome/pull/3612 has the refactoring. Because there's risk in re-working functional code, separate PR.

justinorringer (dismissed):
LGTM!

scripts/pre-push-check.sh
karelhala, line 1: I am not against using pre-push and commit hooks but these should be in seperate PR so we can focus on it more thoroughly. What I am afraid of is if there's an error with the original intent for this PR we will be forced to roll it back and would loose the git hooks.

catastrophe-brandon: removed

karelhala (approved):

karelhala:
I've pulled the PR locally and checked the code and everything looks good!

PR #3610, opened 2026-07-15
fix: Potential fix for kessel v2 users in notifications

jjaquish:
### Description

A potential fix for the notifications nav items not appearing for v2 users, as suggested by claude.

[RHCLOUD-49445](https://issues.redhat.com/browse/RHCLOUD-49445)

---

#### After:

---

---

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission checks by switching to the default workspace context.
  * Permission checks now fail safely when workspace information can’t be resolved.
  * Added caching to reduce repeated workspace lookups.

* **Tests**
  * Updated and expanded failure-path and retry/caching coverage for workspace-scoped permission checks.

[RHCLOUD-49445]: https://redhat.atlassian.net/browse/RHCLOUD-49445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/VisibilitySingleton.test.ts
+++ b/src/utils/VisibilitySingleton.test.ts
@@ -316,18 +316,33 @@ describe('VisibilitySingleton', () => {
   });
 
   describe('loosePermissionsKessel', () => {
+    const mockWorkspaceId = 'a1b2c3d4-e5f6-7890-abcd-ef1234567890';
+
     beforeEach(() => {
       getUser.mockImplementation(() => Promise.resolve(userMock));
       mockedAxios.post.mockReset();
+      mockedAxios.get.mockReset();
+      mockedAxios.get.mockResolvedValue({ data: { data: [{ id: mockWorkspaceId }] } });
+      // Re-initialize to reset the workspace cache
+      initializeVisibilityFunctions({ getUser, getToken, getUserPermissions, isPreview: false });
+      visibilityFunctions = getVisibilityFunctions();
     });
 
-    test('should return false if org_id is missing', async () => {
-      getUser.mockImplementationOnce(() =>
-        Promise.resolve({
-          ...userMock,
-          identity: { ...userMock.identity, org_id: undefined },
-        })
-      );
+    test('should return false if workspace fetch fails', async () => {
+      mockedAxios.get.mockReset();
+      mockedAxios.get.mockRejectedValueOnce(new Error('Workspace fetch failed'));
+      initializeVisibilityFunctions({ getUser, getToken, getUserPermissions, isPreview: false });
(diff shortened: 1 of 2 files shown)


riccardo-forina (commented):
The workspace-scoped approach is correct — good direction. Two things before this leaves draft:

1. The module-level `workspacePromise` cache never clears on failure. If the workspace fetch fails (network blip, 503), the `null` result is cached permanently until the next full page load. Since a cached null means all Kessel permission checks silently return false (hiding nav items for all users), this is the same class of bug that caused the July 8 outage. Fix: reset `workspacePromise = null` in the `.catch()` before returning null, so a retry is possible.

2. The Jira link is `RHCLOUD-XXXXX` — needs a real ticket before merge.

riccardo-forina (commented):
Broader question for @bflorkie @khala: rather than adding more raw Kessel API calls in Chrome, should we bite the bullet and adopt the Kessel SDK (`@project-kessel/kessel-sdk-browser`) in Chrome? It already supports batching and would give us a single, maintained abstraction instead of hand-rolled fetch calls that we keep patching (see the July 8 outage).

This PR could be the starting point for that conversation. Happy to discuss in Slack if that's easier.

riccardo-forina (commented):
Created [RHCLOUD-49438](https://redhat.atlassian.net/browse/RHCLOUD-49438) to track the broader proposal: adopt the Kessel SDK in Chrome as a shared singleton, making Chrome the single permission check provider for all tenant apps. This PR could evolve into that, or be a stepping stone — let's discuss.

karelhala:
@riccardo-forina good point on changing the implementation from raw fetch to using the Kessel SDK. I'll throw [RHCLOUD-49438](https://redhat.atlassian.net/browse/RHCLOUD-49438) to Rehor to see if he can handle it.

[RHCLOUD-49438]: https://redhat.atlassian.net/browse/RHCLOUD-49438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

riccardo-forina:
Happy to hear you're aligned. I'd like to own this implementation — Chrome is critical infrastructure and the Kessel permission architecture is something I want to get right hands-on, especially after the July 8 incident. Happy to keep you in the loop as it progresses.

riccardo-forina:
Ah, Rehor already crunched something. Nevermind, I'll review that

PR #3653, opened 2026-08-13
feat(virtual-assistant): hide VA bubble when chatbot enabled

CodyWMitchell:
### Description

[RHCLOUD-50141](https://redhat.atlassian.net/browse/RHCLOUD-50141)

- Hide the Astro VA badge (and related inline links) when the Help Panel chatbot tab is enabled, using the `platform.chrome.help-panel_chatbot` feature flag.

---

#### Before:

https://github.com/user-attachments/assets/6261fec2-6332-4a68-bca1-dd425ec8aece

#### After:

https://github.com/user-attachments/assets/21a8b2fa-2337-4ff4-b76d-6cf15f331d31

---

---

### Checklist

### AI disclosure

[RHCLOUD-50141]: https://redhat.atlassian.net/browse/RHCLOUD-50141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Virtual Assistant links and entry points from appearing when the help-panel chatbot is enabled.
  * Preserved Virtual Assistant availability when its feature is enabled and the chatbot is disabled.
  * Updated empty-search and not-found experiences to display appropriate fallback content.

* **Tests**
  * Added coverage for all combinations of Virtual Assistant and chatbot feature settings.
  * Verified feature settings are evaluated independently across affected experiences.

### DIFF
--- a/src/components/Routes/VirtualAssistant.tsx
+++ b/src/components/Routes/VirtualAssistant.tsx
@@ -17,6 +17,7 @@ const flaggedRoutes: { [flagName: string]: string } = {
 
 const VirtualAssistant = () => {
   const isVAEnabled = useFlag('platform.va.environment.enabled');
+  const isHelpPanelEnabled = useFlag('platform.chrome.help-panel_chatbot');
   const [showAssistant, setShowAssistant] = useAtom(virtualAssistantShowAssistantAtom);
 
   const { pathname } = useLocation();
@@ -49,7 +50,7 @@ const VirtualAssistant = () => {
     }
   }, [flags, pathname, viableRoutes, setShowAssistant, isVAEnabled]);
 
-  if (!isVAEnabled) {
+  if (!isVAEnabled || isHelpPanelEnabled) {
     return null;
   }
 
--- a/src/components/NotFoundRoute/NotFoundRoute.test.tsx
+++ b/src/components/NotFoundRoute/NotFoundRoute.test.tsx
@@ -26,9 +26,13 @@ jest.mock('@redhat-cloud-services/frontend-components/InvalidObject', () => ({
 describe('NotFoundRoute', () => {
   const useFlag = unleashReact.useFlag as jest.MockedFunction<typeof unleashReact.useFlag>;
 
+  const setFlagMock = (flags: Record<string, boolean>) => {
+    useFlag.mockImplementation((flag: string) => flags[flag] ?? false);
+  };
+
   beforeEach(() => {
     jest.clearAllMocks();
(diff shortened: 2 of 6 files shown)


src/components/Routes/VirtualAssistant.tsx
```diff
@@ -49,7 +50,7 @@ const VirtualAssistant = () => {
     }
   }, [flags, pathname, viableRoutes, setShowAssistant, isVAEnabled]);
 
-  if (!isVAEnabled) {
+  if (!isVAEnabled || isHelpPanelEnabled) {
```
karelhala, line 53: 🟡 The render guard hides VA when `platform.chrome.help-panel_chatbot` is on, but the `useEffect` above still only returns early on `!isVAEnabled` and still calls `setShowAssistant(true)` on matching routes. That leaves the atom set while Scalprum is unmounted, unlike the VA-disabled path which the tests treat as “do not touch the atom.”

Use the same condition in the effect (`if (!isVAEnabled || isHelpPanelEnabled) return`), add the chatbot flag to the effect deps, and add a test like the existing “VA disabled → atom stays false” case.

src/components/Routes/VirtualAssistant.tsx
```diff
@@ -17,6 +17,7 @@ const flaggedRoutes: { [flagName: string]: string } = {
 
 const VirtualAssistant = () => {
   const isVAEnabled = useFlag('platform.va.environment.enabled');
+  const isHelpPanelEnabled = useFlag('platform.chrome.help-panel_chatbot');
```
karelhala, line 20: 🔵 This flag is `isChatbotTabEnabled` in NotFoundRoute and EmptySearchState. Here it is `isHelpPanelEnabled`, which layouts already use for `platform.chrome.help-panel`.

Rename to `isChatbotTabEnabled` (including the render/effect usages) so the two help-panel flags stay distinct.

```suggestion
  const isChatbotTabEnabled = useFlag('platform.chrome.help-panel_chatbot');
```

karelhala (commented):
Two follow-ups on VirtualAssistant: the chatbot flag should also gate the showAssistant effect, and the variable name should match the other call sites.

karelhala (approved):

karelhala:
@CodyWMitchell looks like there are some unsigned commits.

PR #3655, opened 2026-08-14
feat(resilience): degraded banner and store

justinorringer:
### Description

- Degraded state banner exposed with remote hook
- Degraded atom as well
- Not exposed through chrome api and shared scope (can add if desired)

Trigger it in the console with debug tool

```
const clear = insights.chrome.enable.degradedStateBanner();
clear(); // Clears degraded state
```
---

#### After:
Screenshot from 2026-08-14 13-14-52

---

---

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a service health banner that alerts users when personalization, entitlements, configuration, or feature-flag services are degraded.
  * The banner lists affected services, provides retry guidance, confirms core functionality remains available, and hides when services recover.
  * Added a public Service Health API for reading and updating degradation status.
  * Added documentation and local testing guidance for the new API.

* **Tests**
  * Added comprehensive coverage for service health state, banner behavior, accessibility, feature-flag handling, and recovery.


catastrophe-brandon:
Bug: UserToggle renders an empty dropdown when userMenu is not provided

  UserToggle now guards every menu item behind userMenu?.showX && (...). When userMenu is undefined -- which happens in any <Header> rendered without toolbarConfig.userMenu -- all items evaluate to falsy and nothing
  renders. The user sees their name/avatar but gets an empty dropdown on click.

  Affected render sites:
  - ProductSelection.tsx -- <Header />
  - BetaInfoModal.tsx -- <Header />
  - FavoritedServices.tsx -- <Header />
  - AllServices.tsx -- <Header breadcrumbsProps={{ hideNav: true }} />

  The old behavior showed all items unconditionally. Suggest defaulting to true when the property is absent, e.g.:

  {(userMenu?.showLogout !== false) && (

  or resolving defaults at the top of DropdownItems:

  const menu = { showMyProfile: true, showMyUserAccess: true, showUserPreferences: true, showInternal: true, showLogout: true, ...userMenu };

  This preserves the existing behavior for callers that don't pass userMenu, while still letting Lightwell (and future layouts) selectively hide items.

justinorringer:
@catastrophe-brandon thanks for taking a look at this one, that bug might be relevant but doesnt have to do with this pr. ill open a separate ticket. https://redhat.atlassian.net/browse/RHCLOUD-50389

catastrophe-brandon (dismissed):

src/components/DegradedStateBanner/DegradedStateBanner.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const degradedServices: string[] = [];
+  if (serviceHealth.userPersonalization) degradedServices.push('User Preferences');
+  if (serviceHealth.entitlements) degradedServices.push('Entitlements');
+  if (serviceHealth.configFromCache) degradedServices.push('Navigation Configuration');
+  if (serviceHealth.featureFlags) degradedServices.push('Feature Flags');
+
+  return (
+    <Banner status="warning" screenReaderText="Warning banner">
+      <div className="pf-v6-u-text-align-center">
+        <ExclamationTriangleIcon /> Core functionality is available, but some services are degraded: <strong>{degradedServices.join(', ')}</strong>. Try again
```
florkbr: @justinorringer Banner text and service labels are hardcoded in English. The rest of our codebase uses react-intl (formatMessage). See Banners/Banner.tsx:11 for reference pattern. We should wrap user-facing strings in intl.formatMessage() with messages defined in src/locales/Messages.ts (especially with the upcoming translation effort).

src/components/DegradedStateBanner/DegradedStateBanner.tsx
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  const degradedServices: string[] = [];
+  if (serviceHealth.userPersonalization) degradedServices.push('User Preferences');
+  if (serviceHealth.entitlements) degradedServices.push('Entitlements');
+  if (serviceHealth.configFromCache) degradedServices.push('Navigation Configuration');
+  if (serviceHealth.featureFlags) degradedServices.push('Feature Flags');
+
+  return (
+    <Banner status="warning" screenReaderText="Warning banner">
```
florkbr: NIT - we could include degraded service names for better screen reader experience: "Warning: some services are degraded: User Preferences".

cypress/component/DegradedStateBanner.cy.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    cy.mount(
+      <FlagProvider config={mockUnleashConfig}>
+        <Provider store={store}>
+          <DegradedStateBanner />
+        </Provider>
+      </FlagProvider>
+    );
+
+    cy.get('.pf-m-warning').should('exist');
```
florkbr: @justinorringer NIT - using this class couples our E2E with PF classes. Not terrible, but we could check for something we control like adding a `data-` attribute on the HTML element/component.

src/components/DegradedStateBanner/DegradedStateBanner.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <DegradedStateBanner />
+      </Provider>
+    );
+
+    expect(screen.getByText(/navigation configuration/i)).toBeInTheDocument();
+  });
+
+  it('should render banner when featureFlags are degraded', () => {
+    const store = createStore();
+    store.set(degradedStateAtom, {
```
florkbr: @justinorringer can we please take a look over all the tests added in this PR and see if we can move any setup/duplication into a `beforeEach`? I see a bit of duplication here.

src/state/atoms/degradedStateAtom.ts
```diff
@@ -0,0 +1,44 @@
...
+  userPersonalization: false,
+  entitlements: false,
+  configFromCache: false,
+  featureFlags: false,
+};
+
+export const degradedStateAtom = atom<ServiceHealthStatus>(initialState);
+
+// Write-only atoms for updating individual services
+export const setUserPersonalizationDegradedAtom = atom(null, (get, set, degraded: boolean) => {
```
florkbr: @justinorringer we should replace the four write-only atoms (setUserPersonalizationDegradedAtom, setEntitlementsDegradedAtom, setConfigFromCacheDegradedAtom, setFeatureFlagsDegradedAtom) with a single generic setter IMO to avoid duplication:

```ts
  export const setServiceDegradedAtom = atom(
    null,
    (get, set, { service, degraded }: { service: keyof ServiceHealthStatus; degraded: boolean }) => {
      const current = get(degradedStateAtom);
      set(degradedStateAtom, { ...current, [service]: degraded });
    }
  );
```

This reduces the boilerplate — right now each service needs its own write-only atom, and adding a fifth service means adding another atom, another setter in the hook, and more test surface. A single generic setter covers all cases and makes it easier to extend later.

  The boolean object shape itself is fine for the current scope, though we may need to expand it later.

config/webpack.plugins.js
```diff
@@ -75,6 +75,7 @@ const plugins = (dev = false, beta = false, restricted = false) => {
         './analytics/intercom/OpenShiftItercom': resolve(__dirname, '../src/components/OpenShiftIntercom/OpenShiftIntercomModule.tsx'),
         './analytics/intercom/useOpenShiftIntercomStore': resolve(__dirname, '../src/state/stores/openShiftIntercomStore.ts'),
         './theme/useDarkModeStore': resolve(__dirname, '../src/state/stores/darkModeStore.ts'),
+        './serviceHealth/useDegradedState': resolve(__dirname, '../src/hooks/useDegradedState.ts'),
```
florkbr: @justinorringer while I am usually all for exposing federated modules, in this case I'm not sure we want to expose the banner. AFAIK the requirements have the banner as something chrome internally controls/renders. Once we expose this as a federated module, anyone can trigger the banner which means we must maintain it as a public API surface (even if it is not exposed through chrome api type or shared scope)


I suggest we leave this off until there is a justification to expose it to consumers?

src/components/DegradedStateBanner/DegradedStateBanner.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    expect(container.textContent).toMatch(/navigation configuration/i);
+    expect(container.textContent).toMatch(/feature flags/i);
+  });
+
+  it('should always show warning variant', () => {
+    const { container } = renderBanner({ userPersonalization: true, entitlements: false, configFromCache: false, featureFlags: false });
+
+    // Verify warning variant (PatternFly v6 Banner)
+    expect(container.textContent).toMatch(/user preferences/i);
+    expect(container.querySelector('[class*="pf-m-warning"]')).toBeInTheDocument();
```
florkbr, line 96: @justinorringer NIT (we can follow up) - but we should update to use the same data attribute here too?

florkbr (approved):
Looks like a good first pass! @justinorringer can you update your branch with master (out of date)? Otherwise :shipit: and check stage after!

PR #3657, opened 2026-08-18
Fix: Potential fix for access inventory call before flags loaded

jjaquish:
### Description

Potential fix for access inventory calls happening before unleash flags load. As per this slack thread: https://redhat-internal.slack.com/archives/C064X43CMLK/p1786459021301489 

---

#### After:

---

---

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for updated inventory permission handling when the relevant feature is enabled.
  * Global filtering now waits for feature flags to finish loading before requesting permissions.

* **Bug Fixes**
  * Prevented premature permission requests while feature flags are loading or unavailable.
  * Ensured cached inventory permissions are cleared when updated handling is enabled.
  * Continued permission requests for non-inventory applications.

### DIFF
--- a/src/auth/fetchPermissions.test.js
+++ b/src/auth/fetchPermissions.test.js
@@ -93,4 +93,50 @@ describe('fetchPermissions', () => {
 
     unleashClientExists.mockReturnValue(false);
   });
+
+  it('should skip inventory RBAC call when hbi.rbac-v2 flag is enabled', async () => {
+    unleashClientExists.mockReturnValue(true);
+    getUnleashClient.mockReturnValue({
+      isEnabled: jest.fn((flag) => flag === 'hbi.rbac-v2'),
+    });
+
+    const result = await fetchPermissions('uSeRtOkEn', 'inventory', true);
+
+    expect(result).toEqual([]);
+    expect(global.rbacApiCalled).toEqual(0);
+
+    unleashClientExists.mockReturnValue(false);
+  });
+
+  it('should invalidate cached inventory results when hbi.rbac-v2 becomes enabled', async () => {
+    // First call caches inventory permissions with flag disabled
+    const result1 = await fetchPermissions('uSeRtOkEn', 'inventory');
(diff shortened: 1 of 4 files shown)


astrozzc:
I got two minor suggestions:
1. flagsError still fires the inventory call — Minor
If Unleash errors, useFlag stays false and the new guard treats flagsError as “go ahead”:
GlobalFilter.tsx Ln 203–205
```
if (!flagsReady && !flagsError) {
  return;
}
```
That can 400 again for V2 orgs when flags fail. Safer: only fetch when flagsReady is true; on error, leave hasAccess false (same as v2). That flagsError proceed pattern exists in visibleBundlesAtom, so this is optional, not a merge blocker.

2. fetchPermissions does not wait for Unleash (This seems not important)
Global Filter now waits. fetchPermissions does not.

The skip in fetchPermissions is only:

if (app === 'inventory' && unleashClientExists() && getUnleashClient().isEnabled('hbi.rbac-v2')) {
  return [];
}
Until Unleash has loaded, that condition is false (isEnabled is false, or the client is not there yet). The helper then still does GET /access/?application=inventory.

Landing (your repro): Global Filter is the caller. It now waits for flagsReady, then sees hbi.rbac-v2 and never calls. That 400 goes away.

Other callers: anything else that calls chrome.getUserPermissions('inventory') — inventory’s <RBACProvider appName="inventory"> is the main one — does not wait. If that runs before flags are ready, chrome still hits v1 and RBAC still returns 400.

So: the PR fixes chrome’s own race. It does not make the shared RBAC helper wait. That only matters if some other app asks for inventory permissions during the same Unleash window.

astrozzc (approved):
LGTM! Thanks!

jjaquish:
@coderabbitai review

jjaquish:
@coderabbitai review

jjaquish:
@coderabbitai resolve

karelhala (approved):

PR #3660, opened 2026-08-21
feat(cache): add IndexedDB fallback cache for critical Chrome config files

petrsimon:
### Description

Adds an IndexedDB-backed fallback cache (via `localforage`) for critical Chrome configuration files that must load for the platform to function. When chrome-service or Akamai returns 5xx errors, Chrome can now serve last-known-good config instead of breaking entirely.

Integrated into:
- **Fed modules** (`loadFedModules`) - Module Federation manifests
- **SSO config** (`loadSSOConfig`) - OIDC/Keycloak endpoints  
- **Bundle navigation** (`fetchNavigationFiles` feoGenerated mode) - Navigation structure

On cache hit, callers receive `fromCache: true` to optionally display degraded-state indicators. Functions with static fallbacks (like `loadSSOConfig`) preserve them as the last resort after cache miss.

https://issues.redhat.com/browse/RHCLOUD-50158

---

**Cache schema versioning:** Cache keys are prefixed with `v1:` to allow safe schema evolution. Incrementing the version invalidates old cache entries.

**TTL strategy:** Default 7-day TTL balances freshness vs availability. Config changes deploy via rolling updates, so stale cache is still functional.

**Test coverage:**
- `cacheFetch.test.ts`: cache envelope, TTL, 4xx-vs-5xx fallback policy, payload guard
- `loadFedModules.test.ts`: live/CSC/IndexedDB fallback chain + per-entry sanitization and Sentry reporting
- `fetchNavigationFiles.test.ts`: malformed-bundle filtering incl. nested/deeply-nested nav items
- `common.test.ts`: `isFedModulesConfig` (incl. `modules`/`routes`) and `sanitizeFedModules` (keep-good/drop-bad, throw-on-unusable)
- Full unit suite green: 100 suites, 1076 tests

**localforage driver priority:** IndexedDB → WebSQL → localStorage. Graceful degradation if storage APIs unavailable.

**Runtime validation & graceful degradation (from review):**
- **Per-entry, not all-or-nothing.** `loadFedModules` now sanitizes each live response (`sanitizeFedModules`): individually malformed apps are dropped and the valid ones are kept, so one bad entry can't discard the entire module map. Verified against live prod/stage config — the real `trust` entry (missing `manifestLocation`) is now dropped-and-reported instead of nuking the whole map. The whole-map guard (`isFedModulesConfig`) is retained only for cache reads, which hold already-sanitized data.
- **Deep validation of consumed shapes.** `isFedModulesConfig` now validates the optional `modules`/`RemoteModule`/`routes` fields that `generateRoutesList` maps over, and `isBundleNavigation` recursively validates nested `navItems`/`routes` before `normalizeBundle`/`normalizeNavItem` destructure them — closing crash paths where malformed-but-accepted config would throw during bootstrap.
- **Observability.** Dropped fed-modules entries are logged (`console.warn` with the app key + reason) and reported to Sentry (`captureException`, `level: warning`, `tags: { area: 'fed-modules' }`) so bad config surfaces in monitoring rather than failing silently.

---

### Checklist
- [x] _(Optional) QE: OUIA changed, test impact, no coverage_
- [x] _(Optional) UX: end-user UX modified, designs need sign-off_

### AI disclosure
Assisted by: Claude Code (Sonnet 4.5)

## Summary by CodeRabbit

- **New Features**
  - Added resilient loading with automatic caching for SSO configuration, navigation content, and federated modules.
  - Cached results remain available for a limited period when services are unavailable.
  - Added fallback service handling and retry behavior to improve content availability.
  - SSO loading now indicates when cached data is being used.

- **Bug Fixes**
  - Improved validation of live and cached configuration and navigation data.
  - Prevented malformed, expired, or outdated cached entries from affecting content.
  - Preserved appropriate error handling when requests cannot be recovered.


petrsimon:
@coderabbitai resolve

karelhala:
P1: `isFedModulesConfig` validates only each module’s `manifestLocation`; it does not validate optional `modules` arrays or their nested `routes`. For example, `{ app: { manifestLocation: "/app.json", modules: [{ module: "App", routes: null }] } }` passes validation, then `generateRoutesList` calls `.map()` on invalid data and crashes Chrome bootstrap. Please validate `modules`, `RemoteModule` entries, and `routes` before caching or using the config.

karelhala:
P1: `isBundleNavigation` validates only top-level fields. A payload containing a valid bundle with `routes: [null]` or a nested `navItems` primitive passes validation, then `normalizeNavItem` destructuring throws. This bypasses IndexedDB fallback and can break navigation bootstrap. Please validate nested navigation items recursively or safely filter malformed children before normalization.

karelhala (approved):

PR #3663, opened 2026-08-21
feat(lightwell): Adjust horizontal navigation

marusak:
Two changes:
    - Do not distinguish between stage and prod. Threat them equally.
    - We introduced two separate flags for Beacon and Lens. Use those and update
    logic.

No visual changes.

## Summary by CodeRabbit

- **New Features**
  - Navigation now displays separate entries for Repositories, Lens, and Beacon based on access permissions.
  - Navigation appears when at least two permitted destinations are available.
  - Links and active-route behavior are consistent across environments.

- **Bug Fixes**
  - Restricted or unavailable destinations are hidden from navigation.
  - Permission loading and service failures no longer expose inaccessible navigation items.
  - Navigation now updates correctly while permissions are being evaluated.


src/components/Navigation/LightwellNavigation.tsx
```diff
@@ -2,102 +2,63 @@ import React, { useEffect, useState } from 'react';
...
-            return { item, visible: false };
-          }
-        })
-      );
+      const [hasLightwell, hasLensAndBeacon] = await Promise.all([isFeatureAccessible(LIGHTWELL_PERMISSION), isFeatureAccessible(LENS_AND_BEACON_PERMISSION)]);
 
       if (!cancelled) {
-        setVisibleItems(results.filter(({ visible }) => visible).map(({ item }) => item));
-        setIsLoading(false);
+        setShowNav(hasLightwell && hasLensAndBeacon);
```
karelhala: Requiring both flags is stricter than the old `visibleItems.length <= 1` check, and it is not equivalent.

Lens and Beacon share `lightwellbeaconandlens.accessible`, so that entitlement is **two** tabs, not one. Master already hid a Repositories-only bar; it still showed Lens + Beacon when only the shared flag was true. `hasLightwell && hasLensAndBeacon` drops that case.

A boolean swap is not enough either: when `showNav` is true this maps all of `NAV_ITEMS`, so users without `lightwell.accessible` would get a Repositories link they should not see.

Restore item-level visibility and hide only when fewer than two items remain, e.g. keep the two permission results, build `visibleItems` from them, `return null` when `visibleItems.length <= 1`, and render `visibleItems` instead of `NAV_ITEMS`.

karelhala (commented):
The all-or-nothing `&&` is stricter than the previous “hide when there is only one tab” rule and drops the Lens+Beacon-only case. Details on the line.

src/components/Navigation/LightwellNavigation.tsx
marusak: You need both - You need lightwell access to see anything under /lightwell. And then LensAndBeacon enables you to see the navigation and content of those two pages. It should never happen that someone would see Lens and Beacon but would not see Repositories.

marusak: Hold on please, we will probably split the lens and beacon into two flags.

karelhala (approved):

src/components/Navigation/LightwellNavigation.tsx
```diff
@@ -23,27 +23,18 @@ const featuresPermission = (accessor: string): LightwellNavItemConfig['permissio
...
+ * | Lens           | lightwelllens   | lightwelllens.accessible     |
+ * | Beacon         | lightwellbeacon | lightwellbeacon.accessible   |
  */
-const PROD_NAV_ITEMS: LightwellNavItemConfig[] = [
+const NAV_ITEMS: LightwellNavItemConfig[] = [
   { label: 'Repositories', path: LIGHTWELL_PATH, permissions: featuresPermission('lightwell.accessible') },
-  { label: 'Lens', path: `${LIGHTWELL_PATH}/lens`, permissions: featuresPermission('lightwellbeaconandlens.accessible') },
-  { label: 'Beacon', path: `${LIGHTWELL_PATH}/beacon`, permissions: featuresPermission('lightwellbeaconandlens.accessible') },
+  { label: 'Lens', path: `${LIGHTWELL_PATH}/lens`, permissions: featuresPermission('lightwelllens.accessible') },
+  { label: 'Beacon', path: `${LIGHTWELL_PATH}/beacon`, permissions: featuresPermission('lightwellbeacon.accessible') },
```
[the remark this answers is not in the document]
karelhala: This is no longer true, if there are more than 2 items in the navigation we show it.

karelhala (approved):

PR #3685, opened 2026-09-14
Fix Chrome cache fallback reliability

petrsimon:
## Summary

Add resilient network-first IndexedDB fallback for generated Chrome configuration, controlled by the runtime Unleash flag `platform.chrome.config-cache-fallback`.

- Post-authenticated configuration uses the flag for persistent cache reads and writes.
- Unknown, unavailable, not-ready, or errored Unleash state fails closed: live requests continue, but cache fallback is disabled.
- Live requests start immediately and are never delayed by flag readiness.
- Disabled mode bypasses IndexedDB initialization, reads, purges, and writes.
- SSO and federated-module bootstrap fallback remains explicitly enabled because it runs before authenticated Unleash initialization; there is no separate Webpack or environment toggle.
- Validation, sanitization, filtering, URL checks, and Sentry reporting remain active for both live and cached data.
- Configuration-source status is updated on cache failures before the original error is rethrown.

Validation completed:

- Full Jest: 108 suites / 1,182 tests passed
- TypeScript type-check passed
- ESLint: 0 errors; existing warnings only
- Prettier passed
- Production build passed; existing webpack asset-size warnings only
- Frontend CRD validation passed
- Playwright cache-fallback spec discovery passed: 8 tests

Related issue: https://issues.redhat.com/browse/RHCLOUD-51276

---

### Description

Generated configuration now uses a network-first cache fallback for selected post-authenticated requests. The `platform.chrome.config-cache-fallback` Unleash flag controls only persistent IndexedDB fallback after the feature-flag provider is available; it does not disable live requests or configuration validation.

Bootstrap SSO and federated-module loading use an explicit cache-enabled policy because those requests occur before authenticated Unleash initialization. The former build-time `CONFIG_CACHE_FALLBACK_ENABLED` Webpack/HTML switch was removed because it was a separate control that could not be driven by the runtime flag.

https://issues.redhat.com/browse/RHCLOUD-51276

---

### Screenshots

N/A — no visible UI changes.

---

- Cache fallback is fail-closed when Unleash is missing, loading, disabled, or in an error state.
- A successful live response is returned immediately; cache writes happen asynchronously only after the cache policy allows them.
- A failed live request can use cache only for network/5xx failures and only when the runtime flag is enabled.
- 4xx and canceled requests never use stale cache data.
- Bootstrap caching intentionally remains enabled independently of the runtime flag.

---

### Checklist

- [x] Accessibility: N/A — no visible UI changes
- [x] _(Optional) QE: OUIA changed, test impact, no coverage_ — unit and browser regression coverage added
- [x] _(Optional) UX: end-user UX modified, designs need sign-off_ — no visual changes

### AI disclosure

Assisted by: [Pi](https://pi.dev) — gpt-5.6-luna high

## Summary by CodeRabbit

- **New Features**
  - Added resilient fallback to cached configuration for generated navigation, search index, service tiles, SSO, and federated modules when live requests fail.
  - Added degraded-state tracking and recovery notifications when cached configuration is being used.
  - Added runtime control through `platform.chrome.config-cache-fallback` for post-authenticated configuration.
  - Kept bootstrap fallback explicitly enabled without a separate build-time toggle.

- **Bug Fixes**
  - Canceled requests no longer trigger cache fallback.
  - Malformed or unusable configuration responses cannot replace valid cached snapshots.
  - Cache fallback is skipped when configuration flags are unavailable or disabled.


petrsimon:
@coderabbit resolve

petrsimon:
@coderabbit review

petrsimon:
@coderabbit full review

florkbr:
@petrsimon dumb Q - but would it be possible to put this behind a dark launch flag that way shipping chrome is not tied to enabling this caching?

petrsimon:
@coderabbit resolve

charlesmulder:
## 🤖 Automated Code Review Findings (`claude-sonnet-5`)

### Summary

| Metric              | Details                                                                           |
|---------------------|-----------------------------------------------------------------------------------|
| **Files Reviewed**  | 21 files (10 with findings, 11 clean)                                             |
| **Total Findings**  | 12 (1 🔴 High, 11 🟡 Medium)                                                      |
| **Key Focus Areas** | Cache fallback timing, validation redundancy, test isolation & telemetry accuracy |
---

### 🔴 High Severity

#### [`src/utils/cacheFetch.ts` (lines 219–223)](src/utils/cacheFetch.ts#L219-L223)
**4xx client errors unnecessarily delayed by Unleash flag timeout in catch block**

* **Issue:** The catch block awaits `cacheFallbackPolicy` before evaluating `shouldUseCacheFallback(err)`. Because `cacheFallbackPolicy` can take up to `FLAG_READY_TIMEOUT_MS` (5s) when Unleash is not yet initialized, client errors (e.g., 401s during bootstrap) and canceled requests are blocked for up to 5 seconds before failing fast.
* **Recommendation:** Run the synchronous `shouldUseCacheFallback(err)` check first, and only await `cacheFallbackPolicy` if the error qualifies for fallback.

---

### 🟡 Medium Severity

#### 📁 Runtime & Caching Logic

* **[`src/utils/cacheFetch.ts` (lines 211–227)](src/utils/cacheFetch.ts#L211-L227) — Duplicate `getCache()` invocations**
  * `getCache()` is called independently in both the write path and the catch read path. If `getCache()` runs non-trivial initialization or stale-version purges, consider resolving the cache instance once per `cacheFetch` invocation.

* **[`src/utils/fetchNavigationFiles.ts` (lines 118–135)](src/utils/fetchNavigationFiles.ts#L118-L135) — All-or-nothing cache admission drops valid data**
  * The fetcher returns raw data while `isBundleNavigationArray` requires 100% validity to persist to IndexedDB. A single bad bundle entry prevents caching even valid subsets, leaving offline caches permanently stale. Consider persisting the filtered valid subset while logging/reporting anomalies separately.

* **[`src/utils/common.ts` (lines 450–452)](src/utils/common.ts#L450-L452) — Failure fallback misreported as fresh network fetch**
  * In `loadSSOConfig` and `loadFedModules`, when a fetch fails completely and falls back to hardcoded defaults (e.g. `DEFAULT_SSO_ROUTES`), `reportConfigSource(..., false)` is called. Passing `false` conflates fresh live network fetches with static fallbacks. Consider distinguishing `cache | live | fallback`.

* **[`src/utils/configCacheStatus.ts` (lines 15–19)](src/utils/configCacheStatus.ts#L15-L19) — Module singleton state may leak across tests**
  * `cachedSources` and `listeners` are module-level sets. Ensure test suites call `resetConfigCacheStatus()` in `afterEach` hooks (or expose a scoped factory) to prevent state bleed across Jest test runs.

---

#### 📁 State & Validation

* **[`src/state/atoms/localSearchAtom.ts` (lines 33–49)](src/state/atoms/localSearchAtom.ts#L33-L49) — Redundant $O(n)$ array validation**
  * `isGeneratedSearchIndex` performs full array checks in the axios `.then` callback and potentially again via `cacheFetch`'s validator. Consolidate validation so it runs only once per fetch/cache read.

* **[`src/state/atoms/visibleBundlesAtom.ts` (lines 74–141)](src/state/atoms/visibleBundlesAtom.ts#L74-L141) — Duplicated type guards**
  * `isServiceLinkConfig` and `isServiceTilesConfig` manually duplicate logic already present in `isAllServicesLink` and `isAllServicesGroup`. Reusing existing type guards prevents schema drift.

* **[`src/state/atoms/visibleBundlesAtom.ts` (lines 108–111)](src/state/atoms/visibleBundlesAtom.ts#L108-L111) — Incomplete group structural validation**
  * When `isGroup === true`, the check validates `title` and `links`, but does not verify that `href` is absent or reject nested groups. Reusing `isAllServicesGroup` resolves this.

* **[`src/state/atoms/visibleBundlesAtom.ts` (lines 82–84)](src/state/atoms/visibleBundlesAtom.ts#L82-L84) — Inconsistent validation on non-FEO path**
  * `loadServiceTiles` strictly validates and throws on malformed payloads when `feoGenerated` is true, but bypasses validation entirely on the non-FEO branch. Both branches should validate payloads uniformly.

---

#### 📁 Test Suite & Playwright Config

* **[`playwright.config.ts` (line 7)](playwright.config.ts#L7) — `testDir` widened to `./playwright`**
  * Broadening discovery from `./playwright/e2e` to `./playwright` may cause helper scripts, fixtures, or global setups matching default test patterns to run as test specs. Verify whether this was intentional, or scope it via explicit `testMatch` filters.

* **[`playwright/integration/cache-fallback.spec.ts` (lines 396–404)](playwright/integration/cache-fallback.spec.ts#L396-L404) — Route interceptor registered before cache seed**
  * In the malformed data test, `setupConfigRoutes` is called before `seedConfigCache`. Because `seedConfigCache` navigates internally, requests fired during seeding can pollute `observations` sets before the test target navigation begins. Move `seedConfigCache` ahead of route registration.

* **[`src/utils/cacheFetch.test.ts` (lines 88–96)](src/utils/cacheFetch.test.ts#L88-L96) — Microtask scheduling race in test**
  * The test relies on synchronous dispatch between `rejectFetch` and `handlers.ready()`. If rejection handling dispatches across microtasks, this introduces test flakiness. Add `await Promise.resolve()` after `rejectFetch` to make sequencing deterministic.

petrsimon:
Thanks @charlesmulder some good catches. I only fixed the first one, the rest, I think, would best be handled as follow up not to bloat this PR anymore:

Followups:
 1. Make cache fallback policy lazy:
     - Avoid background Unleash listeners/timer for 4xx and cancellation errors.
     - Safer as focused followup becausefor network/5xx failures, five-second readiness timeout starts after request fails. Current implementation overlaps that timeout with request duration, so failure could become slower.

 2. Validate legacy non-FEO service-tile responses:
     - Definitely useful runtime hardening.
     - Pre-existing behavior; outside current cache-fallback scope.

3. Replace reportConfigSource(..., boolean) with 'live' | 'cache' | 'built-in' | 'failed'.
    - Distinguishes live data from hardcoded fallback.
    - Makes telemetry and degraded state accurate.
    - Preserves current boolean bridge initially through derived status.
    
---

We dare to reject all the one below :) 

**Medium ones**

 1. Duplicate getCache() calls — reject
     - Write path runs after successful fetch.
     - Read path runs after failed fetch.
     - Paths mutually exclusive.
     - getCache() already memoizes instance.
     - Resolving it early could initialize storage unnecessarily.

 2. All-or-nothing navigation admission — reject
     - Intentional safety rule.
     - Live response may filter malformed entries for current session.
     - Only complete, valid snapshots replace last-known-good cache.
     - Caching partial data could permanently remove navigation entries during outages.

 5. Module singleton test leakage — already handled
     - Tests call resetConfigCacheStatus().
     - Tests retaining listeners track and unsubscribe them.
     - Module-level state is intentional production behavior.
     
**State & Validation**
 1. Redundant search-index validation — reject
     - Axios callback validates live response and throws on malformed data.
     - cacheFetch guard validates cache admission and cached reads.
     - cacheFetch intentionally returns malformed live data when guard alone fails.
     - Removing Axios validation would change behavior and allow malformed live search data through.
     - Two checks serve different contracts.

 2. Duplicated service-tile guards — reject
     - Existing UI guards are not equivalent:
         - isAllServicesGroup() checks only isGroup === true.
         - isAllServicesLink() checks only truthy href.
     - They do not validate titles, arrays, or child structures.
     - Reusing them would weaken cache validation and increase schema-drift risk.

 3. Incomplete group validation — reject
     - Extra href on a group is harmless and allowed by structural validation.
     - Rejecting unknown/extra properties would hurt forward compatibility.
     - Normal nested groups are already rejected because children must pass isServiceLinkConfig.
     - Only a strange hybrid nested group containing both isGroup: true and valid href could pass. No demonstrated production issue.
     - Reusing permissive UI guards would not solve it.
     
**Tests**
 1. testDir: './playwright' — intentional
     - Required to discover playwright/integration/*.spec.ts.
     - Playwright default matching only discovers test/spec filename patterns.
     - Normal helpers, fixtures, and global setup files are not executed as tests.
     - Discovery already reports expected 95 tests across 24 files.
     - Explicit testMatch would be redundant unless future files create naming conflicts

 2. Route interceptor before cache seed — safe
     - seedConfigCache() navigates to silent-check-sso.html.
     - That page does not request tracked config APIs.
     - Observation sets only change for matching config requests or relevant warnings.
     - Therefore seeding does not pollute observations.
     - Moving seed first would improve consistency with other scenarios, but offers no functional fix.

 3. Microtask scheduling race — reject
     - Both possible schedules produce same behavior:
         - Readiness resolves first; catch later reads resolved policy.
         - Catch runs first; it awaits policy until readiness resolves.
     - Test intentionally models flag readiness occurring while failed fetch is being processed.
     - Adding await Promise.resolve() after rejectFetch() changes sequencing but does not make assertion stronger.
     - No demonstrated race or flakiness. Proof us wrong 😁

charlesmulder (approved):

PR #3688, opened 2026-09-15
fix(ci): use Playwright image for scheduled platform infra tests

catastrophe-brandon:
### Description

The scheduled platform infrastructure job fails when Playwright tries to install system dependencies using `su` in CI. Use the Playwright image with browsers and system dependencies preinstalled, removing that installation step.

The test step explicitly runs as the image's `pwuser` (UID/GID 1000) with `runAsNonRoot: true`. Pin the existing `v1.62.1-jammy` image to the multi-platform manifest digest `sha256:b3251f7ff1a9fa559a28d1c67eaa15fc1a9800f7845e82756caea7842967f615` so the image is immutable while retaining AMD64 and ARM64 support. The temporary workspace keeps the checkout and caches writable.

Document the security context and image update procedure in `playwright/README.md`.

[RHCLOUD-47546](https://issues.redhat.com/browse/RHCLOUD-47546)

Example failure without this change:

```
> insights-chrome@0.0.0 postinstall
> rm -rf .webpack-cache
added 1759 packages, and audited 1760 packages in 58s
338 packages are looking for funding
  run `npm fund` for details
12 vulnerabilities (1 low, 10 moderate, 1 high)
To address issues that do not require attention, run:
  npm audit fix
To address all issues (including breaking changes), run:
  npm audit fix --force
Run `npm audit` for details.
Installing Playwright...
BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu24.04-x64 as a fallback.
Installing dependencies...
Switching to root user to install dependencies...
Password: su: Authentication failure
```

---

### Screenshots

N/A

#### After:

---

Validation:

- Production build passed (two warnings).
- All 106 Jest suites passed: 1,161 tests and 33 snapshots.
- Frontend CRD validation, changed-file Prettier checks, YAML parsing, and shell syntax checks passed.
- Verified the pinned manifest SHA256 and that its platform digests match the existing image tag.
- `npm run verify` stopped on an existing formatting issue in `.pytest_cache/README.md`; build and unit tests were run separately.

Runtime validation in Tekton remains pending, including admission of UID/GID 1000 and test-result write permissions.

---

### Checklist

### AI disclosure

Used Codex CLI/OpenAPI to slap together the changes. Tekton is not a good time.

[RHCLOUD-47546]: https://redhat.atlassian.net/browse/RHCLOUD-47546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Updated the automated testing environment to use a Playwright container with browsers preinstalled and pinned for consistency.
  * Configured tests to run as a non-root user with a writable temporary workspace.
  * Improved temporary workspace configuration for more reliable npm and cache operations.
  * Removed the separate browser installation step from the test pipeline.

* **Documentation**
  * Added guidance for scheduled platform infrastructure tests, including stage redirect and route-reachability coverage.


.tekton/platform-infra-tests-pipeline.yaml
```diff
@@ -24,7 +24,8 @@ spec:
             description: "Test output in Konflux standardized format"
         steps:
           - name: run-tests
-            image: registry.access.redhat.com/ubi9/nodejs-20:latest
+            # Keep in sync with @playwright/test in package-lock.json.
+            image: mcr.microsoft.com/playwright:v1.62.1-jammy
```
florkbr: @catastrophe-brandon I realize we use MCR in other tekton pipelines already:
- https://github.com/RedHatInsights/insights-chrome/blob/master/.tekton/run-tests-task.yml#L74
- https://github.com/RedHatInsights/insights-chrome/blob/master/.tekton/insights-chrome-pull-request.yaml#L319

That said, this PR adds another direct MCR image pull. From a supply-chain perspective, could we use a Red Hat/Konflux-approved or mirrored image instead? If not, we should document why MCR is acceptable here.

I understand this avoids `npx playwright install --with-deps`, which requires su/root in some CI environments. My concern is that the replacement image trades install-time privilege requirements for running the full test step as root. Since it still executes npm ci and dependency lifecycle scripts, could we use an approved/mirrored image configured for non-root execution?

If MCR is retained, I’d also suggest pinning the image to an immutable digest rather than the mutable v1.62.1-jammy tag, and configuring it to run as non-root. The Playwright image defaults to root, and this step runs npm ci, which executes dependency lifecycle scripts.

catastrophe-brandon: >  could we use an approved/mirrored image configured for non-root execution?

Based on the docs, Ubuntu-based images are the requirement for Microsoft Playwright. To pull this off, we might need to build our own Ubuntu-based Playwright image and maintain it. I'm not immediately aware of any images out there (Microsoft is treated as the gold standard) but I can take a look around.

catastrophe-brandon: Updated to make use of the `pwuser` for executing the tests, which is a non-root context. Also updated the security context definition in the yaml to restrict to non-root. Pinned the exact SHA so we don't get any surprises.

I'll continue researching if there are any Konflux-approved images for Playwright.

catastrophe-brandon: Scanned https://github.com/konflux-ci/build-definitions for mentions of an approved Playwright image, but found no leads.

.tekton/platform-infra-tests-pipeline.yaml
```diff
@@ -24,7 +24,13 @@ spec:
             description: "Test output in Konflux standardized format"
         steps:
           - name: run-tests
-            image: registry.access.redhat.com/ubi9/nodejs-20:latest
+            # Playwright v1.62.1-jammy; keep in sync with @playwright/test in package-lock.json.
+            image: mcr.microsoft.com/playwright@sha256:b3251f7ff1a9fa559a28d1c67eaa15fc1a9800f7845e82756caea7842967f615
```
justinorringer, line 28: Will the test step fail if the hashes are out of sync, or is this something we will have to remember?

florkbr (approved):

.tekton/platform-infra-tests-pipeline.yaml
catastrophe-brandon: We'll have to keep the hash updated whenever the test-auth package's version of PW is updated. Labels would be easier, but security issues override ease of use. 

This requirement becomes more tricky if/when we adopt test dashboarding tools that extend Playwright's various test packages/classes.

I'm not in love with maintaining hashes but it is a maintenance requirement of our Konflux infrastructure.

